PR #4, opened 2023-08-29
Copy Over Complex Dashboard

jpjsonic:


tgsoverly:
Thanks!  I thought this wasn't used.  We should add a simple test to ensure this compiles since it is open source now.

jpjsonic:
Yup, definitely! I'm going to expand this and add the reference functionality too so that we have some integration testing running on that

test-projects/complex-project/visivo_project.yml
```diff
@@ -0,0 +1,16 @@
+name: coffee_facts
+
+defaults:
+  target_name: snowflake
+
+includes:
+  - path: coffee_trade/coffee_trade.yml
```
jpjsonic, line 7: @tgsoverly I think this type of pattern will be a very good way for people to organize larger projects. It might be beneficial for us to create a `concepts` directory in the docs that sits next to `reference` and `background` where we have some articles and recommendations on how to use the tool

tgsoverly (approved):

PR #8, opened 2023-09-16
Tim/add schema generator

tgsoverly:
Start work on the schema validator


tgsoverly:
Switched to different branch

PR #19, opened 2023-09-29
Find and replace name

tgsoverly:
Move to a new project naming format that is easier to support schema validation on included files.

`*.visivo.yml` in general and the main project file `project.visivo.yml`

Backward compatibility is maintained


mkdocs/topics/linting.md
```diff
@@ -1,20 +1,22 @@
...
 
 ## VSCode Plugin
 
 The YAML plugin by redhat-developer [here](https://github.com/redhat-developer/vscode-yaml)
  can lint yaml with a json schema.
 
-### Configuration
+### Configuration (Optional)
+
+By default any file with the format `*.visivo.yml` will be linted.  If you have a file that is named differently, you can add it to the local configuration.
```
jpjsonic, line 12: Does this mean that you don't need to edit the yaml schema at all if you are using the *.visivo.yml naming convention?

jpjsonic (approved):
🚀

mkdocs/topics/linting.md
tgsoverly: That is correct!

PR #20, opened 2023-09-30
Add Role to Target

jpjsonic:
In this PR: 
* Add roll to target for snowflake


tgsoverly (approved):
This looks good. We should add a card to separate the different target types like we did for trace props. Then we can have type of specific validation.

jpjsonic:
Ah that's a good point. I'll add a ticket for that.

PR #24, opened 2023-10-14
Import everything after we do the command

tgsoverly:
Two items here:
1. Import everything after the command starts to at least give the impression of responsiveness.  Still need to make the imports faster.
2. Add a command like dots to give feedback that something is happening.

Added a Logger class to wrap writing output to either the status or persistent command line.  

Screenshot 2023-10-16 at 9 45 49 PM


visivo/commands/compile.py
```diff
@@ -62,4 +10,11 @@ def compile(working_dir, output_dir, target):
     """
     Parses the files in your working directory, extracting visivo configurations and then using those configurations to build the trace queries and a project.json file in your target directory. Queries are not run on compile, just written.
     """
+    from visivo.logging.logger import Logger
+
+    Logger.instance().debug("Compiling")
+
+    from visivo.commands.compile_phase import compile_phase
```
jpjsonic, line 17: So a lot of the performance benefits are coming from moving imports into the command?

jpjsonic (approved):
Looks great! The Halo spinner seems perfect

visivo/commands/compile.py
jpjsonic: Or it just allows the tool to say it's doing something sooner maybe?

jpjsonic:
I noticed it hangs when doing the env step. Is this just because the first step is going to take the longest while doing the imports right now? I also noticed the same thing when running `visivo test`
https://github.com/visivo-io/visivo/assets/35309169/8094c9ee-1a6c-4c0c-aa75-d0c7045e8b67

jpjsonic:
I also noticed that in git actions it creates a bit of noise with all of the printouts to make the animation: 
image
Probably not a blocker on merging- but I wonder if there's something we can do to remove the animation when we're in a CICD tool. I don't know if there's really any way to know that. Maybe the presence of some env vars that exist in git actions specifically or something like that?

tgsoverly:
> I noticed it hangs when doing the env step. Is this just because the first step is going to take the longest while doing the imports right now? I also noticed the same thing when running `visivo test` https://github.com/visivo-io/visivo/assets/35309169/8094c9ee-1a6c-4c0c-aa75-d0c7045e8b67

Yes that is correct, I can change the log message.

tgsoverly:
> I also noticed that in git actions it creates a bit of noise with all of the printouts to make the animation: image Probably not a blocker on merging- but I wonder if there's something we can do to remove the animation when we're in a CICD tool. I don't know if there's really any way to know that. Maybe the presence of some env vars that exist in git actions specifically or something like that?

Whoops, had something for this.  Added it back for "CI" env.

PR #25, opened 2023-11-03
Speed Up CLI w/ Improved TraceProps

jpjsonic:
In this PR:

- [x] Consolidate Terminal TraceProps objects to create less objects (reduced raw count from 942 to 550)
- [x] Incorporate Chart.Layout objects and validation
- [x] Update Docs Logic to Include Layout


visivo/models/trace_props.py
jpjsonic, line 1: For the `Layout` object I set it up to allow extra inputs because the xaxis2 ... xaxisn and yaxis2... yaxisn are not included in the docs objects even though they are valid configurations. There's probably something clever we can do down the line to create these objects, but for now I thought this solution was adequate.

PR #33, opened 2023-12-10
Make the trace name the cohort on value

tgsoverly:
If the trace has a name, then use that as the cohort on value


jpjsonic (approved):
Looks good!

PR #35, opened 2023-12-11
Jared vis 304 fix postgres string groupby

jpjsonic:


test-projects/simple-database/project.visivo.yml
```diff
@@ -158,7 +158,7 @@ traces:
             - props.y
   - name: 3d Line Two
     model: ref(test table)
-    cohort_on: "'Line Two'"
+    #cohort_on: "'Line Two'"
```
tgsoverly, line 161: Can we get rid of the comments?

tgsoverly (approved):

test-projects/simple-database/project.visivo.yml
jpjsonic: Ah yeah that would be good. I'll get those removed in the next pr 👍

PR #38, opened 2023-12-20
Postgres Integration Testing

jpjsonic:
Largely following [this guide](https://docs.github.com/en/actions/using-containerized-services/creating-postgresql-service-containers) to set up a pg db, populate it with data and run visivo against it.


populate_ci_postgres_db.js
```diff
@@ -0,0 +1,29 @@
+const { Client } = require('pg');
```
jpjsonic, line 1: @tgsoverly where you think the right folder is to store this script?

tgsoverly: I would put it in the test folder, maybe `test/setup`?

tgsoverly (approved):
Nice work!

PR #39, opened 2023-12-27
Clean up validation for traces, unions, and location of error

tgsoverly:
Goals of this PR:
1. The validation is not printing an understandable path on the trace object when there is an error
2. The `ref` object unions are confusing because it gives you two errors one for the object and one for the ref
3. The validation doesn't print line numbers

For first item: by moving the model_validator to be "before" it removed the reference to the function on the model.
```
traces.0.function-after[validate_column_refs(), Trace].name
```
to
```
traces.0.Trace.name
```

The second item can be addressed by having our ref unions be "discriminated", so instead of giving two errors:
```
E       pydantic_core._pydantic_core.ValidationError: 2 validation errors for Chart
E       traces.0.constrained-str
E         Input should be a valid string [type=string_type, input_value={'props': {'type': 'scatt... 'select * from table'}}, input_type=dict]
E           For further information visit https://errors.pydantic.dev/2.5/v/string_type
E       traces.0.Trace.name
E         Field required [type=missing, input_value={'props': {'type': 'scatt... 'select * from table'}}, input_type=dict]
E           For further information visit https://errors.pydantic.dev/2.5/v/missing
```
It is just the one depending on if the object was a str or dict.

Finally, if the input can be found in the files, we print that information.

Screenshot 2024-01-01 at 1 48 38 PM


visivo/logging/logger.py
```diff
@@ -37,23 +37,16 @@ def info(self, message: str):
...
         elif self.spinner:
-            self.spinner.succeed(message)
+            self.spinner.succeed(text=message)
 
     def error(self, message: str):
         if self.echo:
             self.echo(message)
         elif self.spinner:
             self.spinner.stop_and_persist(text=message)
-    def info_with_symbol(self, message):
```
jpjsonic, line 53: This makes more sense. Thanks for cleaning it up.

visivo/parsers/line_validation_error.py
```diff
@@ -0,0 +1,61 @@
...
+    def __str__(self):
+        message = f"{self.validation_error.error_count()} validation errors in {self.validation_error.title}\n"
+        file_found = False
+        for error in self.validation_error.errors():
+            message = message + f"{'.'.join(map(lambda l: str(l), error['loc']))}\n"
+            message = message + f"  {error['msg']}\n"
+            line_message = self.get_line_message(error)
+            if line_message:
+                file_found = True
+                message = message + f"  The input used ({error['input']}) was found: \n"
```
jpjsonic, line 53: I wonder if we need to print the input when we find a line number? Do you think there's benefits to keeping it in there? To me it feels like the line number + the validation error message give you the details that you need to debug.

visivo/command_line.py
```diff
@@ -40,6 +41,9 @@ def load_env(env_file):
 def safe_visivo():
     try:
         visivo(standalone_mode=False)
+    except (ValidationError, LineValidationError) as e:
+        Logger.instance().error(str(e))
+        exit(1)
```
jpjsonic, line 46: This would exit the command with an error on the first validation error found right? It might be better to collect all of the validation errors first and then fail the command so that you don't have the run the command multiple times to find all of your errors.

visivo/logging/logger.py
```diff
@@ -37,23 +37,16 @@ def info(self, message: str):
...
             self.echo(message)
         elif self.spinner:
-            self.spinner.succeed(message)
+            self.spinner.succeed(text=message)
 
     def error(self, message: str):
         if self.echo:
             self.echo(message)
         elif self.spinner:
             self.spinner.stop_and_persist(text=message)
```
jpjsonic, line 52: I think there's an `spinner.error` halo object that we can use rather than stop and persist. I think that if we use the error object we'll get some nice formatting benefits.

jpjsonic (approved):
This looks awesome 🥳 Added a few suggestions.

visivo/command_line.py
tgsoverly: All the parse errors are collected by pydantic before it raises.

visivo/logging/logger.py
tgsoverly: Done.

tgsoverly:
I ran into some issues doing some testing of this feature, I need to rethink the line numbers.  I am doing the loading to record the locations of the object during load.  I think this is the only way to get the information reliably.

tgsoverly:
Updated output:
Screenshot 2024-01-03 at 11 24 08 AM

visivo/parsers/line_validation_error.py
tgsoverly: I kept it, I figured if the location of the error message can't be found, then all the help the better.

visivo/parsers/yaml_ordered_dict.py
```diff
@@ -0,0 +1,59 @@
+import yaml
+import json
+from collections import OrderedDict
+
+
+class YamlOrderedDict(OrderedDict):
+    """
+    An OrderedDict that was loaded from a yaml file, and is annotated
+    with file/line info for reporting about errors in the source file
```
jpjsonic, line 9: This is great, because I could see us wanting to use this context in other functionality outside of error messages (ie search functions, info logs ect.)

visivo/utils.py
```diff
@@ -77,19 +75,38 @@ def extract_value_from_function(function_text, function_name):
     value = match.group(1)
     return value.strip()
 
+
+def set_location_recursive_items(dictionary, file):
```
jpjsonic, line 79: Is this no longer used?

jpjsonic (approved):
🚀

tests/models/test_project.py
```diff
@@ -63,9 +63,9 @@ def test_Project_validate_project_trace_refs():
     error = exc_info.value.errors()[0]
     assert (
         error["msg"]
-        == f'Value error, The reference "ref(trace_name)" on item "Chart - chart" does not point to an object.'
```
jpjsonic, line 66: @tgsoverly Not blocking merge- It could be helpful to do a fuzzy string match (maybe something like jaro-winkler >= 90) against other trace names in the product to make a suggestion for the trace that the user was trying to reference.

jpjsonic:
@tgsoverly I found that if you save an invalid configuration while serving you end up getting the error message on loop, which isnt ideal. However with how it currently works if the user fixes the validation error then the project works again without having to re-serve which is super nice.

It would be awesome to have serve just print the error message once and then keep the server open so that we can check for additional file saves that fix the validation. I think this alone will make for a really really fast feedback loop for Dan & team. 


https://github.com/visivo-io/visivo/assets/35309169/29edaaa1-2ce4-458d-9049-95d4ecfe4270

visivo/parsers/line_validation_error.py
```diff
@@ -0,0 +1,61 @@
...
+    input_line = re.sub(
+        r"^[\s-]+", "", yaml.dump(input_dict), flags=re.MULTILINE
+    ).split("\n")[0]
+    for line_number, file_line in enumerate(file_lines):
+        if input_line in file_line:
+            return line_number + 1
+    return -1
+
+
+class LineValidationError(Exception):
```
jpjsonic, line 24: @tgsoverly made updates to the message that results in these changes. I think it helps readability, but curious what your thoughts are: 

Old Message: 
Screenshot 2024-01-04 at 8 40 17 AM
New Message:
Screenshot 2024-01-04 at 8 39 08 AM

jpjsonic (approved):

PR #40, opened 2024-01-04
Add is atty to logger

tgsoverly:
When piped.

Screenshot 2024-01-04 at 5 27 56 PM


jpjsonic (approved):
Got it, makes sense 👍

PR #41, opened 2024-01-07
Serve, run, and compile a subset of items

tgsoverly:
This filters the traces that are run and compiled for any node with a name.  Serve was more work to support because it is a local serve that works on the project json, I only implemented the dashboard level filtering for now.

Simple Project: `visivo serve -nf "Simple Dashboard"`

### DIFF
--- a/tests/models/test_project.py
+++ b/tests/models/test_project.py
@@ -15,32 +15,12 @@ from ..factories.model_factories import (
 from pydantic import ValidationError
 import pytest
 import networkx
-import click
 
 
 def test_Project_simple_data():
     data = {"name": "development"}
     project = Project(**data)
     assert project.name == "development"
-
-
-def test_Project_filter_trace():
-    project = Project(traces=[])
-    assert project.filter_traces(pattern="trace") == []
-
-    cat_trace = TraceFactory(name="cat")
-    cat_trace.model.name = "cat_model"
-    bobcat_trace = TraceFactory(name="bobcat")
-    bobcat_trace.model.name = "bobcat_trace"
-    project = Project(traces=[cat_trace, bobcat_trace])
-    assert project.filter_traces(pattern="cat") == [cat_trace, bobcat_trace]
-
-    cat_trace = TraceFactory(name="cat")
-    cat_trace.model.name = "cat_model"
-    bobcat_trace = TraceFactory(name="bobcat")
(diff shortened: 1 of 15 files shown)


tests/models/test_project.py
```diff
@@ -24,25 +23,6 @@ def test_Project_simple_data():
     assert project.name == "development"
 
 
-def test_Project_filter_trace():
```
tgsoverly, line 27: Removed this old filter

visivo/commands/serve_phase.py
```diff
@@ -12,14 +12,22 @@
...
     project_json = ""
     with open(f"{output_dir}/project.json", "r") as f:
         project_json = json.load(f)
+    
+    if name_filter:
+        dashboards = [d for d in project_json['dashboards'] if d["name"] == name_filter]
+        if len(dashboards) == 1:
+            project_json["dashboards"] = dashboards
+        else:
+            raise click.ClickException(f"Currently the serve command name filtering only supports filtering at the dashbaord level.  No dashboard with {name_filter} found.")
```
jpjsonic, line 25: 👍 Nice this a good error.

visivo/commands/serve_phase.py
```diff
@@ -12,14 +12,22 @@
...
 
 
-def get_project_json(output_dir):
+def get_project_json(output_dir, name_filter=None):
     project_json = ""
     with open(f"{output_dir}/project.json", "r") as f:
         project_json = json.load(f)
+    
+    if name_filter:
+        dashboards = [d for d in project_json['dashboards'] if d["name"] == name_filter]
```
jpjsonic, line 21: I wonder if there's way for us to read the project from the json back into a project object that has a dag so we can use the same filtering that you generated for the compile & run phase.

jpjsonic (approved):
Looks great!

visivo/commands/serve_phase.py
tgsoverly: We could do that for sure, might be a bit slower.

jpjsonic: Ah yeah, maybe after we address import performance that will be more practical

tgsoverly: Also, I just had some business questions about what would happen if you served a trace, for example

PR #42, opened 2024-01-08
Better Query Logs & Query Concurrency

jpjsonic:
In this PR

- [x] Threading database calls for faster execution.
- [x] Add time to run queries & more context to logs
- [x] Improve error message for incorrect SQL failures
- [x] Testing

### DIFF
--- a/visivo/logging/logger.py
+++ b/visivo/logging/logger.py
@@ -29,7 +29,7 @@ class Logger:
         else:
             self.spinner = Halo(text="Loading", spinner="dots")
             self.spinner.start()
-            time.sleep(0.1)
+            time.sleep(0.1)  # do we still need this?
             self.echo = None
 
     def debug(self, message: str):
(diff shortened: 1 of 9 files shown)


jpjsonic:
Now able to set threads for serve and run in the command line. Default concurrency is 8 threads. Also we now are printing a lot more information about the job and the run. 

image

Can also now fail a query while serve is running, fix it, save it and get back on with it. 

Screenshot 2024-01-08 at 1 39 00 AM

visivo/logging/logger.py
```diff
@@ -29,7 +29,7 @@ def set_type(self, type: TypeEnum):
         else:
             self.spinner = Halo(text="Loading", spinner="dots")
             self.spinner.start()
-            time.sleep(0.1)
+            time.sleep(0.1)  # do we still need this?
```
tgsoverly, line 32: Oh wow. Maybe not. I was trying to figure out an issue I don't even remember now.

tgsoverly (approved):

visivo/logging/logger.py
jpjsonic: Nice! Gonna pull it out then. I wonder if this is impact our import speed 🤔

PR #43, opened 2024-01-10
Typo regarding python version

ayazhafiz:
Hi, I noticed this said visivo requires python 10 - I'm assuming it should be 3.10? Feel free to close if this is inaccurate.


jpjsonic:
Ah, yup that's a typo, thank you!

PR #44, opened 2024-01-13
Fix double chart reference error

tgsoverly:
Items were getting dropped because of python equality.


jpjsonic (approved):
Nice, looks good! Thanks for jumping on this one

PR #45, opened 2024-01-13
Simplify how targets are handled

tgsoverly:
Remove the ability to pass in default target as a JSON string, and make it optional to have a default target if you only have traces that specify target names.

### DIFF
--- a/visivo/commands/utils.py
+++ b/visivo/commands/utils.py
@@ -9,35 +9,23 @@ from visivo.parsers.core_parser import PROFILE_FILE_NAME
 from visivo.utils import load_yaml_file
 
 
-def find_or_create_target(project: Project, target_or_name: str) -> Target:
-    if len(project.targets) == 0 and not target_or_name:
+def find_default_target(project: Project, target_name: str) -> Target:
+    if len(project.targets) == 0 and not target_name:
         raise click.ClickException(
-            f"The project must contain a target or a target must be passed as an object on the command line."
+            f"The project must contain a target."
         )
 
-    if not target_or_name and len(project.targets) == 1:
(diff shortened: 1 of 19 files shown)


visivo/commands/utils.py
```diff
@@ -9,35 +9,28 @@
 from visivo.utils import load_yaml_file
 
 
-def find_or_create_target(project: Project, target_or_name: str) -> Target:
-    if len(project.targets) == 0 and not target_or_name:
+def find_default_target(project: Project, target_name: str) -> Target:
+    if len(project.targets) == 0 and not target_name:
         raise click.ClickException(
-            f"The project must contain a target or a target must be passed as an object on the command line."
+            f"The project must contain a target."
```
jpjsonic, line 15: I wonder if it might be helpful to link to the target in the docs. I think our URLs are pretty stable at this point: https://docs.visivo.io/reference/configuration/Target/

visivo/models/project.py
```diff
@@ -101,6 +101,21 @@ def validate_default_names(self):
...
+        defaults = self.defaults
+        if defaults and defaults.target_name:
+            return self
+
+        if len(self.trace_objs) == 1:
+            return self
+
+        for trace in self.trace_objs:
+            if not trace.target_name:
+                raise ValueError(f"'{trace.name}' does not specify a target and project does not specify default target")
```
jpjsonic, line 116: This is a good error

jpjsonic (approved):
🚀

visivo/commands/utils.py
tgsoverly: The only reason I wouldn't want to do that is if stuff moves then we have to go back and change it all.

PR #49, opened 2024-01-24
Tim vis 330 escape character in secrets causes error

tgsoverly:


test-projects/simple-database/project.visivo.yml
```diff
@@ -5,7 +5,7 @@ includes:
   - path: visivo-io/example-include.git@main
 
 defaults:
-  target_name: local-sqlite
+  target_name: local-postgres
```
jpjsonic, line 8: Hmm, looking at this I wonder if we should actually keep the `-t` option for run and serve even though we discussed removing it.  

It's actually pretty useful to be able to pass a target in the command line to use instead of the default in instances like these where you can't use env vars because sqlite and postgres have different connection requirements, but might occasionally want to do a run with a different target.

visivo/utils.py
```diff
@@ -94,7 +94,7 @@ def set_location_recursive_items(dictionary, file):
 
 def load_yaml_file(file):
     def env_var(key):
-        return os.getenv(key, "NOT-SET")
+        return os.getenv(key, "NOT-SET").replace("\\", "\\\\").replace('"', '\\"')
```
jpjsonic, line 97: Ah nice! 
image

jpjsonic (approved):
🚀

test-projects/simple-database/project.visivo.yml
jpjsonic: Separately I think this change is failing the [deploy-dashboard](https://github.com/visivo-io/visivo/actions/runs/7647316700/workflow?pr=49) action. I'm not sure why the postgres test is failing tho.

tgsoverly: Good point, I was mostly talking about removing the ability to pass in a json target, which is already gone.  I kept the "default target".

PR #50, opened 2024-01-27
Tim vis 300 target flows into model rather than trace

tgsoverly:
Move the target to the model.

Also make it a `ref` or `Target` model.  This allows it to be more consistent with our other objects.

This is going to be a breaking change that will require a new version.


visivo/models/model.py
```diff
@@ -1,19 +1,28 @@
...
 
-    name: Optional[str] = Field(
+    sql: str = Field(
         None,
-        description="Name of the model",
+        description="The sql used to generate your base data",
     )
 
-    sql: str = Field(
+    target: Optional[generate_ref_field(Target)] = Field(
```
jpjsonic, line 19: 🥳 so it pretty much functions the same way that it did with the trace where if you don't specify it then it will use the default, that's great.

visivo/commands/utils.py
```diff
@@ -4,38 +4,11 @@
 from pathlib import Path
 from visivo.models.project import Project
 from visivo.models.target import Target
-from sqlalchemy import create_engine, MetaData, Table, Integer, Column, insert 
+from sqlalchemy import create_engine, MetaData, Table, Integer, Column, insert
 from visivo.parsers.core_parser import PROFILE_FILE_NAME
 from visivo.utils import load_yaml_file
 
 
-def find_default_target(project: Project, target_name: str) -> Target:
```
jpjsonic, line 12: We're you able to remove this because the flow of the dag is more complete now? That seems like a big win.

visivo/models/base/parent_model.py
```diff
@@ -18,39 +19,45 @@ def child_items(self):
...
+                        root=root,
+                        item=item,
+                        parent_item=parent_item,
                     )
-                    if len(dereferenced_items) == 1:
-                        dereferenced_item = dereferenced_items[0]
-                    else:
-                        raise PydanticCustomError(
-                            "bad_reference",
-                            f'The reference "{item}" on item "{parent_item.id()}" does not point to an object.',
```
jpjsonic, line 42: Did you find that the default error produced a better message?

visivo/query/runner.py
```diff
@@ -82,10 +82,9 @@ def run(self):
     def _run_trace_query(self, queue: Queue):
```
jpjsonic, line 82: The other thing that this change unlocks is being able to parallelize query runs  across different targets in addition to threading them within the target now that the dag flows from target to model to trace.

jpjsonic (approved):
This is great! Super excited for the features this unblocks

visivo/commands/utils.py
tgsoverly: Yea, I moved it over to using the dag, and setting the default on the project if it is passed in.

visivo/models/base/parent_model.py
tgsoverly: No I just moved this into a private function down the page.

PR #53, opened 2024-02-04
Add in cli version to project json

tgsoverly:
Start recording this on the project to allow for better debugging and adaptive UI


jpjsonic (approved):
Nice! This will be interesting to track in our dash too.

PR #54, opened 2024-02-05
Improve Query Failure Error Message

jpjsonic:


tgsoverly (approved):
Looks fine, worth an assertion on the test?

PR #55, opened 2024-02-05
Fix Target Bug

jpjsonic:


visivo/commands/run_phase.py
```diff
@@ -25,8 +28,9 @@ def changed(trace):
     traces = list(filter(changed, traces))
 
     Logger.instance().info(
-        f"Running project with {len(traces)} traces(s) across {threads} threads.\n"
+        f"Running project with {len(traces)} traces(s) across {threads} threads and default target {default_target}\n"
```
tgsoverly, line 31: Probably only worth adding this message if "default_target" is present.

tgsoverly (approved):

visivo/commands/run_phase.py
jpjsonic: Good point, adding that and then merging 👍

PR #58, opened 2024-02-07
Idea on how to make models even more flexible

tgsoverly:
Add a new script model and then allow for models to have sub_models as a type.

### DIFF
--- a/visivo/models/model.py
+++ b/visivo/models/model.py
@@ -1,14 +1,32 @@
-from typing import Optional
+from typing import List, Optional, Union
 from pydantic import Field, validator
 from visivo.models.base.named_model import NamedModel
 from visivo.models.base.parent_model import ParentModel
 from visivo.models.target import DefaultTarget, Target
 from .base.base_model import generate_ref_field
+from abc import ABC
 
 
-class Model(NamedModel, ParentModel):
+class Model(ABC):
     """
-    Models are queries that return base data that is used in Traces
+    Models return base data that is used in Traces
+    """
+
+
+class ScriptModel(Model, NamedModel):
+    """
+    Model that executes a script that returns csv data
+    """
+
+    run: str = Field(
+        None,
+        description="The command to run that returns CSV data",
+    )
+
+
+class SqlModel(Model, ParentModel, NamedModel):
+    """
+    Model that executes sql against a given database target
     """
 
     sql: str = Field(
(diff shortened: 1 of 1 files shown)


visivo/models/model.py
```diff
@@ -1,14 +1,32 @@
...
+class Model(ABC):
     """
-    Models are queries that return base data that is used in Traces
+    Models return base data that is used in Traces
+    """
+
+
+class ScriptModel(Model, NamedModel):
+    """
+    Model that executes a script that returns csv data
```
jpjsonic, line 18: This is super cool. Not a blocker- but security will be an interesting consideration down the line when we give people the ability to set up scheduled updates from the cloud app.

visivo/models/model.py
```diff
@@ -26,3 +44,21 @@ def child_items(self):
             return [self.target]
         else:
             return [DefaultTarget()]
+
+
+Models = Union[SqlModel, ScriptModel]
+
+
+class JoinModel(Model, NamedModel, ParentModel):
```
jpjsonic, line 52: Rather than having a separate join class I was thinking we could just add an attribute to the `sqlModel` & `ScriptModel` classes that tells visivo to store their output in memory maybe looks something like this- 

```
class SqlModel(Model, ParentModel, NamedModel):
    """
    Model that executes sql against a given database target
    """
class SqlModel(Model, ParentModel, NamedModel):
    """
    Model that executes sql against a given database target
    """

    sql: str = Field(
        None,
        description="The sql used to generate your base data",
    )
    store-output: Optional[Boolean] = Field(
       false, 
       description= "Stores the output of your model in the target cached sqlite database" 
    )
    target: Optional[generate_ref_field(Target)] = Field(
        None,
        description="A target object defined inline or a ref() to a chart. Override the defaults.target_name",
    )
    def child_items(self):
        if self.target:
            return [self.target]
        else:
            return [DefaultTarget()]
```

Then since it's stored in a sqlite database the user can write normal sql base models agains the sqlite database to join the items together.

visivo/models/model.py
tgsoverly: We would run those scripts in a docker, so the script would be inside that container.

visivo/models/model.py
tgsoverly: One of the issues we are going to run into order of operations.  If you have a model that uses other models, now they have to run in a specific order. 
 
 I was having a thought that maybe these should be targets instead.  I think it might simplify things, I am going to give that a shot.

tgsoverly:
```
models: 
  - name: Model Name
     sql: select * from other_model join third_model on ...
     sub_models:
        - ref(Other Model)
        - ref(Third Model)
  - name: Other Model
     sql: select * from table
  - name: Third Model
     sql: select * from table
```

jpjsonic: Yup, we'll need to order the runner que based on the DAG, but I don't think that will actually be too much work

PR #65, opened 2024-02-23
Tim vis 358 pool query runs

tgsoverly:
Rework the way the queries are run to check for dependencies and break out the thread logic.

An issue we are running into is that the threaded model doesn't play well with the DB pooling that is on by default in sqlalchemy.  A significant rework to move the engine outside the threads is needed.

- [x] Have a standard response from the jobs and have a printer print the start of a job and the response of the job
- [x] Add more tests
- [x] Move tracking into a model for better seperation


visivo/query/jobs/job.py
```diff
@@ -0,0 +1,60 @@
...
+        self.target = target
+        self.action = action
+        self.kwargs = kwargs
+        self.dependencies = dependencies
+        self.future: Future = None
+
+    def set_future(self, future):
+        self.future = future
+
+    def start_message(self):
```
tgsoverly, line 29: Test me

visivo/query/jobs/run_csv_script_job.py
```diff
@@ -0,0 +1,49 @@
...
+        failure_message = format_message_failure(
+            details=f"Failed query for model \033[4m{csv_script_model.name}\033[0m",
+            start_time=start_time,
+            full_path=csv_script_model.get_database(output_dir=output_dir),
+            error_msg=str(repr(e)),
+        )
+        return JobResult(success=False, message=failure_message)
+
+
+def jobs(dag, project, output_dir):
```
tgsoverly, line 32: Test me

visivo/query/jobs/run_trace_job.py
```diff
@@ -0,0 +1,83 @@
...
+        return targets[0]
+
+    model = ParentModel.all_descendants_of_type(type=Model, dag=dag, from_node=trace)[0]
+    if isinstance(model, CsvScriptModel):
+        return model.get_target(output_dir)
+    else:
+        return model.target
+
+
+def jobs(dag, output_dir, project):
```
tgsoverly, line 63: Test me

visivo/query/target_job_tracker.py
```diff
@@ -0,0 +1,126 @@
...
+        self.job_queue = Queue()
+
+    @property
+    def target_names(self):
+        return list(
+            map(lambda target_limit: target_limit.target_name, self.target_limits)
+        )
+
+    @property
+    def all_tracked_job_names(self):
```
tgsoverly, line 60: Test missing methods here

visivo/query/jobs/job.py
```diff
@@ -0,0 +1,61 @@
...
+        return _format_message(
+            details=f"Running job for item \033[4m{self.name}\033[0m", status="RUNNING"
```
jpjsonic, line 32: I might be nice to use the pydantic model (node) type rather than `item` in the message. We could further color the node type blue so that it stands out and makes it easy for people to see if a model or trace is being updated. 
image
In the image above It would be `Running job for Trace(blue) Double Simple Line`

visivo/query/jobs/job.py
```diff
@@ -0,0 +1,61 @@
...
+    details = textwrap.shorten(details, width=80, placeholder="(truncated)") + " "
+    num_dots = total_width - len(details)
+    dots = "." * num_dots
+    if not full_path:
+        return f"{details}{dots}[{status}]"
+    current_directory = os.getcwd()
+    relative_path = os.path.relpath(full_path, current_directory)
+    error_str = "" if error_msg == None else f"\n\t\033[2merror: {error_msg}\033[0m"
+    return (
+        f"{details}{dots}[{status}]\n\t\033[2mquery: {relative_path}\033[0m" + error_str
```
jpjsonic, line 48: For the model's the `relative_path` links to the database file rather than the query so we might want to create some logic for the word `query` to change depending on the node type being formatted.

jpjsonic (commented):
I noticed when running `visivo serve` locally that all of the traces updated when I made and saved a change to one rather than just the one I edited. I'm trying to figure out why that's happening. Probably something to do with the runner?

jpjsonic:
It also would be helpful to collect all of the errors and print them at the end of the job run. Production projects will likely get to a point where they are over 1k nodes running and scrolling through the logs won't be practical.

jpjsonic (approved):

visivo/query/jobs/run_csv_script_job.py
```diff
@@ -0,0 +1,56 @@
...
+        )
+        return JobResult(success=False, message=failure_message)
+
+
+def jobs(dag, output_dir: str, project: Project, name_filter: str):
+    csv_script_models = ParentModel.all_descendants_of_type(
+        type=CsvScriptModel, dag=dag, from_node=project
+    )
+
+    if name_filter:
```
tgsoverly, line 38: Test name filter

visivo/query/jobs/run_trace_job.py
```diff
@@ -0,0 +1,87 @@
...
+    else:
+        return model.target
+
+
+def jobs(
+    dag, output_dir: str, project: Project, name_filter: str
+):
+    jobs = []
+
+    traces = project.filter_traces(name_filter=name_filter)
```
tgsoverly, line 69: Test name filter and changed

visivo/query/runner.py
```diff
@@ -1,79 +1,65 @@
...
 
 class Runner:
     def __init__(
         self,
-        traces: List[Trace],
         project: Project,
         output_dir: str,
         threads: int = 8,
         soft_failure=False,
+        run_only_changed=False,
```
tgsoverly, line 26: Test two items

visivo/query/jobs/job.py
tgsoverly: Didn't do the blue just to keep it small, but reworked the job to have the item and put that in here.

tgsoverly:
> It also would be helpful to collect all of the errors and print them at the end of the job run. Production projects will likely get to a point where they are over 1k nodes running and scrolling through the logs won't be practical.

How do you feel about doing this item separately?  Just in the interest in managing scope?

jpjsonic:
> > It also would be helpful to collect all of the errors and print them at the end of the job run. Production projects will likely get to a point where they are over 1k nodes running and scrolling through the logs won't be practical.
> 
> How do you feel about doing this item separately? Just in the interest in managing scope?

Yup, that makes sense to me

PR #67, opened 2024-03-02
Jared vis 363 meta card for documentation updates part 2

jpjsonic:


mkdocs/topics/deployments.md
```diff
@@ -34,9 +31,97 @@ It's highly recommended that you create a staging version of your project anytim
...
+      python-version: 3.10.0
+  
+  - key: install-visivo
+    use: [python]
+    run: pip install git+https://github.com/visivo-io/visivo.git@latest #(4)!
+  
+  - key: gcloud-cli # (1)!
+    call: google-cloud/install-cli 1.0.1
+  
+  - key: gcloud-login # (2)!
```
tgsoverly, line 78: I think I would break out the cloud SQL into a different item because I think most people won't need this.

jpjsonic: Yeah that's a good point. I'll tab them out with the basic script showing by default.

jpjsonic: Done 👍

PR #68, opened 2024-03-09
Fix Visivo Init

jpjsonic:
Fix init and fix testing of init.


tests/commands/test_init.py
```diff
@@ -8,8 +8,8 @@
 
 
 def test_init_with_sqlite():
-    if os.environ.get("CI"):
```
jpjsonic, line 11: @tgsoverly is there a reason we'd want to skip this test in CI?

tgsoverly: I wasn't ever able to get the testing working in a headless environment.

I have been thinking about the init method and wonder if we should remove it in favor of easier generators that create files with specific yml in them.

For example

`visivo generate target postgres` then it would only ask questions regarding postgres connection.  The output would then be in a file like `targets/postgres-target-name.yml` and it would print something like `Target generated in file ... add it to your project with "includes - targets/postgres-target-name.yml"`

jpjsonic: I think that is probably the right approach so I'll just do something that patches this up and we can look at that as something that makes the whole set up experience better.

PR #73, opened 2024-03-24
New Jinja Functions & Better Docs

jpjsonic:
In this PR- 

- [x] Add new datetime jinja functions 
- [x] Add documentation for JInja Function 
- [x] Add docs for Jinja Looping 
- [x] Docs for jinja variables 
- [ ] Docs for Visivo Run Time Variables


mkdocs/reference/functions/index.md
```diff
@@ -0,0 +1,4 @@
+# Functions
+## What Can Functions Do For You?
+Functions enable you to connect your Visivo configurations, create re-usable object, inject context and produce different outputs in different environments.
```
tgsoverly, line 3: What do yo mean "connect your Visivo configurations"?

visivo/templates/render_yaml.py
```diff
@@ -0,0 +1,109 @@
+import time
+import datetime
+from dateutil import parser
+import jinja2
+import os
+
+
+def env_var(key):
+    return os.getenv(key, "NOT-SET").replace("\\", "\\\\").replace('"', '\\"')
```
tgsoverly, line 9: Where was this?  I think it should be `${key}-ENVIRONMENT-VARIABLE-NOT-SET` if possible.

visivo/utils.py
```diff
@@ -93,14 +93,10 @@ def set_location_recursive_items(dictionary, file):
 
 
 def load_yaml_file(file):
-    def env_var(key):
```
tgsoverly, line 96: Oh it was here... :-)

tgsoverly (approved):

visivo/templates/render_yaml.py
jpjsonic: I moved this function over as it was from utils.py. I can definitely update it to  say that instead.

mkdocs/reference/functions/index.md
jpjsonic: I was talking about how ref connects objects together in the dag

PR #79, opened 2024-05-11
Improve loading and layout

tgsoverly:
Loaders now have text and are centered and the overall layout has a minimum width and then changes the layout to a single column
Screenshot 2024-05-10 at 10 38 18 PM


jpjsonic (approved):
This is great! 🚀

PR #83, opened 2024-05-18
Have local merge models support references

tgsoverly:
Local merge models did not properly support `ref` functions.  I needed to push the dag down further and base more things on it.  I think this is a good thing.


test-project/local.db
jpjsonic, line 1: Ah my bad, I think I accidentally committed this.

visivo/query/jobs/run_trace_job.py
```diff
@@ -55,8 +57,10 @@ def _get_target(trace, dag, output_dir):
         return targets[0]
 
     model = ParentModel.all_descendants_of_type(type=Model, dag=dag, from_node=trace)[0]
-    if isinstance(model, CsvScriptModel) or isinstance(model, LocalMergeModel):
+    if isinstance(model, CsvScriptModel):
         return model.get_sqlite_target(output_dir)
+    elif isinstance(model, LocalMergeModel):
+        return model.get_sqlite_target(output_dir, dag)
```
jpjsonic, line 63: Is the idea that passing the dag enables nesting the local merge models?

jpjsonic (approved):
🚀

visivo/query/jobs/run_trace_job.py
tgsoverly: Yea because determining the models is dependent on ref, which needs the dag.

PR #84, opened 2024-05-19
Add demo project and make hello world example for init

tgsoverly:
Screenshot 2024-05-20 at 1 55 10 PM


test-projects/demo/dashboards/welcome.visivo.yml
```diff
@@ -0,0 +1,82 @@
+models:
+  - name: Waterfall Model
```
tgsoverly, line 2: @jpjsonic I would like to replace this with a CSV.  You know off the top of your head what does this data look like?

jpjsonic: Yup, I can push up a quick PR that turns this into a csv 👍

PR #88, opened 2024-05-24
Idea on how the project could look

tgsoverly:


test-projects/integration/charts.visivo.yml
```diff
@@ -0,0 +1,44 @@
...
+
+    # Introduce the idea of wild card ref.
+    on_select: ref(* Chart)
+    traces:
+      - name: Clickable Trace
+        # In addition to the select event, we could add on_click, for when a users
+        # clicks a specific value in the items trace.
+        on_click:
+          target: ref(Child Chart)
+          value: y
```
jpjsonic, line 24: Would the value be whatever the value of the trace prop.y is where the values was clicked? So if maybe you had something where for whatever reason you wanted to pull a value from a nested trace prop you would specify it with dot notation ie.  line.color

test-projects/integration/charts.visivo.yml
```diff
@@ -0,0 +1,44 @@
...
+
+        props:
+          type: scatter
+
+      - ref(Second Line)
+
+  - name: Child Chart
+    selector:
+      default: Simple Line
+      is_multi: false
```
jpjsonic, line 34: I like this. I think it provides a nice way to set up the selector. Especially with the defaults. Then we can start to show people examples of how they might include multiple grains ie.) several date grains. We'll have to think through how the selector will pick select traces across multiple charts though. 

Using the date grain example- Say you had a dashboard with 3 charts- sales over time, profits over time and employees over time. On each of these three charts you have 5 traces different traces. So on the `sales over time` chart you would have traces `sales per day`, `sales per week`, `sales per month`, `sales per quarter`,  and `sales per year`. Same pattern is repeated for the other two charts. 

You would want the selector to filter change the date grain, so we'd need some sort of tag or meta data concept on the traces so that the selector can operate on attributes other than the trace name.

test-projects/integration/charts.visivo.yml
```diff
@@ -0,0 +1,44 @@
+selectors:
+  - name: Common Selector
+    default: Simple Line
+    is_multi: false
```
jpjsonic, line 4: Nice, this is good to be able to split it out. Then we can include it in dashboards as an independent item.

test-projects/integration/charts.visivo.yml
```diff
@@ -0,0 +1,44 @@
...
+    selector:
+      default: Simple Line
+      is_multi: false
+    traces:
+      - ref(Simple Line)
+      - ref(Second Line)
+
+  - name: Common Selector Chart
+    # Since this chart uses the common selector, it would respond to the "Command Chart" selection above.
+    selector: ref(Common Selector) # In this situation we might want to hide the selector on this chart.
```
jpjsonic, line 41: Yeah, I think if we're using a common one it should probably be hidden, or maybe a button that takes the users mouse to the common selector replaces the in chart selector.

test-projects/integration/charts.visivo.yml
```diff
@@ -0,0 +1,44 @@
...
+  - name: Parent Chart
+
+    # Introduce the idea of wild card ref.
+    on_select: ref(* Chart)
+    traces:
+      - name: Clickable Trace
+        # In addition to the select event, we could add on_click, for when a users
+        # clicks a specific value in the items trace.
+        on_click:
+          target: ref(Child Chart)
```
jpjsonic, line 23: What about update rather than target? I think we should reserve target for database connections.

test-projects/integration/charts.visivo.yml
```diff
@@ -0,0 +1,44 @@
+selectors:
+  - name: Common Selector
+    default: Simple Line
+    is_multi: false
+
+charts:
+  - name: Command Chart
+    # The on_ commands can take a chart, table, or selector reference
+    # It would then set the value of the selector to the selected trace/cohort from this item
+    on_select: ref(Common Selector)
```
jpjsonic, line 10: What's the difference between `on_select` here and the `selector` block on line 32?

jpjsonic: Couldn't we call both `selector`?

tgsoverly: Would it make more sense if the "on_select" was moved:

```
chart: 
  name: Chart
  selector:
    on_select: ref(Other Chart)
```

It might even make sense to call that line: `update_on_select` ?

test-projects/integration/charts.visivo.yml
tgsoverly: I added the idea of a "matcher"

jpjsonic: A regex match could work. I could see that getting tricky for larger project too though. 

What if the selectors were able to select traces based on things other than the trace name? We could include a `meta` attribute on traces where people could put arbitrary key value pairs. Then people can configure sets of traces to have similar ids that a selector could look at. 

So on a trace you could have 
```
traces: 
  - name: tracy-trace
    meta: 
      date-grain: daily
```

then on the selector you could specify what you want to look for in the trace to make the section of traces 
```
selectors:
  - name: date-grain
    is_multi: false
    default: 0 # could base this on the position in the list or the title.
    options: 
      - title: day
        attribute: trace.meta.date-grain # location of attribute on trace
        value: daily
      - title: week
        attribute: trace.meta.date-grain # could do any trace attribute. 
        value: weekly # we could still allow for regex here in sitatuions where you might to select multiple traces with one option. 
      - title: month
        attribute: trace.meta.date-grain # location of attribute on trace
        value: monthly

```

tgsoverly: Oh, interesting.  I like the cleanliness of that.  This meta idea is a bit like "tags" I had in mind for other items.

```
selectors: 
  - name: date-grain 
     is_multi: false 
     default: 0 # could base this on the position in the list or the title. 
     options: 
        - title: day 
           attribute: meta.date-grain. # I don't think you need the trace
           value: daily
```

One issue I see with this is you would need to specific all the options.  It would work great for trace selection, but not cohort_on selection.

jpjsonic: Right, due to the cohort_on we do still need a way to just have the selector options just be all trace-cohorts on a chart, or do a regex based on the plotly trace name that gets generated for the trace-cohorts

jpjsonic: I'm down to call the generic key value store `tags` rather than meta.

jpjsonic: >RE:
>        ` attribute: meta.date-grain. # I don't think you need the trace`
I was thinking it might be helpful to include that if we wanted to expand the selector to look at other attributes that are further up stream from the trace on the dag like chart layout values.

PR #90, opened 2024-05-25
Enable tables to take an array of traces and simplify usage

tgsoverly:
To drive consistency, and allow for the selection of data in tables, we need to change tables to allow an array of traces

Todo
- [x] Update table toolbar to include selector


test-projects/demo/dashboards/welcome.visivo.yml
```diff
@@ -71,9 +71,14 @@ dashboards:
...
-                - header: Widget Type
-                  column: x_data
-                - header: Widget Sales
-                  column: y_data
+                - trace_name: Fibonacci Waterfall Trace
+                  column_defs:
+                    - header: Widget Type
+                      key: columns.x_data
+                    - header: Widget Sales
+                      key: props.y
```
tgsoverly, line 82: I am thinking this way of referencing could make it easier to use and you would have to add column to the trace.  Thoughts?

jpjsonic:
> To drive consistency, and allow for selection of data in tables, we need to change tables to allow an array of traces

We could also allow for data selection in tables by using the material react column filtering. THe other way we could eventually build interactivity into tables is to enable actions on other charts/traces to group and aggregate tables using the material react table functionality.

test-projects/demo/dashboards/welcome.visivo.yml
jpjsonic: Yeah I like this better.

test-projects/demo/dashboards/welcome.visivo.yml
```diff
@@ -71,9 +71,14 @@ dashboards:
           - width: 2
             table:
               name: Fibonacci Waterfall Table
-              trace: ref(Fibonacci Waterfall Trace)
+              traces:
+                - ref(Fibonacci Waterfall Trace)
               columns:
```
jpjsonic, line 76: We should make this optional and pull in all of the trace data if it's not specified.

test-projects/demo/dashboards/welcome.visivo.yml
```diff
@@ -71,9 +71,14 @@ dashboards:
           - width: 2
             table:
               name: Fibonacci Waterfall Table
-              trace: ref(Fibonacci Waterfall Trace)
+              traces:
```
jpjsonic, line 74: Do we want to scope in allowing people to create tables directly from models as well?

test-projects/demo/dashboards/welcome.visivo.yml
```diff
@@ -71,9 +71,14 @@ dashboards:
...
+                - ref(Fibonacci Waterfall Trace)
               columns:
-                - header: Widget Type
-                  column: x_data
-                - header: Widget Sales
-                  column: y_data
+                - trace_name: Fibonacci Waterfall Trace
+                  column_defs:
+                    - header: Widget Type
```
jpjsonic, line 79: We should also make the header optional and just parse the key name pulling out the last value splitting by the dot, removing underscores and capitalizing ie `key.split('.')[-1].replace('_', '').capitalize`

jpjsonic: So in this example if you didn't specify a header you would get `X Data`

viewer/src/components/items/Table.js
```diff
@@ -1,28 +1,48 @@
 import Loading from "../Loading";
-import React from "react";
+import React, { useState } from "react";
 import { MaterialReactTable } from 'material-react-table';
```
jpjsonic, line 3: Do we want to provide a way for people to enable [aggregation & grouping](https://www.material-react-table.com/docs/guides/aggregation-and-grouping) in this pr while we're working on the tables?

test-projects/demo/dashboards/welcome.visivo.yml
tgsoverly: Perfect, same page.

test-projects/demo/dashboards/welcome.visivo.yml
tgsoverly: I do think that is a good idea, but maybe a follow up?

viewer/src/components/items/Table.js
tgsoverly: I could just turn it on if that is as easy as it needs to be.

test-projects/demo/dashboards/welcome.visivo.yml
tgsoverly: Included.

test-projects/demo/dashboards/welcome.visivo.yml
tgsoverly, line 82: Add tests around new names for header and key, and just key.

jpjsonic (approved):
This is a big one, looks great!

jpjsonic (approved):
Looked great testing it out!

PR #94, opened 2024-06-06
Added mint dashboard deploy with commenting

jpjsonic:
* Test out using the GH command line & Mint Apps. 
* Set up integration test for mint deployment

### DIFF
--- a/.mint/deploy_dashboard.yml
+++ b/.mint/deploy_dashboard.yml
@@ -27,7 +27,7 @@ tasks:
   
   - key: install-visivo
     use: [python]
-    run: pip install git+https://github.com/visivo-io/visivo.git@latest
+    run: pip install git+https://github.com/visivo-io/visivo.git@{{ init.head-ref }}
   
   
   - key: run-visivo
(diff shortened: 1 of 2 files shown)


.mint/deploy_dashboard.yml
```diff
@@ -0,0 +1,62 @@
...
+      visivo --version
+      visivo run
+    
+  - key: deploy-ci-stage
+    if: ${{ init.deploy }}
+    use: [run-visivo, github-cli]
+    run: |
+      cd test-projects/integration
+      visivo deploy -s ${{ init.head-ref }}-mint | tee /dev/stderr | grep 'Deployed to: ' > deployed.txt
+      gh pr comment ${{ init.pr-num }} -F deployed.txt
```
jpjsonic, line 48: It would be good to add the flag `--edit-last` but we have to find a way to only do it when it's the 2-n time deploying.

tgsoverly: Is there a way to fetch them?

tgsoverly: https://github.com/marocchino/sticky-pull-request-comment/blob/main/src/comment.ts

tgsoverly: Looks like the search for previous comment. Can be a second pass.

PR #95, opened 2024-06-08
Link selectors between charts and tables

tgsoverly:
Be able to reference one selector from one chart and have it change another charts or tables selection.

Note: If you reference another selector, it hides it on this chart.

https://github.com/visivo-io/visivo/assets/482572/9680e64f-5dee-42ae-8c96-db14692b5300


test-projects/demo/dashboards/welcome.visivo.yml
```diff
@@ -60,6 +60,9 @@ dashboards:
           - width: 2
             chart:
               name: Fibonacci Waterfall Chart
+              selector:
+                name: Fibonacci Waterfall Selector
+                type: single
```
jpjsonic, line 65: Nice this really slick. Super easy to configure.

visivo/models/selector.py
```diff
@@ -0,0 +1,47 @@
+from enum import Enum
+from pydantic import ConfigDict, Field, PrivateAttr, model_serializer
+
+from visivo.models.base.base_model import BaseModel
+from visivo.models.base.named_model import NamedModel
+
+
+class SelectorType(str, Enum):
+    single = "single"
```
jpjsonic, line 9: What's the default selection if you choose single? I don't know if it makes sense to do in this pass but we should allow people to specify a default selection(s) for both single and multi. I guess single would be one string and multi would be a list of strings maybe?

visivo/models/selector.py
```diff
@@ -0,0 +1,47 @@
...
+    multiple = "multiple"
+
+
+class Selector(NamedModel, BaseModel):
+    """
+    Selectors can be used to add interactivity between charts and tables.
+
+    Below is how you would link two charts to show the same selected data.
```
jpjsonic, line 17: Yay for docs! We could add a a gif of your six second video in the docs too.

jpjsonic (approved):
This is awesome!

visivo/models/selector.py
tgsoverly: It grabs the "first" one.  Yea I agree with that!

visivo/models/selector.py
tgsoverly: Does the video live in the docs here or on "interactivity"?

PR #101, opened 2024-06-30
Working test structure

tgsoverly:
Working conversation to the python-based tests.


tests/testing/test_runner.py
```diff
@@ -19,20 +23,22 @@ def test_TestQueryStringFactory_errors(capsys):
         },
         "model": {"sql": "select * from test_table", "target": "ref(target)"},
         "tests": [
-            {"coordinate_exists": {"coordinates": {"x": 3, "y": 2}}},
-            {"coordinate_exists": {"coordinates": {"x": 19, "y": 26}}},
+            {"logic": "assert_that(numpy.sum(trace.props.x)).is_equal_to(1)"},
```
tgsoverly, line 26: Here is the new syntax

jpjsonic (approved):
Looks great! I think we can work tests into the DAG in a future pass. We could set it up so that tests run in conjunction with models & traces. Maybe that's a whole new command. I do like the idea of being able to do them separately too.

PR #102, opened 2024-07-01
Working dist command

tgsoverly:
Creates a dist folder, I uploaded it to netlify and it worked:

https://6682bb59a6c6572444499838--harmonious-pithivier-d5635a.netlify.app/


mkdocs/topics/deployments.md
```diff
@@ -477,4 +477,8 @@ jobs:
         env:
           VISIVO_TOKEN: ${{ secrets.VISIVO_TOKEN }}
 ```
-{% endraw %}
\ No newline at end of file
+{% endraw %}
+
+## Static
+
+You can also manage and host the dashboard internally if you wish to manage it entirely within your systems.  To create a self-contained build all you need to do is run the `visivo dist` command.  This will create a folder that you can use to host in any manner that suits your needs.
```
jpjsonic, line 484: I don't think this blocks merge, but we could add something about how to deploy it with different providers like netlify too.

requirements.txt
jpjsonic, line 1: Did you mean to add this?

jpjsonic (approved):
🚀 This is great!

mkdocs/topics/deployments.md
tgsoverly: Netlify doesn't support python 3.10 ... very odd.

PR #106, opened 2024-07-13
Pytest CI Pushes Tests Results to Webhook

jpjsonic:
- [x] Write pytest results to fivetran webhook for analysis & use of pytest dashboard
- [x] Fix excessive mint-dashboard posts


tgsoverly (approved):
Two nice additions

PR #107, opened 2024-07-15
Improve CLI speed

tgsoverly:
Takes version command down from 8 secs to less than one.

The big changes here:
1. Move trace props into separate files
2. Only load all the classes when we need them to generate the schema and docs
3. Only load the needed trace_props based on the type when needed.


visivo/command_line.py
```diff
@@ -18,12 +18,34 @@
...
 @click.group()
+@click.option("--profile", is_flag=True)
 @click.option("-e", "--env-file", default=".env")
 @click.version_option(version=importlib.metadata.version("visivo"))
-def visivo(env_file):
+def visivo(env_file, profile):
     Logger.instance().set_type(TypeEnum.spinner)
     load_env(env_file)
 
+    if profile:
```
jpjsonic, line 28: This is smart, then we can check on it in our CI with a benchmark too.

visivo/command_line.py
```diff
@@ -18,12 +18,34 @@
...
+        pr = cProfile.Profile()
+        pr.enable()
+
+        def exit():
+            pr.disable()
+            Logger.instance().info("Profiling completed")
+            s = io.StringIO()
+            pstats.Stats(pr, stream=s).sort_stats("cumulative").print_stats()
+            Logger.instance().info(s.getvalue())
+            with open("profile_results", "w") as file:
```
jpjsonic, line 44: Would be cool to set up a dashboard that analyses this data with visivo and use our testing suite to set benchmark levels that our cli speed needs to pass.

jpjsonic: Might also be interesting to push the data to snowflake so we can monitor the speed over time.

jpjsonic (approved):
I like the profiler idea. Have you seen any speed improvements from the imports that you moved or is this more so setting up the monitoring?

visivo/models/targets/sqlalchemy_target.py
```diff
@@ -14,7 +12,10 @@ class SqlalchemyTarget(Target, ABC):
     def get_dialect(self):
         raise NotImplementedError(f"No dialect method implemented for {self.type}")
 
-    def read_sql(self, query: str) -> DataFrame:
+    def read_sql(self, query: str):
+        from pandas import DataFrame
```
jpjsonic, line 16: We could probably refactor this to not use pandas at all, if that would help.

mkdocs.yml
```diff
@@ -109,8 +109,9 @@ nav:
                   - reference/configuration/Models/CsvScriptModel/index.md
               - LocalMergeModel:
                   - reference/configuration/Models/LocalMergeModel/index.md
-          - Table:
```
tgsoverly, line 112: @jpjsonic Do you know why this would happen?

jpjsonic: Hmm, I'm not sure. That's pretty surprising behavior. I probably need to refactor the mkdocs parser of our schema in general it's a little confusing to make make changes to and test right now.

jpjsonic: I'll spend some time with it and see why it's doing that.

jpjsonic: Per our thread in slack- it seems like this is pertaining to having two pydantic models named Table. We can get rid of the plotly trace prop for table since we dont want to support it anyways.

visivo/models/base/parent_model.py
```diff
@@ -146,9 +149,12 @@ def name_match(trace):
 
     @staticmethod
     def show_dag(dag):
```
jpjsonic, line 151: Huh, I never new we had a way to show the built dag

tgsoverly: I use it all the time!

visivo/command_line.py
tgsoverly: I think these are great ideas, but this is massive already... 😬

visivo/models/targets/sqlalchemy_target.py
tgsoverly: I think right now it is fine, we can look into other solutions if we need in the future.

PR #108, opened 2024-07-21
Hide menu when select not visible

tgsoverly:
Fix the display bug I saw in the demo


jpjsonic:
I installed this version and re-ran the linear dashboard and still saw the same issue happening. Maybe I did the install wrong? - `pip install git+https://github.com/visivo-io/visivo.git@tim-hide-menu-when-select`

tgsoverly:
It won't be in the package until release. That is when we build the JavaScript.

PR #111, opened 2024-07-31
Append selection for item.selector

tgsoverly:
This causes all `item.selectors` to always push there state into the URL.  This causes those more global selectors to cause interaction upon load.


viewer/src/components/items/CohortSelect.test.js
```diff
@@ -63,6 +65,26 @@ test('renders single select', async () => {
   expect(visible).toEqual(true);
 });
 
+test('renders with push to history', async () => {
+  let history = createBrowserHistory();
```
jpjsonic, line 69: Nice this is cool. So this will enable people to use the browser back button?

jpjsonic (approved):
🚀

PR #112, opened 2024-08-01
Too deep of a search resulted in too many traces on a chart

tgsoverly:
With selectors it was finding all traces on a chart.  This meant if a chart was referencing a selector, that referenced different traces, it would go "through" the sector and return all the traces.  The solution was to limit the depth of the search to only include those items of a type directly attached to the given object when dereferencing.

### DIFF
--- a/visivo/models/base/parent_model.py
+++ b/visivo/models/base/parent_model.py
@@ -84,21 +84,27 @@ class ParentModel(ABC):
                     )
 
     @staticmethod
-    def all_descendants(dag, from_node=None):
+    def all_descendants(dag, from_node=None, depth=None):
         import networkx.algorithms.traversal.depth_first_search as dfs
 
-        return dfs.dfs_tree(dag, from_node)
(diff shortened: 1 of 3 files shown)


visivo/models/base/parent_model.py
```diff
@@ -150,10 +156,10 @@ def name_match(trace):
     @staticmethod
     def show_dag(dag):
         import matplotlib.pyplot as pyplot
-        from networkx import planar_layout, draw_networkx
+        from networkx import spring_layout, draw_networkx
 
         options = {}
-        pos = planar_layout(dag)
+        pos = spring_layout(dag)
```
jpjsonic, line 162: It would be interesting to look at this for the linear project. Could be cool to add this chart to the viewer by default.

jpjsonic (approved):
Thanks for jumping on this!

visivo/models/base/parent_model.py
tgsoverly: I think we need a different layout, it is quite ugly.  :-(

PR #115, opened 2024-08-10
Remove deprecated pkg_resources

tgsoverly:
Migrated to use the stdlib package


jpjsonic (approved):
👍 Ah nice easy enough

PR #117, opened 2024-08-16
Example changes

tgsoverly:


test-projects/demo/dashboards/welcome.visivo.yml
```diff
@@ -9,30 +9,37 @@ models:
...
+          # Equivalent to:
+          # color:
+          #   static: #b97a9b
       decreasing:
         marker:
           color: "#edbdb5"
-    order_by:
-      - query(row asc)
-    filters:
-      - query( waterfall is not null )
```
jpjsonic, line 33: I think I actually like requiring the query function here. Even though it's more verbose it makes it really easy to glean what's going into the trace query without knowing much about the attributes of the trace

test-projects/demo/dashboards/welcome.visivo.yml
```diff
@@ -72,9 +79,9 @@ dashboards:
           - width: 2
             table:
               name: Fibonacci Waterfall Table
-              selector: ref(Fibonacci Waterfall Selector)
+              selector: Fibonacci Waterfall Selector
               traces:
-                - ref(Fibonacci Waterfall Trace)
```
jpjsonic, line 77: What's your thinking behind getting rid of refs? I find them to be pretty useful to really quickly see when visivo is referring to another object in a configuration block.

tgsoverly: My thought is that it isn't needed.  We never have a situation where we can have a string or an object that isn't a ref.

test-projects/project.visivo.yml
```diff
@@ -0,0 +1,79 @@
+# V2 Spec
```
tgsoverly, line 1: Looking at this I think there are three types of things that we need to be consistent about:
1. Items that reference the project itself. `trace.meta.date-grain`
2. Items that reference the output of queries
3. Items that have executable code

In addition we have emerging need to combined at least 1 and 3 if we do `Selector` where you can have a matcher.

test-projects/project.visivo.yml
```diff
@@ -0,0 +1,79 @@
...
+  - name: if_production
+    logic: if env == production
+  - name: if_morning_notifications
+    logic: if env["MORNING_NOTIFICATION"] == true
+
+alerts:
+  - name: Failure During Business Hours
+    condition: if_during_business_hours
+    message: |
+      Trigger {{ trigger.name }} has failed
```
tgsoverly, line 23: We could always do something like `{{ ... }}` for project references.  But `project(...)` would be more consistent.  The starting point will always need to be understood.  Relative to what point.

tgsoverly: I do like {{ }} because it stand out.

tgsoverly: - logic: assert_that( {{ Aggregated Line.props.marker.line.color[1] }} ).is_equal_to("#D25946")

PR #120, opened 2024-08-28
Rename to source

tgsoverly:
Well, here it is.

Renamed everything in the view, but not in the projects to ensure it is backward compatible.


visivo/models/defaults.py
```diff
@@ -1,30 +1,31 @@
 from typing import Optional
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, ConfigDict, Field
 
 
 class Defaults(BaseModel):
+    model_config = ConfigDict(populate_by_name=True)
```
tgsoverly, line 6: This allows for both `source` and `target` to be used.

jpjsonic (approved):
Looks good! I think sources is a lot cleaner

PR #121, opened 2024-08-28
Create alerts two point ohhhhh yea!

tgsoverly:
Well this is a whopper.

This PR does more than create Alerts 2.0, it also creates two more core concepts:
1. A `ContextString` a reference to some other object or property
2. An `EvalString` a logical statement that evaluates to true or false.

These two concepts allow for the tests and alerts to reference and be triggered by many other objects in the project DAG.

This PR also has some larger organization in the project and dag constructor classes.


visivo/models/base/eval_string.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def get_context_strings(self) -> List[ContextString]:
+        return list(
+            map(
+                lambda m: ContextString(m),
+                re.findall(INLINE_CONTEXT_STRING_REGEX, self.value),
+            )
+        )
+
+    def evaluate(self) -> Any:
```
jpjsonic, line 41: Is the idea here to make it safer than just running `eval()` around the string?

visivo/models/base/parent_model.py
```diff
@@ -153,19 +153,92 @@ def name_match(trace):
...
-
-        options = {}
-        pos = spring_layout(dag)
-        draw_networkx(dag, pos, **options)
-
-        ax = pyplot.gca()
-        ax.margins(0.20)
-        pyplot.axis("off")
-        pyplot.show()
+    def show_dag(self, dag=None):
```
jpjsonic, line 156: This is cool, are you thinking of including this chart in the viewer by default?

visivo/models/test.py
jpjsonic, line 1: For some reason the test model is no not rendering in the docs. I don't think we need to solve that in this PR but wanted to call that out.

visivo/models/base/parent_model.py
tgsoverly: Not yet, but eventually.  If we are using plotly, lets start going that way!

visivo/models/base/eval_string.py
tgsoverly: Yes, it can only execute permitted operations

visivo/parsers/evaluator.py
```diff
@@ -0,0 +1,143 @@
+from typing import Any
+import ast
+import operator as op
+import os
+
+SUPPORTED_NUMPY_FUNCTIONS = ["sum", "all", "mean"]
```
jpjsonic, line 6: @tgsoverly why not allow all numpy functions? Are you thinking that would be dangerous?

visivo/parsers/evaluator.py
```diff
@@ -0,0 +1,143 @@
...
+            raise ValueError(f"Unsupported node type: {node.__class__.__name__}")
+
+    def eval_subscripts(node):
+        nodes = build_list_of_nodes(node)
+        nodes = nodes[::-1]  # Reverse the order of nodes
+        if hasattr(nodes[0], "id") and nodes[0].id == "project":
+            value = get_object_from_node(project, 1, nodes)
+            return value
+        else:
+            raise ValueError(f"Unsupported node type: {nodes[0].id}")
```
jpjsonic, line 62: What's a node? Is that an object in the dag of your project?

visivo/commands/options.py
```diff
@@ -54,13 +54,6 @@ def source(function):
     return function
 
 
-def alert(function):
```
jpjsonic, line 57: We'll have to update some of our docs that use these options.

visivo/models/alert.py
```diff
@@ -1,156 +1,40 @@
...
-    alerts:
-      - name: email-destination #any unique name of your choosing
-        type: email
-        subject: "[ALERT] Your Visivo Tests Have Failed" #can be any message you want
-        to: someone@your_company.com
-        port: 2525 #is this port by default
-        host: your_company_email_server.com
-        username: someones_username
-        password: {% raw %}{{ env_var('EMAIL_PASSWORD')}}{% endraw %} #We'd recommend using environment variables here for security
+### Example
```
jpjsonic, line 13: ❤️

visivo/models/alert.py
```diff
@@ -1,156 +1,40 @@
...
+``` yaml
+alerts:
+    - name: Example Alert
+      if: >{ anyTestFailed() && env.ENVIRONMENT == "PRODUCTION" }
+      - destinations:
```
jpjsonic, line 18: I think maybe you wanted destinations to be a key and not a list?

visivo/models/base/context_string.py
```diff
@@ -0,0 +1,70 @@
...
+
+INLINE_REF_REGEX = r"\${\s*ref\(([a-zA-Z0-9\s'\"\-_]+?)\)[\.\d\w]*\s*}"
+INLINE_REF_PROPS_PATH_REGEX = r"\${\s*ref\([a-zA-Z0-9\s'\"\-_]+?\)([\.\d\w\]\[]*)\s*}"
+INLINE_PATH_REGEX = r"\${\s*([a-zA-Z0-9\s'\"\-_\.\[\]]+?)\s*}"
+
+
+class ContextString:
+    """
+    Represents a string that contains a reference to named object in the project model.
+    Currently you can reference another object by using the ref() function.  Example:
```
jpjsonic, line 15: Could be good to use [admonitions](https://squidfunk.github.io/mkdocs-material/reference/admonitions/) to call out the deprecation of the old `ref()` function
ex- 
``` 
!!! warning "<v1.037 behavior" 
    In previous versions of visivo you were able to reference other models using the naked `ref()` 
    function
```

visivo/models/base/eval_string.py
```diff
@@ -0,0 +1,83 @@
...
+
+"""
+EvalString represents a string that contains expressions to be evaluated.
+
+This class is used for handling strings that may contain references or expressions
+enclosed in '>{' and '}' delimiters. It provides functionality to validate the
+string format and extract embedded ContextString instances.
+
+Available injected methods:
+    - any_test_failed() -> bool
```
jpjsonic, line 17: I wonder if it makes sense to nest these in a `run` or `test` object that gets updated when you run visivo cli commands. Then you could reference it like `test.any_test_failed`

jpjsonic (commented):
Looks like you're making a ton of progress!

visivo/parsers/evaluator.py
tgsoverly: A node in this context is an node in the Abstract Syntax Tree (AST).  This is how language gets parsed, so I am finding what can be done in our EvalStrings

visivo/parsers/evaluator.py
tgsoverly: I want to make sure they work, and have a test for them.  Easy to add, we can add a bunch of them.

test-projects/integration/project.visivo.yml
```diff
@@ -163,7 +164,8 @@ dashboards:
                   order_by:
                     - query(x asc)
                   tests:
-                    - logic: assert_that(Double Simple Line.props.x[2]).is_equal_to(3)
+                    - assertions:
+                        - ">{ ${ref(Double Simple Line).props.x[2]} == 3 }"
```
jpjsonic, line 168: ❤️

jpjsonic (approved):
🦜

visivo/models/base/eval_string.py
tgsoverly: I am not sure I followed that.

visivo/models/base/context_string.py
tgsoverly: I like this.

PR #124, opened 2024-09-04
Tim fix default selection

tgsoverly:
Needed to fix this before doing the row selector


jpjsonic (approved):
Looks good! I think this might fix another bug I discovered tonight

jpjsonic:
I think maybe the issue that I'm running into occurs when you use indicator traces now. For some reason I can't get the selector to work with them.

PR #126, opened 2024-09-05
Enable selector for row and items

tgsoverly:
Allow rows to be shown or hidden with a selector.  

This selector is a bit of a change from the "trace" selector.  You don't need to have the selector also referenced on the chart, because the chart is actually selecting traces and traces can be used many times, and you wouldn't want to hide the same trace everywhere.  (Maybe we do what to change this, but that isn't in this PR)

To use it all you need to do is define a selector, and have the rows to show/hide as options in that trace.

https://github.com/user-attachments/assets/37f9f31b-f165-4f8b-a5a0-6bc72ee2bdec


viewer/src/components/Dashboard.js
```diff
@@ -87,10 +90,15 @@ const Dashboard = (props) => {
     }
 
     const renderRow = (row, rowIndex) => {
+        const selector = getSelectorByOptionName(project, row.name)
```
tgsoverly, line 93: Test this.

viewer/src/components/items/CohortSelector.js
```diff
@@ -0,0 +1,34 @@
+import Loading from "../Loading";
```
tgsoverly, line 1: Test and organize

viewer/src/components/items/NameSelector.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
```
tgsoverly, line 1: Test and organize

viewer/src/components/items/Selector.js
```diff
@@ -1,26 +1,38 @@
...
+import NameSelector from "./NameSelector";
 
 export const ChartContainer = tw.div`
     relative
 `;
 
 const Selector = ({ selector, project, itemWidth }) => {
-    const traceNames = selector.options.map((trace) => trace.name)
-    const tracesData = useTracesData(project.id, traceNames)
+    const isTraces = selector.options.every((option) => option.hasOwnProperty("props"))
```
tgsoverly, line 11: Test this line, it is probably broken.

viewer/src/models/Project.js
```diff
@@ -0,0 +1,19 @@
+export const findAllSelectors = (project) => {
```
tgsoverly, line 1: Test

visivo/models/project.py
```diff
@@ -149,6 +149,23 @@ def validate_dag(self):
 
         return self
 
+    @model_validator(mode="before")
+    def set_names_on_named_models(cls, values):
```
tgsoverly, line 153: Rename and test

visivo/parsers/serializer.py
```diff
@@ -36,31 +36,40 @@ def replace_item_ref(item):
...
                             )
                 if item.selector:
-                    item.selector = ParentModel.all_descendants_of_type(
+                    item.selector = ParentModel.first_descendant_of_type(
                         type=Selector, dag=dag, from_node=item, depth=1
-                    )[0]
-                    item.selector.options = ParentModel.all_descendants_of_type(
-                        type=Trace, dag=dag, from_node=item.selector, depth=1
                     )
+                    options = [
```
tgsoverly, line 67: Test

jpjsonic (approved):
Looks good to me once the tests go green 👍 

Also summarizing what we discussed over zoom yesterday- 

1. Add selectors for items as well as rows in this PR.
2. In a follow up PR create a dashboard section item that can house multiple rows, be referenced in an item and in a name selector.

PR #127, opened 2024-09-12
Small Tweaks, Bug Fixes and Docs Improvements

jpjsonic:


mkdocs/topics/sources.md
```diff
@@ -2,7 +2,7 @@
 Sources are data sources. They are databases that contain your data. This page describes some best practices how to set sources up. 
 
 !!! tip
-    There is specific documentation on how to connect to a variety of databases in the Configuration documentation ([ie. PostgresSQL](/../reference/configuration/Sources/PostgresqlSource/))
+    There is specific documentation on how to connect to a variety of databases in the Configuration documentation ([ie. PostgresSQL](/../reference/configuration/targets/PostgresqlSource/))
```
tgsoverly, line 5: Think this should be sources

tgsoverly (approved):

viewer/src/components/Dashboard.js
```diff
@@ -16,13 +16,19 @@ const Dashboard = (props) => {
...
+            return 128
+        } else if (height === 'small') {
             return 256
         } else if (height === 'medium') {
             return 396
-        } else {
+        } else if (height === 'large') {
             return 512
+        } else if (height === 'xlarge') {
+            return 792
```
tgsoverly, line 29: Nit: 768 is more binary!

mkdocs/topics/sources.md
jpjsonic: It should be sources. For some reason our parser is still labeling it as targets. I think maybe its coming from the aliasing? I didn't want to spend too much time digging into it, but it does bother me too

viewer/src/components/Dashboard.js
jpjsonic: Oh good call will update.

mkdocs/topics/sources.md
tgsoverly: Any chance it is the capitalization?

PR #131, opened 2024-09-15
Tim add dag viewer

tgsoverly:
A hidden and starting point to have a dag viewer
Screenshot 2024-09-17 at 2 13 44 PM


visivo/commands/serve_phase.py
```diff
@@ -36,13 +41,14 @@ def app_phase(output_dir, working_dir, default_source, name_filter, threads):
...
 
-    run_phase(
+    runner = run_phase(
         output_dir=output_dir,
         working_dir=working_dir,
         default_source=default_source,
         name_filter=name_filter,
         threads=threads,
     )
+    write_dag(project=runner.project, output_dir=output_dir)
```
jpjsonic, line 51: @tgsoverly are you planning on adding this to run and deploy in a separate PR so we can show it in the app?

tgsoverly: Yes, I would like to try it in serve for a bit first

PR #136, opened 2024-09-20
Add error to browser

tgsoverly:
Displays any command line error in the viewer.

Screenshot 2024-09-21 at 11 35 42 AM


tgsoverly:
https://linear.app/visivo/issue/VIS-480/console-errors-on-serve-show-up-in-browser

jpjsonic (approved):
I think this is a step in the right direction! 

I like the way the errors are summarized at the top - 
image

You do lose the error if you scroll down the dash - 

How much more work wou

https://github.com/user-attachments/assets/56a6e764-0029-40e5-84c0-618493875460

It would be really powerful to also overlay the _individual_ errors that are downstream from each item in the dashboard over the impacted items while still showing the old item behind the error ie - 
image
In the example above most of the items are impacted because the error is on a shared target.

tgsoverly:
An overlay is easy, but working it into the day is a larger lift.

PR #138, opened 2024-09-23
Add empty error

tgsoverly:
The dist didn't have a response for the error.  This is fine for now, but if we have other ones for static sites we should think about doing it differently.


jpjsonic (approved):
Thanks for the quick fix!

PR #149, opened 2024-10-19
Enable dbt sync on compile, run, and serve

tgsoverly:


jpjsonic:
@tgsoverly should we create this project as a fixture or under the `test-projects` directory?

tests/fixtures/dbt-unit-test-project/models/example/my_first_dbt_model.sql
```diff
@@ -13,7 +13,7 @@ with source_data as (
 
     select 1 as id
     union all
-    select null as id
+    select 2 as id
```
jpjsonic, line 16: @tgsoverly this fixes the dbt failure

test-projects/dbt/models/example/my_first_dbt_model.sql
```diff
@@ -0,0 +1,27 @@
...
+    Try changing "table" to "view" below
+*/
+
+{{ config(materialized='table') }}
+
+with source_data as (
+
+    select 1 as id
+    union all
+    select null as id
```
jpjsonic, line 16: @tgsoverly changing this to null fixes the dbt failure

visivo/commands/dbt_phase.py
```diff
@@ -0,0 +1,143 @@
...
+            raise click.ClickException(
+                f"Profile '{dbt_profile}' not found in profiles.yml"
+            )
+        if not dbt_target:
+            dbt_target = profiles[dbt_profile]["target"]
+
+        manifest_file = f"{dbt_root}/{dbt_project['target-path']}/manifest.json"
+        if not os.path.exists(manifest_file):
+            raise click.ClickException(
+                f"Manifest file not found at '{manifest_file}'. You might need to run dbt."
```
jpjsonic, line 124: It might be good to provide the command that they need to run.  
> f"Manifest file not found at '{manifest_file}'."+ "\n\tYou might need to run `dbt compile`."

I think in a follow up we should give people the option to include the `dbt compile` command in the dbt_phase every time automatically to avoid them getting out of sync. I think right now you could update your dbt project, not run any dbt commands and use an old manifest.json file

visivo/commands/dbt_phase.py
```diff
@@ -0,0 +1,143 @@
...
+                target_found = True
+            source = target_value
+            source["name"] = f"dbt_{profile_name}_{target_name}"
+            if "schema" in source:
+                source["db_schema"] = source.pop("schema")
+            if "user" in source:
+                source["username"] = source.pop("user")
+            if "type" in source and source["type"] == "snowflake":
+                Logger.instance().info(
+                    f"Configuring threads for snowflake is not supported.  Ignoring."
```
jpjsonic, line 21: Threads are now supported in snowflake.

visivo/commands/dbt_phase.py
```diff
@@ -0,0 +1,143 @@
...
+        for path, node in manifest["nodes"].items():
+            if path.startswith("model."):
+                names = []
+                if "database" in node:
+                    names.append(node["database"])
+                if "schema" in node:
+                    names.append(node["schema"])
+                names.append(node["name"])
+                model = {
+                    "name": f"dbt_{node['name']}",
```
jpjsonic, line 49: I think we should give the user the ability to set the prefix `dbt_` in the dbt block. That way we have clear documentation how this behavior and people can just set it to blank if they dont want a prefix. 

Maybe something like this? 
```
dbt: 
  enabled: true
  prefix: "dbt_" #The prefix that visivo will use to create the names of your models and sources in the visivo dag
```

jpjsonic (changes requested):
Looks great! Super cool to see this come to life after work us talking about this for the past few years 🦜 

I left a few comments and recommendations. Also found a bug with `visivo serve` that we should fix before merging (see video). Basically serve loops since it's watching the target folder. 

https://github.com/user-attachments/assets/a48e125a-f877-4f4b-8d59-8d5d02e6c722

visivo/commands/dbt_phase.py
tgsoverly: Oh yea

jpjsonic (approved):
It's working great!

PR #151, opened 2024-10-20
Jared vis 466 speed up deploy

jpjsonic:
Benchmark- test-projects/docs-examples takes `487 s` to upload with the synchronous deploy 

Current State With Changes: `60 s`


visivo/commands/deploy_phase.py
```diff
@@ -1,19 +1,117 @@
...
+from time import time
+from tenacity import retry, stop_after_attempt, wait_fixed
 from visivo.commands.utils import get_profile_file, get_profile_token
 from visivo.discovery.discover import Discover
 from visivo.logging.logger import Logger
 from visivo.parsers.serializer import Serializer
 from visivo.parsers.parser_factory import ParserFactory
 
+# Limit concurrent uploads to avoid overloading the API
+semaphore = asyncio.Semaphore(50)
```
tgsoverly, line 17: I would be curious what the performance improvement is with a lower number, mostly I would be surprised if there is a large difference between 10 and 50

visivo/commands/deploy_phase.py
```diff
@@ -1,19 +1,117 @@
...
 
 def deploy_phase(working_dir, user_dir, output_dir, stage, host):
     """
-    Sends the current version of your project, traces & data to app.visivo.io where it can be viewed by other users on your account. You must specify a stage when deploying a project. The stage allows multiple versions of your project to exist remotely. This is very useful for setting up different dev, CI and production environments.
+    Synchronous function to manage the deployment, including initiating asynchronous operations.
     """
-    profile_token = get_profile_token(get_profile_file(home_directory=user_dir))
+    deploy_start_time = time() 
+    # Retrieve profile token for authentication
+    Logger.instance().debug("Retrieving profile token...")
```
tgsoverly, line 108: 👍

visivo/commands/deploy_phase.py
```diff
@@ -1,19 +1,117 @@
...
 from visivo.commands.utils import get_profile_file, get_profile_token
 from visivo.discovery.discover import Discover
 from visivo.logging.logger import Logger
 from visivo.parsers.serializer import Serializer
 from visivo.parsers.parser_factory import ParserFactory
 
+# Limit concurrent uploads to avoid overloading the API
+semaphore = asyncio.Semaphore(50)
+
+@retry(stop=stop_after_attempt(3), wait=wait_fixed(2))
```
tgsoverly, line 19: I would really love a test for these functions, they don't have to be exhaustive, but I also don't want to accidentally break them.

visivo/commands/deploy_phase.py
jpjsonic: @tgsoverly At ten it's actually much slower - 251.3

visivo/commands/deploy_phase.py
jpjsonic: Yeah that's a good point. I'll add some tests for these

visivo/commands/deploy_phase.py
jpjsonic: I do wonder if we should be uploading more of the trace files in a single call rather than one call for every small file.

tgsoverly: Oh wow.

tgsoverly (approved):

PR #152, opened 2024-10-22
Fixed base to allow for arrays and lists

jpjsonic:


visivo/models/trace_props/bar.py
```diff
@@ -687,7 +687,7 @@ class Bar(TraceProps):
         None,
         description=""" string<br>Set several traces linked to the same position axis or matching axes to the same alignmentgroup. This controls whether bars compute their positional range dependently or independently. """,
     )
-    base: Optional[float | constr(pattern=INDEXED_STATEMENT_REGEX)] = Field(
+    base: Optional[float | constr(pattern=INDEXED_STATEMENT_REGEX) | constr(pattern=STATEMENT_REGEX) | List] = Field(
```
tgsoverly, line 690: Isn't this auto generated?

tgsoverly (approved):

visivo/models/trace_props/bar.py
jpjsonic: @tgsoverly It is auto generated. I just made this edit to the code to quickly get the attribute fixed for RWX. I made a second card to revisit and update the auto generation logic so that we get the categorical coordinate strings right - https://linear.app/visivo/issue/VIS-512/fix-number-or-categorical-coordinate-string-validation-in-trace-props

PR #153, opened 2024-10-24
Improve Snowflake Run & General Run Logging

jpjsonic:
- [x] Parallelization of snowflake trace query jobs 
- [x] Utilize SQLAlchemy for snowflake
- [x] Improve connector error logging
- [x] Missing Database Context issue

### DIFF
--- a/visivo/models/sources/snowflake_source.py
+++ b/visivo/models/sources/snowflake_source.py
@@ -42,7 +42,7 @@ class SnowflakeSource(SqlalchemySource):
         description="The access role that you want to use when running queries.",
     )
     timezone: Optional[str] = Field(
-        None,
+        'America/Los_Angeles',
         description="The timezone that you want to use by default when running queries.",
     )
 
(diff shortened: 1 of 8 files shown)


visivo/models/sources/snowflake_source.py
```diff
@@ -41,35 +41,38 @@ class SnowflakeSource(Source):
         None,
         description="The access role that you want to use when running queries.",
     )
+    timezone: Optional[str] = Field(
+        None,
+        description="The timezone that you want to use by default when running queries.",
+    )
 
     type: SnowflakeType
+    connection_pool_size: Optional[int] = Field(
```
tgsoverly, line 50: I thought this would be inherited from the parent class now.

tgsoverly (approved):

visivo/models/sources/snowflake_source.py
jpjsonic: I think we're still setting them on each individual source.

PR #154, opened 2024-10-24
Looking at some possible culprits of continued connection leaks

jpjsonic:


visivo/models/sources/sqlalchemy_source.py
```diff
@@ -49,7 +49,9 @@ def get_engine(self):
         if not self._engine:
             if hasattr(self, "connection_pool_size"):
                 self._engine = create_engine(
-                    self.url(), pool_size=self.connection_pool_size
+                    self.url(), 
+                    pool_size=self.connection_pool_size,
+                    max_overflow=0 # I was reading that this is set to 10 by default. This is likely the issue.
```
jpjsonic, line 54: https://docs.sqlalchemy.org/en/20/core/engines.html#sqlalchemy.create_engine.params.max_overflow

jpjsonic: Yeah, looks to be 10 by default. So that means we'd always be allowing at least 11 connections if someone set their pool size to 1.

visivo/models/sources/sqlalchemy_source.py
```diff
@@ -45,13 +47,12 @@ def get_connection(self):
 
     def get_engine(self):
         from sqlalchemy import create_engine
+        from visivo.logging.logger import Logger
 
         if not self._engine:
-            if hasattr(self, "connection_pool_size"):
```
jpjsonic, line 50: 🪦

jpjsonic (commented):
Looks great!

PR #155, opened 2024-10-27
Allow threads in defaults

tgsoverly:
Allows you to set the number of threads at a project level


jpjsonic (approved):
This is a good addition to defaults

PR #159, opened 2024-10-30
Markdown Improvements

jpjsonic:
- [x] Fix bug where text doesn't fit the whole item div 
- [x] Enable changing the alignment of the markdown
- [x] Enable more markdown options with plugins


visivo/models/item.py
```diff
@@ -37,6 +37,10 @@ class Item(NamedModel, ParentModel):
     markdown: Optional[str] = Field(
         None, description="Markdown text to include in the dashboard."
     )
+    align: Optional[Literal["left", "center", "right"]] = Field(
```
tgsoverly, line 40: Hmm, I ran into this issue before. I wonder if we have enough data to distinguish between a chart and table.  The markdown is currently "the" item, and chart and table are "inside" the item.  It would be great if we could move chart and table up a level and not have it be "item".  But that is outside this PR.

tgsoverly (approved):

visivo/models/item.py
jpjsonic: @tgsoverly I was thinking about this as I was writing the validator and tests for the feature. It does feel like we should be able to infer the type. I think the markdown should be it's own object that you can reference else where in the dag kinda like the tables and charts too. That will be really powerful once we introduce analysis that you can write in markdown files

jpjsonic: I'll make a card for us to further discuss this. I think there's a lot of architectural implications here

PR #163, opened 2024-11-10
Leverage DAG to improve job running developer experience

tgsoverly:
External functionality is the same, now, it is easier to add new jobs.


visivo/query/runner.py
```diff
@@ -60,73 +64,71 @@ def run(self):
...
+        failed_items = [result.item for result in self.failed_job_results]
+        successful_items = [result.item for result in self.successful_job_results]
+        for terminal_node in terminal_nodes:
+            descendants = all_descendants(dag=self.project_dag, from_node=terminal_node)
+            if terminal_node in successful_items or terminal_node in failed_items:
+                self.job_dag.remove_node(terminal_node)
+                continue
+            elif any(descendant in failed_items for descendant in descendants):
+                Logger.instance().info(
+                    f"Skipping job for {terminal_node} because it has a failed dependency"
```
jpjsonic, line 102: I like this. It's a good idea to print out the ones that we skip for clarity.

visivo/query/runner.py
```diff
@@ -60,73 +64,71 @@ def run(self):
...
+                )
+                self.job_dag.remove_node(terminal_node)
+                continue
+            elif self.name_filter and not family_tree_contains_named_node(
```
jpjsonic, line 106: So this is where we'd implement the dbt selector style syntax?

visivo/query/runner.py
```diff
@@ -60,73 +64,71 @@ def run(self):
...
-            name_filter=self.name_filter,
-        )
-        jobs = jobs + run_local_merge_jobs(
-            dag=self.dag,
-            output_dir=self.output_dir,
-            project=self.project,
-            name_filter=self.name_filter,
-        )
-        return jobs
+    def create_jobs_from_item(self, item: ParentModel):
```
jpjsonic, line 124: We should add tests to the DAG as well. I dont know if that makes sense to do here or somewhere else.

jpjsonic (approved):
Looks good 👍

visivo/query/runner.py
jpjsonic: The other thing we should consider adding to the dag is breaking out creating the engine pool. Would enable us to give the user better logs and understand where our run bottlenecks are.

visivo/query/runner.py
tgsoverly: Correct!

visivo/query/runner.py
tgsoverly: I didn't follow the comment around the pools.

tgsoverly: > We should add tests to the DAG as well. I dont know if that makes sense to do here or somewhere else.

This is a great idea, and I think that isn't a huge lift, but this one is getting long in the tooth.  I think most of the work there is how do we want to trigger the test run since it is part of the larger dag.  Do they run by default?  Do you have to explicitly say not to run them?

visivo/query/runner.py
```diff
@@ -60,73 +63,94 @@ def run(self):
...
+    def job_callback(self, future: Future):
+        job_result: JobResult = future.result(timeout=0.01)
+        if job_result.success:
+            Logger.instance().success(str(job_result.message))
+            self.successful_job_results.append(job_result)
+        else:
+            Logger.instance().error(str(job_result.message))
+            self.failed_job_results.append(job_result)
+
+    def create_job_dag(self):
```
tgsoverly, line 87: Per our conversation with Dan, we can leverage this dag by doing two smaller pieces of work:
1. Add a command to generate a dag: `visivo dag --jobs`.  This command would print a standard output of this or the project dag.
2. Allow you to run a singular job: `visivo run --job job_name`.  This command would bypass our dag and simply run the job for that item.

What can that do you ask?  On mint:
1. Run the dag command and then use dynamic tasks to generate a tree of tasks that duplicate our dag.  https://www.rwx.com/docs/mint/dynamic-tasks
2. Mint becomes a dag runner for Visivo

PR #164, opened 2024-11-12
Move ordering to trace names

tgsoverly:
Move the ordering of traces one level higher to have the trace names available.


jpjsonic (approved):
Will this sort apply to tables too?

tgsoverly:
> Will this sort apply to tables too?

Tables only do one trace at a time, so it shouldn't matter.  What were you thinking?

PR #165, opened 2024-11-14
Jared vis 536 color way attribute

jpjsonic:


visivo/models/color_palette.py
```diff
@@ -0,0 +1,129 @@
+from typing import Union, List, ClassVar, Dict
+from pydantic import BaseModel, Field, model_validator
+
+class ColorPalette(BaseModel):
+    palette: Union[List[str], str] = Field(
+        "Warm Earth Tones",
+        description="""string or list of colors<br>Sets the default trace colors. Can be either a list of colors or a string corresponding to a palette name."""
+    )
+
+    PREDEFINED_PALETTES: ClassVar[Dict[str, List[str]]] = {
```
tgsoverly, line 10: That is a lot of options...

tgsoverly (approved):

visivo/models/color_palette.py
jpjsonic: Yeah for sure. Going to make docs for them so that people can reference them and quickly see how the options look.

PR #169, opened 2024-12-01
Add connection job to the runner dag.

tgsoverly:
Add a new handler for connection tests.


jpjsonic (approved):
This is great! I thin this will be super helpful for people getting started. 

The only other thing we'll want to consider here is the overhead for checking the source. For example if we're only running one trace query against the source then checking the source connection is a redundant step. I dont think that will be the case with most databases. i think snowflake is maybe the only one right now where it might add overhead in that situation

PR #170, opened 2024-12-01
Convert to use the dag filter for compile and run

tgsoverly:
Changes the name filter to a dag filter with appropriate +'s

### DIFF
--- a/tests/models/test_dag.py
+++ b/tests/models/test_dag.py
@@ -1,5 +1,5 @@
 from visivo.models.base.named_model import NamedModel
-from visivo.models.dag import family_tree_contains_named_node
+from visivo.models.dag import family_tree_contains_named_node, filter_dag
 from tests.factories.model_factories import SqlModelFactory, TraceFactory
 import networkx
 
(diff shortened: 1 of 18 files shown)


tests/models/test_dag.py
```diff
@@ -20,3 +20,42 @@ def test_family_tree_contains_named_node():
...
+    model = Model(name="model")
+    child = Model(name="child")
+    grandchild = Model(name="grandchild")
+    dag.add_node(grandparent)
+    dag.add_edge(grandparent, parent)
+    dag.add_edge(parent, model)
+    dag.add_edge(model, child)
+    dag.add_edge(child, grandchild)
+
+    filtered_dag = filter_dag(dag, "1+model+1")
```
jpjsonic, line 39: 😄 this is exciting!

visivo/commands/options.py
```diff
@@ -55,11 +55,11 @@ def dist_dir(function):
     return function
 
 
-def name_filter(function):
+def dag_filter(function):
     click.option(
-        "-nf",
-        "--name-filter",
-        help="Run the command to only include the dag that includes the node with the given name",
+        "-df",
```
jpjsonic, line 60: I wonder if we should make this `--select` or `-sl` rather than dag filter? I think dag filter is more clear. Only reason we'd do select is because it's what dbt uses.

jpjsonic (approved):
This is fantastic!

visivo/commands/options.py
tgsoverly: Honestly happy to do either, small change.

PR #172, opened 2024-12-12
Added big query source

jpjsonic:


.mint/test_visivo.yml
```diff
@@ -115,6 +115,17 @@ tasks:
...
     filter: [test-projects/integration]
 
+  - key: test-bigquery
+    use: visivo-install
+    run: |
+      cd test-projects/integration
+      visivo run -s remote-bigquery
+    env:
+      STACKTRACE: true
+      BIGQUERY_BASE64_CREDENTIALS: ${{ secrets.BIGQUERY_BASE64_CREDENTIALS }}
```
tgsoverly, line 125: Looks good

test-projects/integration/traces.visivo.yml
```diff
@@ -99,7 +99,7 @@ traces:
       type: scatter3d
       x: query(x)
       y: query(y)
-      z: query( -2*(x%2)*x + x + y*(y-3))
+      z: query( -2*(x/16)*x + x + y*(y-3))
```
tgsoverly, line 102: Why did this change?

tests/models/sources/test_bigquery_source.py
```diff
@@ -0,0 +1,25 @@
...
+    data = {
+        "name": "source",
+        "database": "database",
+        "type": "bigquery",
+        "credentials_base64": "base64string", 
+        "project": "project-project"
+    }
+    source = BigQuerySource(**data)
+
+    assert "**********" in source.model_dump_json()
```
tgsoverly, line 25: You probably copied what I had but maybe we could assert that it has the key and then the stars to make sure that the password field is the one that has the stars.

visivo/models/sources/bigquery_source.py
```diff
@@ -0,0 +1,99 @@
...
+        description="The Google Cloud service account credentials JSON string base64 encoded. Turn your JSON into a base64 string in the command line with `python -m base64 < credentials.json > encoded.txt`. ",
+    )
+
+    database: Optional[str] = Field(
+        None,
+        description="The default BigQuery dataset to use for queries.",
+    )
+
+    type: BigQueryType
+    connection_pool_size: Optional[int] = Field(
```
tgsoverly, line 84: I thought the connection pool size was already in the sequel alchemy

tgsoverly (approved):
Looks good a few non-blocking comments

test-projects/integration/traces.visivo.yml
jpjsonic: Big query doesn't support the `%` as a modulus. The logic was arbitrary anyways so I switched it so we could easily add another integration test for bigquery

tests/models/sources/test_bigquery_source.py
jpjsonic: Yeah that's a good idea to get more specific.

visivo/models/sources/bigquery_source.py
jpjsonic: I dont think so, unless I'm missing something

PR #179, opened 2025-01-08
Fix alias on if fields

tgsoverly:
We were writing the old names during many processes.  Switched to write the correct name, but the field "if" is a bit special.


jpjsonic (approved):
👍

PR #182, opened 2025-01-19
Batch the calls to the server for deployment

tgsoverly:
Change to batch create and finish file uploads.


visivo/commands/deploy_phase.py
```diff
@@ -16,92 +17,140 @@
...
-async def upload_trace_data(trace, output_dir, form_headers, host, progress):
+async def create_trace_files(batch, form_headers, host, progress):
     """
     Asynchronously uploads trace data files.
     """
+    files = list(map(lambda trace: {"filename": f"{trace.name}.json"}, batch))
+    url = f"{host}/api/files/direct/start/"
     attempt.set(attempt.get(0) + 1)
-    async with semaphore:
+    async with semaphore_1:
```
jpjsonic, line 34: Just curious, why limit the concurrency to 1? That would make the operation synchronous effectively right?

jpjsonic (approved):
Looks good! I'm getting the same 139/135 exit error on my PR. Pretty odd.

visivo/commands/deploy_phase.py
tgsoverly: It would, because we are doing batches, that is pretty parallel, so instead of removing that code, I thought we could play with different values to create the fastest deployment.

PR #183, opened 2025-02-01
Migrate the query syntax to ?{}

tgsoverly:
Query Strings:
1. Add a select to the query `contents as key` before the query.
2. Take the key and substitute the results into the object after the query is run.


visivo/commands/aggregate.py
```diff
@@ -13,7 +13,5 @@ def aggregate(output_dir, json_file):
 
     Logger.instance().debug("Aggregating")
 
-    from visivo.query.runner import Runner
```
jpjsonic, line 16: Oh that's a good catch. It might be worth setting up some linting for unused imports. I've found a number of them hanging around in different files.

jpjsonic (approved):
This will still be compatible with the `column()` `query()` syntax for now right?

visivo/commands/aggregate.py
tgsoverly: I looked into this, and apparently, python is bad for catching this.  You can have imports that are unused, but still needed...  🤷

jpjsonic: Oh sure, that's definitely true in testing suites. I think our app doesn't use as much of that type of package magic. But I could see the packet magic making it difficult to figure out what's not required vs what is.

jpjsonic (changes requested):
need to fix surface chart rendering- from integration project 

image

jpjsonic:
Can also see this issue in simple contour 
image

tgsoverly:
> Looks good! This replaces `query()` with `?{}` right? `Column()` still functions the same as before right?

Correct.

visivo/logging/logger.py
```diff
@@ -32,10 +32,11 @@ def set_type(self, type: TypeEnum):
             self.echo = None
 
     def debug(self, message: str):
```
jpjsonic, line 34: If we want to make this refactor we might have to rework a lot of files. Right now we use debug as a temporary spinner that will not persist the message once the process finishes.

jpjsonic (approved):
Looks good! Left one comment on the logging change

PR #184, opened 2025-02-03
Serve GUI & Explorer

jpjsonic:
WIP
- [x] Updating / adding tests
- [x] Get CI running with playwright
- [x] Rip out viewer thumbnail generation JS

Done
- [x] JS & flask_api updates to support the Explorer page
- [x] Dashboard view JS updates & flask server updates to handle fetching the thumbnails
- [x] Refactor of CLI phases. Aiming to decouple different functionality & add bypass flags (ie --skip-compile)
- [x] Replace liveserve with custom implementation that enables us to serve the site and also run non-blocking jobs in the background and manage the websockets that tells the site to update when a job is done.
- [x] New job for thumbnail generation based on dashboard node
- [x] Refactored the run to use the serve_phase when thumbnails are needed.
- [x] Added a ton of time based logging to all of the CLI commands.
- [x] Visivo dist no longer performs a run. I want to decouple commands when they dont already share CLI options naturally.


visivo/commands/serve_phase.py
```diff
@@ -51,6 +51,84 @@ def app_phase(output_dir, working_dir, default_source, dag_filter, threads):
     )
     write_dag(project=runner.project, output_dir=output_dir)
 
+    @app.route("/data/explorer.json")
```
tgsoverly, line 54: We are going to want to start breaking the serve_phase up into endpoints.

viewer/src/components/Project.js
```diff
@@ -1,46 +1,400 @@
...
 import { Link } from 'react-router-dom';
 import Dashboard from "./Dashboard";
 import Loading from "./Loading";
 import Heading from "./styled/Heading";
 import { Container } from "./styled/Container";
+import { TextInput, Badge, Card, Tooltip } from 'flowbite-react';
+import { HiSearch, HiTemplate } from 'react-icons/hi';
+import html2canvas from 'html2canvas';
+
+// Thumbnail generation component
```
tgsoverly, line 11: Probably need to break up this into smaller components.

jpjsonic: Yup, definitely. Did that this morning.

visivo/query/jobs/run_thumbnail_job.py
```diff
@@ -0,0 +1,107 @@
...
+    from playwright.sync_api import sync_playwright
+    thumbnail_path = get_thumbnail_path(dashboard.name, output_dir)
+
+    with sync_playwright() as p:
+        browser = p.chromium.launch()
+        context = browser.new_context(viewport={'width': 1200, 'height': 750})
+        page = context.new_page()
+
+        # Navigate to dashboard
+        page.goto(f"file://{output_dir}/index.html#{dashboard.name}")
```
jpjsonic, line 35: WIP, unsure the best approach here. Thinking maybe borrow some things from visivo dist? I'd rather not run another flask app, but that's an option too.

visivo/query/jobs/run_thumbnail_job.py
```diff
@@ -0,0 +1,107 @@
...
+                full_path=thumbnail_path,
+            )
+            return JobResult(item=dashboard, success=True, message=success_message)
+
+        try:
+            thumbnail_path = generate_thumbnail(dashboard, output_dir, timeout_ms)
+        except Exception as e:
+            if "BrowserType.launch: Executable doesn't exist" in str(e):
+                Logger.instance().info("Installing Playwright browser...")
+                import subprocess #PR question: Is this the best way to do this? It works, but feels meh
```
jpjsonic, line 75: @tgsoverly Let me know what you think. Maybe there's a place that we should install this when visivo gets installed rather than here?

visivo/commands/parse_project_phase.py
jpjsonic, line 1: Broke this out so that we can get the project without doing the write steps in situtations where you dont want a full recompile or already have done one and just need the project.

visivo/server/flask_app.py
```diff
@@ -0,0 +1,215 @@
+import click
```
jpjsonic, line 1: @tgsoverly this used to be the app_phase in the serve_phase.py file. Moved it here and renamed

visivo/server/hot_reload_server.py
jpjsonic, line 1: @tgsoverly Custom implementation that replaces livereload and gives us a LOT more flexability. Makes the runner non-blocking so that you can immediately serve up files that may exist in your target directory.  

I do need to add error handling for when the files that it wants to read dont exist yet. That's a possibility if you deleted your target directory and then served without compiling. Probably could run a compile phase before kicking off the server to make it super safe.

visivo/commands/utils.py
```diff
@@ -23,34 +23,6 @@ def get_source_types():
     return [postgresql_type, mysql_type, sqlite_type, snowflake_type, bigquery_type, duckdb_type]
 
 
-def parse_project_file(working_dir, output_dir, default_source):
```
jpjsonic, line 26: @tgsoverly Moved into the parse_project_phase

visivo/commands/serve_phase.py
```diff
@@ -1,138 +1,104 @@
...
-            project_json["dashboards"] = dashboards
-        else:
-            raise click.ClickException(
-                f"Currently the serve command name filtering only supports filtering at the dashbaord level.  No dashboard with {dag_filter} found."
-            )
-
-    return project_json
-
-
-def app_phase(output_dir, working_dir, default_source, dag_filter, threads):
```
jpjsonic, line 37: Moved into server/flask_app.py

visivo/commands/serve_phase.py
```diff
@@ -1,138 +1,104 @@
...
-from flask import Flask, current_app, send_from_directory
-from livereload import Server
-from .run_phase import run_phase
-import datetime
-import importlib.resources as resources
-
-VIEWER_PATH = resources.files("visivo") / "viewer"
-
-
-def write_dag(project, output_dir):
```
jpjsonic, line 15: Moved into compile_phase

visivo/commands/run.py
```diff
@@ -1,41 +1,96 @@
...
         working_dir=working_dir,
-        dag_filter=dag_filter,
-        threads=threads,
-        dbt_profile=dbt_profile,
-        dbt_target=dbt_target,
+        output_dir=output_dir,
+        default_source=source,
     )
+
+    # Create temporary server for thumbnail generation if needed
```
jpjsonic, line 55: Spin up a quiet version of the server that runs the initial job and and then closes if there's thumbnails to generate.

visivo/commands/run.py
```diff
@@ -1,41 +1,96 @@
...
+        )
+        
+        # Start server in one-shot mode
+        server.serve(
+            host="0.0.0.0",
+            port=port,
+            on_server_ready=on_server_ready,
+            one_shot=True  # Server will shut down after initial run
+        )
+    else:
```
jpjsonic, line 79: @tgsoverly Run the run_phase as normal if there's no thumbnails. I'm debating removing this since the server really doesnt bring much overhead.

visivo/commands/dist_phase.py
jpjsonic, line 1: @tgsoverly removing the run component of the dist phase. I think it's cleaner if we just require a run before dist. I added error messages that will tell people that they need to do the run first so I dont think well run into confusion.

.mint/deploy_dashboard.yml
```diff
@@ -26,17 +26,24 @@ tasks:
   - key: github-cli
     call: github/install-cli 1.0.1
 
+  - key: install-playwright
```
tgsoverly, line 29: Isn't this installed with poetry since we use it?

test-projects/duckdb-concurrency/project.visivo.yml
```diff
@@ -2,6 +2,7 @@ path: project
 name: duckdb-concurrency
 defaults:
   source_name: Example Source
+  thumbnail_mode: none
```
tgsoverly, line 5: It is odd that the project has the value of thumbnail mode.  When would you want the project to dictate this?  Can it just be a command line argument?

tests/commands/test_compile_phase.py
```diff
@@ -119,3 +119,103 @@ def test_compile_csv_script_model_with_nested_local_merge_model():
...
+    
+    compile_phase(
+        default_source="source",
+        working_dir=working_dir,
+        output_dir=output_dir,
+        dag_filter=None
+    )
+    
+    # Verify explorer.json was created
+    assert os.path.exists(f"{output_dir}/explorer.json")
```
tgsoverly, line 182: Is a flattened project structure what you want to create?  Is there a specific explorer format that would be more useful?

tests/commands/test_dist.py
```diff
@@ -5,30 +5,99 @@
...
 runner = CliRunner()
 
+@pytest.fixture
+def output_dir():
+    return temp_folder()
 
-def test_dist():
-    output_dir = temp_folder()
-    project = ProjectFactory()
+@pytest.fixture
```
tgsoverly, line 17: Nice using fixtures.

viewer/src/api/project.js
```diff
@@ -6,3 +6,14 @@ export const fetchProject = async () => {
         return null
     }
 }
+
```
tgsoverly, line 9: This should be in an "explorer" file.

viewer/src/components/thumbnail/DashboardThumbnail.js
```diff
@@ -0,0 +1,117 @@
+import React, { useState, useEffect } from "react";
+import html2canvas from 'html2canvas';
+import Dashboard from "../Dashboard";
+
+const ASPECT_RATIO = 16/10;
+
+function DashboardThumbnail({ dashboard, project, onThumbnailGenerated }) {
```
tgsoverly, line 7: Okay this is going away.

viewer/src/components/QueryExplorer.js
```diff
@@ -0,0 +1,506 @@
...
+              )}
+            </div>
+          </Panel>
+        </RightPanel>
+      </MainContent>
+    </Container>
+  );
+};
+
+export default QueryExplorer;
```
tgsoverly, line 506: This is a pretty big component, it probably should be broken up into pieces.  The explorer is probably a good opportunity to start having global state.

viewer/src/components/dashboard/DashboardSection.js
```diff
@@ -0,0 +1,80 @@
+import React, { useState, useEffect } from 'react';
+import DashboardCard from './DashboardCard';
+import { HiChevronRight, HiInformationCircle } from 'react-icons/hi';
+import { Tooltip } from 'flowbite-react';
+
+const levelDescriptions = {
+  L0: "The most important dashboards and metrics for the organization.",
```
tgsoverly, line 7: Having hard coded level seems like a solution that might not work for everyone.  Is there a way to make this dynamic?

viewer/src/components/dashboard/FilterBar.js
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
+import { Badge } from 'flowbite-react';
+import { HiSearch } from 'react-icons/hi';
+
+function FilterBar({ searchTerm, setSearchTerm, selectedTags, setSelectedTags, availableTags, totalCount }) {
```
tgsoverly, line 5: If we use global state we can avoid passing down so many properties.

viewer/src/components/items/Table.js
```diff
@@ -39,15 +40,40 @@ const Table = ({ table, project, itemWidth, height, width }) => {
...
-    }, [selectedTableCohort, tracesData, table]);
+    }, [selectedTableCohort, tracesData, table, isDirectQueryResult]);
 
     useEffect(() => {
         if (selectedTableCohort && columns) {
+            // Handle trace-based queries
             setTableData(tableDataFromCohortData(selectedTableCohort.data, columns));
+        } else if (isDirectQueryResult) {
+            // Handle direct query results
+            setTableData(table.traces[0].data.map((row, index) => {
```
tgsoverly, line 64: Might be good to have this in a function like the above setTableData

viewer/src/services/queryService.js
```diff
@@ -0,0 +1,35 @@
+export const executeQuery = async (query, projectId, sourceName) => {
```
tgsoverly, line 1: Nice to break these into different api files describing what they do.  api/query and api/trace ..

visivo/commands/dist_phase.py
```diff
@@ -1,50 +1,54 @@
...
-        project_json = json.load(f)
-    created_at = (datetime.datetime.now().isoformat(),)
-    with open(f"{dist_dir}/data/project.json", "w") as f:
-        f.write(
-            json.dumps(
-                {
+    
+    # Copy dashboard thumbnails if they exist, otherwise create an empty folder
+    try:
+        thumbnail_dir = os.path.join(output_dir, "dashboard-thumbnails")
```
tgsoverly, line 18: Checking that all these new files land in the same place as the URL endpoints you have above?  Otherwise a dist being served will fail.

visivo/commands/run.py
tgsoverly: Not following?  What is the trade off?

visivo/commands/run_phase.py
```diff
@@ -8,23 +10,52 @@ def run_phase(
...
 ):
     from visivo.logging.logger import Logger
     from visivo.query.runner import Runner
-    from visivo.commands.compile_phase import compile_phase
+    from time import time
+    
+    if not server_url and thumbnail_mode == "all":
+        raise Exception("Thumbnail mode is set to 'all', but no server URL is provided. A running server is required to generate thumbnails.")
+    #Replace compile phase with parse project phase if skip_compile is True. Injects the project if it's available.
+    if project and skip_compile:
```
tgsoverly, line 25: I do struggle with the trade off between automatic chaining of commands.  Java has this in a tool called mvn and it is sometimes hard to know what phase runs other phases by default.  🤷

visivo/query/jobs/run_thumbnail_job.py
```diff
@@ -0,0 +1,132 @@
...
+        # URL encode the dashboard name to handle special characters
+        from urllib.parse import quote
+        encoded_dashboard_name = quote(dashboard.name)
+        # Navigate to dashboard
+        page.goto(f"{server_url}/{encoded_dashboard_name}")
+        
+        try:
+            # Wait for dashboard to load and render
+            page.wait_for_selector(".dashboard-row", timeout=timeout_ms)
+            page.wait_for_timeout(1000)  #TODO: Replace this with a wait based on the actual item loading
```
tgsoverly, line 41: Let me know if you need help with this.

visivo/query/jobs/run_thumbnail_job.py
tgsoverly: Yea it feels like you can do that when visivo gets install, playwright is already in the package.

visivo/server/hot_reload_server.py
```diff
@@ -0,0 +1,137 @@
...
+        self.app = app
+        self.watch_path = watch_path
+        self.ignore_patterns = ignore_patterns or []
+        self.observer = None
+        self.server_thread = None
+        self.stop_event = Event()
+        self.socketio = SocketIO(self.app, cors_allowed_origins="*")
+        
+        # Configure Flask logging based on STACKTRACE
+        if not os.environ.get('STACKTRACE'):
```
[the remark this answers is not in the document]
tgsoverly: Hmmm.  It might also be an option to implement the "debug" logger which could a command line.  Also good how it is, to limit scope.

tgsoverly (commented):
Tried to leave some comments to keep you from getting blocked.

visivo/server/hot_reload_server.py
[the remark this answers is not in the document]
jpjsonic: Are you suggesting that we change the STACKTRACE=true pattern to a CLI option that the user can pass into visivo? I do think that makes sense, but agree on limiting scope at this point.

visivo/commands/run.py
jpjsonic: Basically you only need to run the serve_phase during a run if you plan to generate images. So if you're running  `visivo run --thumbnail-mode none` then you don't need to spin up the server.

test-projects/duckdb-concurrency/project.visivo.yml
jpjsonic: I set it up to be similar to default source where your project can set a default but the CLI can override that default too with an options.py function. 

I thought this might be useful because with it in the project you can change the default behavior for different environments. We could pull it out from here.

tests/commands/test_compile_phase.py
jpjsonic: I think so. The flat project structure enables really easy retrieval of named objects. We can probably refactor/alter this in the future easily too

viewer/src/components/thumbnail/DashboardThumbnail.js
jpjsonic: Yup, getting rid of this now, tbd

visivo/commands/run_phase.py
jpjsonic: Yeah it does become a bit confusing 😅  I think there's still a bit of refactoring left to go to make our phases more self contained. We should probably map out how it works now and I think we'll get a clearer picture of what the ideal breakout of roles would be.

visivo/query/jobs/run_thumbnail_job.py
jpjsonic: @tgsoverly thinking I'd ship it with the one second wait and come back to it? If you know of an obvious solution though let's do that.

viewer/src/components/dashboard/DashboardSection.js
jpjsonic: Agreed, I was thinking we could add a pydantic model to sit below the Project called DashboardLayout where you could change the dashboard sections title and descriptions. 

I decided to leave that out of scope for now. but can definitely add it in pretty easily if you think it's important to ship it with that feature.

.mint/deploy_dashboard.yml
jpjsonic: Turns out it's a python PEP to not add hooks to python package installs so we will need to do the subprocess on the first run or make a new command `visivo deps`? that installs the playwright browser

visivo/commands/serve_phase.py
```diff
@@ -91,6 +175,57 @@ def viewer_file(path):
...
+            data = request.get_json()
+            if not data or "thumbnail" not in data:
+                return jsonify({"message": "No thumbnail data provided"}), 400
+
+            # Extract base64 data
+            thumbnail_data = data["thumbnail"].split(",")[1]
+            thumbnail_bytes = base64.b64decode(thumbnail_data)
+
+            # Save thumbnail using safe name
+            thumbnail_path = os.path.join(thumbnail_dir, f"{safe_name}.jpg")
```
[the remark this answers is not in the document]
jpjsonic: Out so scope

test-projects/duckdb-concurrency/project.visivo.yml
tgsoverly: Seems odd that a project could specify it doesn't want thumbnails.  Seems like a display concern and not a project concern.

visivo/query/jobs/run_thumbnail_job.py
tgsoverly: The 1 second can work, if it is always enough.  It would be pretty straight forward to add a "onDashboardLoadComplete" hook on the dashboard, that would render a div that could be read from the testing side and know it is complete.

jpjsonic: I think it's usually enough, but it will run the job faster if we take that approach of setting up a listener. That's how I had configured the JS beforing moving it python. I guess i could have kept those parts.

test-projects/duckdb-concurrency/project.visivo.yml
jpjsonic: Hmm, yeah maybe it does make sense to move this into the new `view` object

PR #187, opened 2025-02-13
Backend: Worksheet Tabs in the Explorer

jpjsonic:
Enable worksheet tabs so that user's can have multiple explorations and persist relevant state across sessions (page views & start/stop server). 

Backend:
- [x] Persisting explorations & tab state across sessions in sqlite repository. 
- [x] Work on multiple query tabs at once. 

[Frontend](https://github.com/visivo-io/visivo/pull/188):
- [x] User can run worksheet Crud operations in an intuitive way


visivo/commands/dbt_phase.py
```diff
@@ -140,10 +140,11 @@ def dbt_phase(working_dir, output_dir, dbt_profile, dbt_target):
...
         
-        models_written = ', '.join([model['name'] for model in models])
-        relative_output_file = os.path.relpath(output_file, working_dir)
-        Logger.instance().info(f"dbt base visivo models written to {relative_output_file}: {models_written}")
-        sources_written = ', '.join([source['name'] for source in sources])
-        Logger.instance().info(f"dbt base visivo sources written to {relative_output_file}: {sources_written}")
+        if os.environ.get("STACKTRACE"):
+            models_written = ', '.join([model['name'] for model in models])
+            relative_output_file = os.path.relpath(output_file, working_dir)
+            Logger.instance().info(f"dbt base visivo models written to {relative_output_file}: {models_written}")
```
tgsoverly, line 146: What if you add a `debug` to the logger, and that only prints if there is a DEBUG env variable?

jpjsonic: Oh that would be a nice refactor for quality of life.

tests/server/models/conftest.py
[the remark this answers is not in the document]
jpjsonic: It's a special pytest file that creates fixtures  https://stackoverflow.com/questions/34466027/what-is-conftest-py-for-in-pytest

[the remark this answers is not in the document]
jpjsonic: Probably only  need it in the models folder where the fixtures are shared

tgsoverly (approved):
I think the structure and testing is good here.

jpjsonic:
> I think the structure and testing is good here.

Do you see any scalability issues with it? My hope is that using the SQLite repositories+models will make it easier for us to build similar end points and models into the core when were ready. 

I guess the downside is that it kinda locks us into a structure where if we change the schema in a future version we could create breaking changes for users who upgrade.

tgsoverly:
We can probably write migrations for the data to convert it to new versions.

jpjsonic:
> We can probably write migrations for the data to convert it to new versions.

Should I store the schema version in the database then so we can be ready for migrations?

visivo/commands/parse_project_phase.py
```diff
@@ -1,21 +1,42 @@
...
+from time import time
+
+def parse_project_phase(working_dir, output_dir, default_source, dbt_profile, dbt_target):
+    
+    # Run and Track dbt phase
+    dbt_start = time()
+    Logger.instance().debug("    Running dbt phase...")
+    dbt_phase(working_dir, output_dir, dbt_profile, dbt_target)
+    dbt_duration = round(time() - dbt_start, 2)
+    if os.environ.get("STACKTRACE"):
```
tgsoverly, line 19: Still going to say to do the debug thing.

visivo/server/models/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .base import Base
+from .worksheet import WorksheetModel
+from .session_state import SessionStateModel
+from .result import ResultModel
+
+__all__ = ['Base', 'WorksheetModel', 'SessionStateModel', 'ResultModel']
```
tgsoverly, line 6: What does this do?

tgsoverly (approved):

visivo/server/models/__init__.py
jpjsonic: This makes it so that you can import them from the visivo.serve.models path rather than the specific files that they are in. We probably dont need to do this though.

visivo/commands/parse_project_phase.py
jpjsonic: Yeah, it will only persist in CI though. On the user's machine it will just flash while it's running.

PR #195, opened 2025-02-20
Add number formatting to table cells

tgsoverly:


viewer/src/components/items/Table.js
```diff
@@ -106,6 +106,8 @@ const Table = ({ table, project, itemWidth, height, width }) => {
                             {value}
                         </Markdown>
                     );
+                } else if (/^[\d.]+$/.test(`${value}`) && `${value}`.length < 18) {
```
jpjsonic, line 109: What's the thinking behind limiting the casting to strings under 18 characters?

jpjsonic (approved):
I think this works, but it does feel like we should probably just use the built in type casting in Material React Table rather than formatting the strings: https://www.material-react-table.com/docs/guides/best-practices#type-safety

Then people could configure number based filters.

PR #196, opened 2025-02-20
Get visivo dist working with thumbnails

jpjsonic:


viewer/src/components/Project.js
```diff
@@ -28,15 +28,27 @@ function Project(props) {
   const projectId = props.project?.id || props.project?.project_id;
   const dashboardNames = allDashboards.map(d => d.name);
 
+  // Add debug logging for query conditions
+  console.warn('Query conditions:', {
```
tgsoverly, line 32: Do you still need this logs?

jpjsonic: Oh good point, no

visivo/commands/dist_phase.py
```diff
@@ -34,6 +34,9 @@ def dist_phase(output_dir, dist_dir):
                     }
                 )
             )
+        with open(f"{dist_dir}/_redirects", "w") as f:
```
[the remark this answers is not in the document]
jpjsonic: yeah it makes it so that you can navigate directly to a dashboard without having to first to to the index page. Currently that doesnt work out of the box for netlify without this redirect

PR #204, opened 2025-03-20
Working refactor to have no project.json

tgsoverly:


jpjsonic (commented):
Is the primary change turning the app function into a class?

tgsoverly:
> Is the primary change turning the app function into a class? That does make more sense.

Yes that is the primary change, because then we can track the state of the project and other objects in the future like repositories.  I think I am going to do a quick follow and set up repos for each of the objects.

PR #208, opened 2025-03-26
Fix the dist by upgrading all the things

tgsoverly:
This PR does a hand full of things:
1. It removes CRA since it is now deprecated
2. Changes the build process to have a dist only viewer that doesn't have additional items.


viewer/.gitignore
```diff
@@ -5,6 +5,8 @@
 /.pnp
 .pnp.js
 
+src/Providers.jsx
```
jpjsonic, line 8: Ah that's cool, so providers is getting built by yarn?

viewer/src/index.css
```diff
@@ -1,6 +1,75 @@
-@tailwind base;
-@tailwind components;
-@tailwind utilities;
+@import 'tailwindcss';
+
+@plugin '@tailwindcss/typography';
+
+@theme {
```
jpjsonic, line 5: This is good, we needed this.

visivo/commands/dist_phase.py
```diff
@@ -37,22 +36,22 @@ def dist_phase(output_dir, dist_dir):
...
-            trace_name = os.path.basename(os.path.normpath(trace_dir))
-            if os.path.exists(f"{output_dir}/{trace_name}/data.json"):
-                os.makedirs(f"{dist_dir}/data/{trace_name}", exist_ok=True)
-                shutil.copyfile(
-                    f"{output_dir}/{trace_name}/data.json",
-                    f"{dist_dir}/data/{trace_name}/data.json",
-                )
-
-        shutil.copytree(VIEWER_PATH, dist_dir, dirs_exist_ok=True)
+        # with open(f"{dist_dir}/data/error.json", "w") as f:
```
jpjsonic, line 39: @tgsoverly did you mean to leave this commented out?

visivo/query/query_writer.py
```diff
@@ -10,7 +10,7 @@ def __init__(self, output_dir: str, trace: Trace, query_string: str):
         self.query_string = query_string
 
     def write(self):
-        trace_directory = f"{self.output_dir}/{self.trace.name}"
+        trace_directory = f"{self.output_dir}/traces/{self.trace.name}"
```
jpjsonic, line 13: Probably not for this PR, but we should remove writing the trace to the file & just make it a method on the trace object to return the query string.

jpjsonic (approved):
Looks good! This is great, It will be good to enabling building different things into the dist and the viewer.

PR #214, opened 2025-04-09
Fix missing thumbnails because there was no ID

tgsoverly:


visivo/commands/dist_phase.py
```diff
@@ -28,6 +28,7 @@ def dist_phase(output_dir, dist_dir):
             f.write(
                 json.dumps(
                     {
+                        "id": "id",
```
tgsoverly, line 31: This was it...

viewer/src/components/Project.jsx
```diff
@@ -61,8 +59,8 @@ function Project(props) {
 
       return Object.fromEntries(results.filter(Boolean));
     },
-    enabled: Boolean(projectId) && allDashboards.length > 0,
```
tgsoverly, line 64: Seemed like the docs indicated you should have the values that are in the keys.

visivo/commands/dist_phase.py
jpjsonic: Oh my god.

viewer/src/components/Project.jsx
jpjsonic: What do you mean?

jpjsonic (approved):
:shipit:

viewer/src/components/Project.jsx
tgsoverly: allDashboards were not part of the query trigger, so it could change an the query wouldn't reload or enable

PR #218, opened 2025-04-26
Refactor job runner to use DAG and filter

tgsoverly:
This PR reworks the job runner to work off the filter and the project DAG directly.

- [x] Optimize the filter to return the minimum number of DAGs
- [x] Have the server change compute the changed files and create a filter
- [x] Add default filter to run only dashboards if nothing is given


visivo/jobs/dag_runner.py
```diff
@@ -104,38 +98,6 @@ def job_callback(self, future: Future):
             Logger.instance().error(str(job_result.message))
             self.failed_job_results.append(job_result)
 
-    def create_job_dag(self):
```
jpjsonic, line 107: This is a nice deletion

visivo/models/models/csv_script_model.py
```diff
@@ -102,10 +102,6 @@ class CsvScriptModel(Model):
     args: List[str] = Field(
         description="An array of the variables that build your command to run."
     )
-    changed: Optional[bool] = Field(
```
jpjsonic, line 105: ❤️

visivo/models/trace.py
```diff
@@ -92,10 +92,6 @@ class Trace(NamedModel, ParentModel):
         alias="name",
         description="The unique name of the object across the entire project.",
     )
-    changed: Optional[bool] = Field(
```
jpjsonic, line 95: ❤️

visivo/query/query_writer.py
```diff
@@ -13,9 +13,5 @@ def write(self):
         trace_directory = f"{self.output_dir}/traces/{self.trace.name}"
```
jpjsonic, line 13: I think we should move all this into the trace now that we dont use it for the runner! Keep the query generation in memory in a method

jpjsonic: Excited for this.

jpjsonic (approved):
📈

visivo/query/query_writer.py
tgsoverly: I will add that to the larger card, and do it in a separate PR!

PR #223, opened 2025-05-08
Tim migrate to dag

tgsoverly:


jpjsonic:
> Two things are accomplished in this PR:
> 
> 1. Clean up the use of zustand stores to be slices of the same store
> 2. Move the generation of a DAG into a common location so it can be used in different places in the app.
> 3. Speed up DAG generation
> 4. Clean up the lineage view a bit more.
> 
>  Screen.Recording.2025-05-12.at.7.58.29.AM.mov

That's really cool

PR #226, opened 2025-05-13
Remove writing of the query files

tgsoverly:
Stop writing the query files and only generate them in the job when needed.

Also did a couple clean up items:
1. Removed ROR
2. moved our generated model sources into a "models" folder


visivo/models/models/csv_script_model.py
```diff
@@ -106,9 +106,10 @@ def sql(self):
         return f"select * from {self.table_name}"
 
     def get_duckdb_source(self, output_dir) -> DuckdbSource:
+        os.makedirs(f"{output_dir}/models", exist_ok=True)
         return DuckdbSource(
             name=f"model_{self.name}_generated_source",
-            database=f"{output_dir}/{self.name}.duckdb",
+            database=f"{output_dir}/models/{self.name}.duckdb",
```
jpjsonic, line 112: Oh yeah that's a big improvement

visivo/query/query_writer.py
jpjsonic, line 1: ❤️

visivo/jobs/run_trace_job.py
```diff
@@ -29,31 +31,41 @@ def action(trace, dag, output_dir):
...
-            else:
-                message = repr(e)
-            failure_message = format_message_failure(
-                details=f"Failed query for trace \033[4m{trace.name}\033[0m",
-                start_time=start_time,
-                full_path=trace_query_file,
-                error_msg=message,
-            )
-            return JobResult(item=trace, success=False, message=failure_message)
+    query_string = _get_query_string(trace, dag, output_dir)
```
jpjsonic, line 34: Do you think we should put these methods in the job or as a method on the trace? If it's a method on the trace we could get the trace query independent of running a job. That might be useful for compile time validation of the SQL independent of a run once we integrate SQLGlot. Not blocking,  just a thought

visivo/jobs/run_csv_script_job.py
```diff
@@ -1,8 +1,5 @@
 from visivo.logging.logger import Logger
-from visivo.models.base.parent_model import ParentModel
```
jpjsonic, line 2: There's gotta be some way we can bulk remove all of our excess imports.

jpjsonic (approved):
Looks good! Left some thoughts on where to put the sql generate logic, but nothing blocking

visivo/jobs/run_trace_job.py
tgsoverly: Well you need the dag to get the query because you need to write for the type of source it is pointing to.  This is more difficult in the trace only view.

visivo/jobs/run_csv_script_job.py
tgsoverly: I looked into it and there isn't a great way in python.

PR #228, opened 2025-05-14
Cast to string the table items

tgsoverly:
This fixed the project JSON return type error.  We could make it even better overtime, but this stops errors.


viewer/src/components/explorer/Explorer.jsx
```diff
@@ -261,104 +261,8 @@ const QueryExplorer = () => {
     }
   };
 
-  const executeQueryWithStats = React.useCallback(
```
tgsoverly, line 264: This was a duplicate of the one in the store that was called on click.

jpjsonic (approved):
:shipit:

PR #233, opened 2025-05-15
Fix deploy for traces not in dashboards

tgsoverly:


jpjsonic (approved):
Works on the empora project with the extra traces 👍

PR #238, opened 2025-05-30
Allow key for snowflake and restrict sources for schema

tgsoverly:
This PR allows specifing the key for connecting to snowflake.  It also did a bit of cleanup to only pass allowable params from the profile to our source definition.


tgsoverly:
Confirmed dbt uses the same format: https://github.com/datacoves/balboa/blob/main/automate/dbt/profiles.yml

PR #241, opened 2025-06-01
Add versions to dbt

tgsoverly:
Add dbt versions and test and a quality of live "list" command.


visivo_schema_new.json
jpjsonic, line 1: Did I accidentally add this?

tgsoverly: Nope

PR #242, opened 2025-06-02
fix: relax dependency constraints

danielgafni:
This PR relaxes dependency constraints in `pyproject.toml` to use `>=` specifiers instead of `^` and `=`. 

The previously used constraints were too restrictive and made `visivo` incompatible with many projects. 

I understand that you probably want to keep some of the constraints more tight than the initial version of the PR (ex: do you want to keep `pandas < 2.0` like your original `^` constraint did?). Let's review them and make the necessary changes case-by-case.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ write-schema-json = "visivo.generate_project_schema_json:write_schema_json"
 
 [tool.poetry.dependencies]
 python = " >=3.10,<3.13"
-pydantic = ">=2.9.0"
+pydantic = ">=2.9.0,<=2.9.2"
 requests = ">=2.28.2"
 Jinja2 = ">=3.1.2"
 PyYAML = ">=6.0"
--- a/poetry.lock
+++ b/poetry.lock
@@ -3963,4 +3963,4 @@ h11 = ">=0.9.0,<1"
 [metadata]
 lock-version = "2.1"
 python-versions = " >=3.10,<3.13"
-content-hash = "92beb07d5db707a1d8cc722e1df5f56139eac14c9ee503600ae98b8bc2eac17d"
+content-hash = "70df74dd660279dc1335977b79cced16be1d4e9d2a056820afa0ff5fdb791e08"


pyproject.toml
```diff
@@ -12,67 +12,67 @@ write-schema-json = "visivo.generate_project_schema_json:write_schema_json"
 
 [tool.poetry.dependencies]
 python = " >=3.10,<3.13"
-pydantic = ">=2.9.0,<=2.9.2"
```
tgsoverly, line 15: I believe we had an issue with >2.9.2

pyproject.toml
```diff
@@ -12,67 +12,67 @@ write-schema-json = "visivo.generate_project_schema_json:write_schema_json"
...
-playwright = "1.50.0"
-flask-socketio = "^5.5.1"
-jsonschema-rs = "^0.30.0"
+pydantic = ">=2.9.0"
+requests = ">=2.28.2"
+Jinja2 = ">=3.1.2"
+PyYAML = ">=6.0"
+sql-formatter = ">=0.6.2"
+click = ">=8.1.3"
+pandas = ">=1.5.3"
```
tgsoverly, line 21: We would love to remove pandas entirely, (outside this PR)

pyproject.toml
danielgafni: Understood. Can keep it this way for now but looks like a better long-term solution would be to resolve that issue in the first place

danielgafni:
Side note: I think you guys should setup optional dependencies (extras) for data sources. E.g. there is no reason to install `google-cloud-bigquery` fore non-bigquery users.

pyproject.toml
tgsoverly: @jpjsonic Do you remember what the issue was?

jpjsonic:
> Side note: I think you guys should setup optional dependencies (extras) for data sources. E.g. there is no reason to install `google-cloud-bigquery` fore non-bigquery users.

Yup, completely agree with this. It was fine when we started out with just a few sources, but is now getting to to a tipping point where it makes sense to move to as needed installs

pyproject.toml
tgsoverly: I went and searched slack and didn't find anything.

PR #250, opened 2025-06-04
Tweak the error message

tgsoverly:


visivo/models/models/csv_script_model.py
```diff
@@ -123,8 +123,8 @@ def insert_csv_to_duckdb(self, output_dir):
         import pandas
         import subprocess
 
-        process = subprocess.Popen(self.args, stdout=subprocess.PIPE)
```
tgsoverly, line 126: Move this inside the try for a better message

jpjsonic (approved):
👍

PR #252, opened 2025-06-04
fix: move non-runtime dependencies to dev group

tgsoverly:
Moved the work from #243 into this one.


pyproject.toml
```diff
@@ -12,63 +12,64 @@ write-schema-json = "visivo.generate_project_schema_json:write_schema_json"
...
 httpx = ">=0.27.2"
-aiofiles = ">=24.1.0"
-tenacity = ">=9.0.0"
+jsonschema-rs = ">=0.30.0"
+matplotlib = ">=3.7.1"
+networkx = ">=3.1"
+numpy = ">=1.26.0,<2.0.0"
+pandas = ">=1.5.3,<2.0.0"
+playwright = ">=1.50.0"
 plotly = ">=5.24.0"
```
jpjsonic, line 33: @tgsoverly I think we can remove plotly too right? We used it at one point for the dag chart but moved that to react-flow pure JS

jpjsonic (approved):
👍

pyproject.toml
tgsoverly: We still use it but only in dev.  Moved

PR #254, opened 2025-06-12
Remove playwright and build installers for install improvements

tgsoverly:
This PR moves the thumbnail generation to javascript to eliminate the need for playwright and builds bundled binaries for people so they don't need to have python installed


tests/models/test_chart_layout.py
```diff
@@ -1,5 +1,5 @@
 from visivo.models.color_palette import ColorPalette
-from ..factories.model_factories import ChartFactory
```
tgsoverly, line 2: Relative imports cause issues with packagers.

PR #259, opened 2025-06-17
Refactor API to allow dynamically loading URLs

tgsoverly:
Ability to create dist for non-root deployments


viewer/src/config/urls.js
```diff
@@ -0,0 +1,203 @@
+/**
+ * Centralized URL configuration system for Visivo viewer
+ * Supports different environments (local, dist) with configurable host and deployment root
+ */
+
+// URL patterns for different environments
+const URL_PATTERNS = {
```
jpjsonic, line 7: This is really nice.

viewer/src/contexts/URLContext.js
```diff
@@ -0,0 +1,68 @@
+import React, { createContext, useContext, useMemo } from 'react';
+import { createURLConfig } from '../config/urls';
```
jpjsonic, line 2: So this would theoretically allow us to write many dist projects to a single examples.visivo.io route right?

jpjsonic (approved):
Looks great! This will be a lot easier to manage the dist and local

PR #272, opened 2025-07-03
Fix Polars Type Casting

jpjsonic:


visivo/models/sources/sqlalchemy_source.py
```diff
@@ -29,7 +30,14 @@ def read_sql(self, query: str):
             results.close()
         # Convert to dict of columns for Polars
         if data:
+            data_dict = {}
             data_dict = {col: [row[i] for row in data] for i, col in enumerate(columns)}
```
tgsoverly, line 34: Would love a test if possible.

tgsoverly (approved):

PR #276, opened 2025-07-07
ray3du vis 364 link to object in dashboard

ray3du:
**Feature: Shareable Dashboard Items**

- Dashboard charts and markdown items are now shareable.
- On hover, a share icon appears on each chart and markdown block.
Screenshot from 2025-07-07 11-51-44
image

- On tables the share icon is always visible as part of table Toolbar
image

- Clicking the share icon copies the item’s unique URL to the clipboard, allowing easy sharing.
- When the shared URL is opened, the page automatically scrolls to the shared dashboard item for quick access.

**Feature: Scroll position Persistence**
Scroll position will persist across live-reloads and when switching between dashboards

**Feature: Table Overflow Breaks Item Div**

- The search input is replaced with a compact search icon on smaller screens.
Screenshot from 2025-07-07 11-49-27

- When the icon is clicked, the search input is revealed below the table toolbar, ensuring the toolbar remains clean and usable without overflow.
Screenshot from 2025-07-07 11-49-38

- This improves responsiveness and provides a better user experience on mobile and smaller devices.


viewer/src/components/items/Markdown.jsx
tgsoverly, line 1: Nice

viewer/src/stores/store.js
```diff
@@ -11,7 +11,10 @@ const useStore = create(
   devtools((...a) => ({
     ...createEditorSlice(...a),
     ...createExplorerSlice(...a),
-    ...createCommonSlice(...a),
+    ...persist(createCommonSlice, {
```
tgsoverly, line 14: Wonder if the persist should be part of the create common slice, to keep this clean.

viewer/src/components/project/Dashboard.jsx
```diff
@@ -1,17 +1,19 @@
...
 import { getSelectorByOptionName } from '../../models/Project';
-import remarkGfm from 'remark-gfm';
-import rehypeRaw from 'rehype-raw';
-import rehypeSanitize from 'rehype-sanitize';
+import useScrollToElementById from '../../hooks/useScrollToElementById';
+import Markdown from '../items/Markdown';
 
 const Dashboard = ({ project, dashboardName }) => {
   const [searchParams] = useSearchParams();
+  const elementId = searchParams.get('element_id');
```
tgsoverly, line 13: I did some work recently to store search params in the store.  Is element_id stored there too?

tgsoverly (approved):

viewer/src/components/project/Dashboard.jsx
ray3du: No, I only use it for scrolling to the element, so its just a URL param.

viewer/src/stores/store.js
ray3du: Yeah create createCommon should not be there, it was probably added by mistake during the merge

jpjsonic (approved):
Looks great!

PR #282, opened 2025-07-11
Claude idea for refactoring the docs to be more tutorial heavy

jpjsonic:
I think we'll end up closing this PR and doing something smaller, but it's got some interesting ideas in it for how can change our documentation to improve it's structure.


tgsoverly:
I do like the first section refactor that is more lead you through a process, rather than what we have.

PR #308, opened 2025-08-12
Add ev and football examples to the onboarding flow

tgsoverly:
This updates the onboarding card to include two new items.  
It also fixes the issues:
1. Thumbnails had the wrong url locally.
2. It would try to load the cloned project before it was there, causing empty charts and errors.

Screenshot 2025-08-15 at 8 25 48 AM


jpjsonic (approved):
Looks great! It does feel like we could clean up this UI a bit. I like the idea of continuing to sell the benefits, but I do think the info cards between the examples are crowding the page a bit. 

Image

PR #310, opened 2025-08-19
Fix loading example project when project directory is applied

ray3du:
### Summary

This PR enhances the `CoreParser` to properly resolve file paths for Python scripts and database connections with working directory used.

### Fixes / Changes

* Ensures `.py` script paths in `args` are updated relative to the YAML file path.
* Loads env created on the `working_dir` when loading example project.

Example rojects fail to locate Python scripts and databases correctly when run from different directories. This ensures portability and environment configurability.

### DIFF
--- a/visivo/parsers/core_parser.py
+++ b/visivo/parsers/core_parser.py
@@ -34,7 +34,7 @@ class CoreParser:
 
     def __merged_project_data(self):
         project_data = self.project_file_data()
-
+        self.__recursively_add_file_path(project_data, self.project_file)
         data_files = {}
         for file in self.files:
             if file == self.project_file:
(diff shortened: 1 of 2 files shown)


visivo/parsers/core_parser.py
```diff
@@ -34,7 +34,7 @@ def __build_project(self):
 
     def __merged_project_data(self):
         project_data = self.project_file_data()
-
+        self.__recursively_add_file_path(project_data, self.project_file)
```
tgsoverly, line 37: I wonder if there is a better way to accomplish this.  It feels brittle.  The issue makes sense, we don't find anything because it is in a folder.  Would it make sense to keep everything working make the change in a different location.  Maybe update the working dir or similar to keep it simple here.

PR #311, opened 2025-08-19
Jared semantic layer

jpjsonic:


visivo/jobs/extract_dimensions_job.py
```diff
@@ -0,0 +1,167 @@
...
+
+def extract_dimensions_for_model(model: Any, source: Any) -> None:
+    """
+    Extract dimensions for a model directly (called from within other jobs).
+
+    This function modifies the model in-place by adding _implicit_dimensions.
+    """
+    try:
+        # Skip if not a SQL-based model
+        if not isinstance(model, (SqlModel, CsvScriptModel, LocalMergeModel)):
```
tgsoverly, line 21: I think the job should only be created for things that apply.

visivo/jobs/extract_dimensions_job.py
```diff
@@ -0,0 +1,167 @@
...
+                    implicit_dimension = Dimension(
+                        name=column_name,
+                        expression=column_name,  # Simple column reference
+                        data_type=column_type,
+                        description=f"Auto-generated dimension for column '{column_name}' ({column_type})",
+                    )
+                    implicit_dimensions.append(implicit_dimension)
+
+            # Store the implicit dimensions on the model (in memory only)
+            if not hasattr(model, "_implicit_dimensions"):
```
tgsoverly, line 142: This feels wrong to store it on the model as a private item, I would then have to rerun to get the columns on every startup.  Should this be stored as a json in a "models" directory?

visivo/jobs/run_csv_script_job.py
```diff
@@ -15,10 +16,15 @@ def action(csv_script_model: CsvScriptModel, output_dir):
     try:
         start_time = time()
         csv_script_model.insert_csv_to_duckdb(output_dir=output_dir)
+
+        # Extract dimensions after successful data insertion
+        duckdb_source = csv_script_model.get_duckdb_source(output_dir=output_dir)
+        extract_dimensions_for_model(csv_script_model, duckdb_source)
```
tgsoverly, line 22: Should this be part of the dag and not hidden in a different job?

visivo/jobs/utils.py
```diff
@@ -0,0 +1,67 @@
...
+        The source for the model, or None if not found
+    """
+    if isinstance(model, CsvScriptModel):
+        return model.get_duckdb_source(output_dir=output_dir)
+    elif isinstance(model, LocalMergeModel):
+        return model.get_duckdb_source(output_dir=output_dir, dag=dag)
+    elif isinstance(model, SqlModel):
+        # Try model.source first, then look in DAG
+        if model.source:
+            # Check if source is a ContextString that needs resolution
```
tgsoverly, line 36: Side comment, but I generally don't find Claude comments useful, just doubles the reading.

visivo/models/dimension.py
```diff
@@ -0,0 +1,45 @@
...
+    Unlike metrics which are aggregates, dimensions are calculated for each row and can be used
+    in GROUP BY clauses or as filter conditions. They allow you to define reusable calculated
+    fields without repeating the logic in every query.
+
+    !!! example
+        ```yaml
+        models:
+          - name: orders
+            sql: SELECT * FROM orders_table
+            dimensions:
```
tgsoverly, line 19: Should we call our "columns" on insights dimensions too?

visivo/jobs/run_csv_script_job.py
jpjsonic: We should do something different. The reason it introduces a new concept that a node in the dag before now never had more than one job. Now csv scripts models have to run first and then we need to do the dimension job.

visivo/jobs/utils.py
jpjsonic: Yeah especially not for Python where it's already readable. Maybe I can update my Claude instructions to never use them

visivo/models/dimension.py
jpjsonic: Good question. Kinda leaning towards getting rid of the concept on insights? As long as we support cross model dimensions the column feels redundant

visivo/jobs/extract_dimensions_job.py
jpjsonic: Are you saying that only models that are used in dashboards should evaluate? Counter argument is that you would probably want all of the metrics and dimensions in the UI to be able to build with. Otherwise you have a chicken and egg problem for UI building

visivo/jobs/extract_dimensions_job.py
jpjsonic: Yeah that's cleaner. We're gonna want to store that info as a sqlglot compatible schema file because then we can do pre-db run validation even more effectively

PR #312, opened 2025-08-20
Tim add insights phase 1

tgsoverly:


visivo/query/insight_tokenizer.py
```diff
@@ -0,0 +1,499 @@
...
+        self.input_dependencies = set()  # All input names referenced
+        self.required_columns = set()  # Columns needed in pre-query
+        self.groupby_statements = set()  # GROUP BY expressions needed
+
+        # Initialize analysis
+        self._analyze_insight()
+
+    def tokenize(self) -> TokenizedInsight:
+        """Main entry point - returns tokenized insight with pre/post queries"""
+        pre_query = self._generate_pre_query()
```
tgsoverly, line 48: We need to be able to generate these two items very well, I don't think we need to store the tokenized values like we did for the trace tokenizer.

visivo/query/insight_tokenizer.py
```diff
@@ -0,0 +1,499 @@
+import re
```
tgsoverly, line 1: This file will be very different, the query generator will be very different using SQL glot.

PR #324, opened 2025-09-30
Added verbose option to run

jpjsonic:


visivo/commands/run.py
```diff
@@ -21,6 +22,7 @@
 @dbt_profile
 @dbt_target
 @skip_compile
+@verbose
```
tgsoverly, line 25: I think you can add this to the same place we put @version.

PR #325, opened 2025-10-06
Tim add insight to deploy

tgsoverly:
Upload the insights to the server.


jpjsonic (commented):
@tgsoverly We'll need to update this and merge it into main for the input and insight changes to work on deploy right?

PR #374, opened 2026-03-27
Better AI loops for Frontend Dev

jpjsonic:


tgsoverly (approved):
I really like the e2e tests that are possible here.  Not sure if I follow but are they included in our test suite by default?

PR #382, opened 2026-04-25
fix(viewer): unwrap indicator scalar slots from arrays (B13)

jpjsonic:
## Summary
Plotly's \`type: indicator\` traces expect **scalar** values for \`value\`, \`delta.reference\`, \`gauge.threshold.value\`, \`title.text\`, etc. The viewer's \`mapQueryResultsToProps\` always produces arrays (even from a 1-row query result), so indicator insights rendered the \`-\` placeholder instead of the actual number.

This PR adds \`unwrapIndicatorScalarProps\` and calls it from \`chartDataFromInsightData\` after props are built when \`type === 'indicator'\`. The unwrap walks a hardcoded list of scalar paths and replaces single-element arrays with their first element (or \`null\` for empty arrays).

Scalar paths covered:
- \`value\`
- \`delta.reference\`, \`delta.relative\`
- \`gauge.threshold.value\`, \`gauge.axis.range.0\`, \`gauge.axis.range.1\`
- \`title.text\`
- \`number.prefix\`, \`number.suffix\`

Diagnostic: \`specs/plan/v1-final-bugfixes/B13-indicator-value-array-wrapping.md\`

## Test plan
- [x] \`yarn test --testPathPattern=Insight.test\` — 40 passing (16 new)
- [x] \`yarn lint\` — clean
- [ ] Reviewer (manual): author an indicator insight with \`value: ?{MAX(...)}\` and verify the rendered number is not \`-\`.

Tests cover:
- \`unwrapIndicatorScalarProps\`: top-level / nested / numeric path segments / empty arrays / no-ops on scalars / null-safe
- \`chartDataFromInsightData\`: indicator unwrap, indicator + split, non-indicator types unaffected

## Empora coordination
Once merged, Empora can migrate KPI indicators back from the trace fallback (\`column(name)[0]\`) pattern to the insight pattern. Follow-up to PR #381 (B14 Decimal cast) — for indicators driven by decimal aggregates, both PRs together produce the right value.

This is **PR #4 of 11** for the v1 final bugfix punch list.

### DIFF
--- a/viewer/src/models/Insight.js
+++ b/viewer/src/models/Insight.js
@@ -309,6 +309,59 @@ function groupDataBySplitKey(data, splitKey) {
 }
 
 /**
+ * Plotly indicator-trace prop paths that expect a scalar (number or string)
+ * rather than an array. mapQueryResultsToProps always produces arrays, so when
+ * the insight type is `indicator` we must unwrap arrays at these paths to the
+ * first element. Plotly renders `-` (placeholder) when an indicator scalar
+ * slot receives an array. See B13 in specs/plan/v1-final-bugfixes/.
+ */
+const INDICATOR_SCALAR_PATHS = [
+  'value',
+  'delta.reference',
+  'delta.relative',
+  'gauge.threshold.value',
+  'gauge.axis.range.0',
+  'gauge.axis.range.1',
+  'title.text',
+  'number.prefix',
+  'number.suffix',
+];
+
+/**
+ * Walk a dotted path on `obj` and, if the leaf value is an array, replace it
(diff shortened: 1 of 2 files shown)


jpjsonic:
Closing — superseded by an in-progress schema-level fix. The hardcoded `INDICATOR_SCALAR_PATHS` list here is brittle and won't generalize beyond indicator. The replacement pair (core schema-generator + visivo grammar/runtime/validation) addresses B13 generically by extending the `query-string` def to accept `?{...}[N]` and `?{...}[a:b]` so any primitive Plotly slot can take a scalar from a query column, with sqlglot-driven type validation at compile time.

PR #386, opened 2026-04-25
fix(migrate): preserve column_defs blocks with cell-level styling (B06)

jpjsonic:
## Summary
The migrate command's table deprecation pass deleted entire \`column_defs\` blocks, silently dropping per-cell styling that has no equivalent in the 2.0 table widget:

- \`markdown: True\` (renders \`<a>\` tags / inline HTML in cells)
- \`justify\` / \`align\` (per-column alignment)
- \`header_name\` (custom column-header text — only safe to delete if the user also aliases the column in SQL, which we can't determine automatically)

Auto-deleting these blocks silently lost user-visible styling. Refuse to auto-delete when any of these flags appear in the block; the deprecation warning then guides the user to migrate manually.

\`column_defs\` blocks that only use \`key:\` (which the 2.0 widget can auto-recover from query data) continue to be auto-deleted.

Diagnostic: \`specs/plan/v1-final-bugfixes/B06-migrate-column-defs-loses-markdown-flag.md\`

## Test plan
- [x] \`poetry run pytest tests/models/deprecations/test_table_deprecation.py\` — 8 passing (5 new B06 tests + existing 3)
- [x] \`poetry run pytest tests/models/deprecations/\` — 58 passing (no regression)
- [x] \`poetry run black --check visivo/models/deprecations/table_deprecation.py\`

Tests cover:
- \`markdown: True\` block not auto-deleted
- \`justify\` block not auto-deleted
- \`align\` block not auto-deleted
- \`header_name\` block not auto-deleted
- Plain \`key:\`-only block still auto-deleted (regression check)

## Empora coordination
After merge, the \`deal-calls.visivo.yml:calls-missing-deal-associations\` table (and similar link-rendering tables) will surface as a deprecation warning rather than being silently mangled. Empora handles each manually as part of Phase 7.

## Future work
A 2.0 per-column directive that preserves \`markdown:\` rendering exists as a feature ask but is intentionally out of scope for this PR (it's a feature, not a bug fix).

This is **PR #8 of 11** for the v1 final bugfix punch list.

### DIFF
--- a/tests/models/deprecations/test_table_deprecation.py
+++ b/tests/models/deprecations/test_table_deprecation.py
@@ -60,3 +60,125 @@ tables:
             assert len(migrations) > 0
             column_def_migrations = [m for m in migrations if "column_defs" in m.old_text]
             assert len(column_def_migrations) > 0
+
+    # ---------------------------------------------------------------------
+    # B06: column_defs that carry cell-level styling MUST NOT be deleted
+    # automatically — auto-deletion drops user-visible markdown / justify /
+    # align / header_name flags that have no equivalent in the 2.0 widget.
+    # ---------------------------------------------------------------------
+
+    def test_column_defs_with_markdown_flag_not_deleted(self):
+        """Per-cell `markdown: True` (used for <a> link rendering) cannot
+        be expressed in a SQL alias. Don't auto-delete the block."""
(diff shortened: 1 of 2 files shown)


jpjsonic:
I actually dont think we should do this one. Maybe it can be a table level attribute

PR #379, opened 2026-04-25
fix(migrate): scope list_all_ymls_in_dir to project files (B01)

jpjsonic:
## Summary
- Replace unconditional `rglob` in `list_all_ymls_in_dir` with an `os.walk` that filters hidden dot-dirs and known vendored/build directories.
- Previously, `visivo migrate --apply` from a repo root would rewrite YAML inside `.git/`, `.venv/`, `.github/workflows/`, `.rwx/`, `dbt_packages/`, `node_modules/`, etc. — corrupting CI workflows and vendored deps.

Excluded names: `node_modules`, `dbt_packages`, `target`, `__pycache__`, `build`, `dist`, `venv`, plus any directory whose name starts with `.`.

Diagnostic: `specs/plan/v1-final-bugfixes/B01-migrate-scope-rglob.md`.

## Test plan
- [x] `poetry run pytest tests/test_utils.py` (5 new scope tests, 17 passing)
- [x] `poetry run pytest tests/models/deprecations/ tests/commands/` (108 passing — no regression)
- [x] `poetry run black --check visivo/utils.py tests/test_utils.py`
- [ ] Reviewer: dry-run `visivo migrate` against a project with `.venv/` / `dbt_packages/` and confirm those files no longer appear in the migration plan.

## Notes
This is **PR #1 of 11** for the v1 final bugfix punch list (Empora migration). PR #10 (B03+B04 silent-skip overhaul) and PR #11 (B02 jinja guards) modify `name_format_deprecation.py` and rebase on this branch — please merge this first to unblock that chain.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10, opened 2023-09-17
Add action to deploy the simple dashboard to app.visivo.io on every pull request.

tgsoverly:
End to end example on how to build and deploy a dashboard on every pull request.

PR #388, opened 2026-04-25
fix(migrate): log per-file errors instead of silent skip (B03, B04)

jpjsonic:
## Summary
The name-format deprecation checker walked YAML files in three passes (collect renames, rewrite refs, rewrite \`trace_name\` / \`insight_name\` / \`source_name\` / \`alert_name\`). Each pass wrapped its file loop in:

\`\`\`python
except Exception:
    continue
\`\`\`

which silently swallowed every error — \`UnicodeDecodeError\`, \`OSError\`, even unexpected \`re.error\`. So a single problematic file would silently drop out of one or more passes, and the user would see "migration looked complete but my refs aren't updated".

This PR introduces a shared \`_read_file_safe\` helper that:

1. Opens with explicit \`encoding="utf-8", errors="replace"\` so a stray non-UTF-8 byte becomes U+FFFD instead of aborting.
2. Catches \`OSError\` / \`UnicodeDecodeError\` / unexpected \`Exception\`, logs a warning naming the file and the underlying error, then returns \`None\` for the caller to continue past it.

The three passes now use the helper and continue past any unreadable file with a **visible warning** instead of silently dropping it.

Diagnostic files:
- \`specs/plan/v1-final-bugfixes/B03-migrate-cascade-silent-skip.md\`
- \`specs/plan/v1-final-bugfixes/B04-migrate-name-whole-file-skip.md\`

## Test plan
- [x] \`poetry run pytest tests/models/deprecations/test_name_format_deprecation.py\` — 19 passing (2 new B03/B04 tests + 17 existing)
- [x] \`poetry run pytest tests/models/deprecations/\` — 55 passing (no regression)
- [x] \`poetry run black --check visivo/models/deprecations/name_format_deprecation.py\`

New tests cover:
- Sibling file with non-UTF-8 bytes does not block migration of other files
- Patched safe-reader path simulates a per-file failure → cascade continues; visited file is observed via spy

## Stacking
**Base branch: \`bugfix/b01-migrate-file-scope\`** (PR #379). This PR shares the file with PR #379 only at the import line / scanner-call sites; no logical conflict, but it's stacked here so PR #11 (B02 jinja guard) can stack on top of it.

After PR #379 merges, rebase this branch on \`main\` and update the base.

This is **PR #10 of 11** for the v1 final bugfix punch list.

### DIFF
--- a/tests/models/deprecations/test_name_format_deprecation.py
+++ b/tests/models/deprecations/test_name_format_deprecation.py
@@ -230,3 +230,77 @@ sources:
 
             # Should have 0 migrations because there's no project.visivo.yml
             assert len(migrations) == 0
+
+
+# ---------------------------------------------------------------------------
+# B03 / B04: silent-skip overhaul. Unreadable files must not stop the
+# cascade from processing other files in the same project, and the
+# previous bare ``except Exception: continue`` must be replaced with a
+# logged warning.
+# ---------------------------------------------------------------------------
+
+
+class TestNameFormatCascadeErrorHandling:
+    """B03/B04 regression coverage."""
+
+    def test_undecodable_file_does_not_block_other_files(self, tmp_path):
+        """A sibling file with non-UTF-8 bytes must not stop the cascade
+        from migrating other files. The new ``errors='replace'`` UTF-8
+        reader handles the bad bytes gracefully."""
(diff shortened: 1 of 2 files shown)


jpjsonic (commented):
This is a pretty straight forward one.

tgsoverly (approved):

PR #390, opened 2026-04-26
feat: query-string slice grammar + runtime + type validation (B13/arrayOk)

jpjsonic:
## Summary
Visivo-side pair for **visivo-io/core#265**. The core PR extends the schema's \`query-string\` \$def to accept \`?{expr}[N|a:b]\` and adds a scalar-broadcast branch to \`data_array\` valTypes. This PR is the consumer:

- Vendored schemas (\`visivo/schema/*.schema.json\`) regenerated from the core branch.
- Grammar updated in \`visivo/query/patterns.py\` to mirror the schema (\`?P<slice>\` named group).
- \`InsightProps.extract_query_strings\` keeps returning \`(path, body)\` pairs (no behavior change for existing callers); two new methods \`extract_query_slices()\` and \`extract_query_strings_with_slices()\` surface the slice metadata.
- \`InsightQueryBuilder\` runs broad-class type validation on any single-index slice. SQL type inferred via sqlglot \`annotate_types\` against the union of dependent model schemas. Numeric vs string mismatches fail at compile time with a clean message; mixed / unknown cases pass through (per the design call: validate when easy, don't go beyond type class).
- \`run_insight_job\` writes \`props_slices\` into the insight JSON.
- Viewer \`Insight.js\` gains \`applySliceExpression\` (Python-style slice grammar in JS) + \`applyPropsSlices\` and applies each authored slice after \`mapQueryResultsToProps\`.

This **replaces** the closed PR #382 (which hardcoded \`INDICATOR_SCALAR_PATHS\` in the viewer — brittle and indicator-only). The schema-driven approach generalises: any primitive Plotly slot now accepts \`?{expr}[0]\` to bind a scalar, and any data_array slot now accepts \`x: 5\` (Plotly's broadcast).

### arrayOk investigation finding
The user's example was \`bar.marker.color\`. After running a Pydantic-level repro, that prop already accepts both primitive and array forms today. The actual scalar-vs-array bug is on \`data_array\` valType props (\`bar.x\`, \`bar.y\`, \`bar.customdata\`, error bars, etc.) — the schema rejected scalars, but Plotly broadcasts. **Core PR #265's \`data_array\` change fixes this** and is included here via the regenerated schemas.

Diagnostic: \`specs/plan/v1-final-bugfixes/B13-indicator-value-array-wrapping.md\`

## Test plan
- [x] Full pytest: \`poetry run pytest tests/ --ignore=tests/integration\` — **1697 passing**, 2 skipped, 5 xfailed, 1 xpassed
- [x] New tests: \`tests/query/test_query_string_grammar.py\` (15) + \`tests/query/insight/test_prop_type_validator.py\` (18)
- [x] Viewer: \`yarn test --testPathPattern=Insight.test\` — 44 passing (16 new for \`applySliceExpression\` + \`applyPropsSlices\` + \`chartDataFromInsightData with props_slices\`)
- [x] \`yarn lint\` — clean
- [x] \`poetry run black --check\` — clean

## Stacking & dependency
- **Depends on visivo-io/core#265** (the core PR ships the canonical schema generator changes; this PR includes the vendored output).
- **Supersedes** the closed visivo PR #382.

## Empora coordination
After merge:
- Empora can author \`value: ?{MAX(...)}[0]\` on indicator insights and have them render correctly with strict type validation.
- Single-bar charts can use \`x: 5\` directly.


visivo/models/props/insight_props.py
```diff
@@ -61,43 +61,75 @@ def extract_query_strings(self, prefix: str = "props") -> List[Tuple[str, str]]:
...
+
+            >>> props = InsightProps(type="indicator", value="?{MAX(x)}[0]")
+            >>> props.extract_query_strings()
+            [('props.value', 'MAX(x)')]
+            >>> props.extract_query_slices()
+            {'props.value': '[0]'}
+        """
+        return [(path, body) for path, body, _slice in self._iter_query_strings(prefix)]
+
+    def extract_query_slices(self, prefix: str = "props") -> dict:
```
tgsoverly, line 85: This only show up in the props?  Feels like a higher concern.

tgsoverly (approved):

visivo/models/props/insight_props.py
jpjsonic: That's a fair point. It does feel like we might use them somehwere else eventually. Maybe in chart.layout, but right now insightProps (once we get rid of traces) are only place you might see the slices like this occur

PR #403, opened 2026-04-30
Migrate test projects from traces to insights

tgsoverly:
## Summary

Migrates the test-project YAML files from the legacy `traces` shape to the new `insights` shape. No production code changes — these are all `test-projects/**/*.visivo.yml` updates.

## Commits

- Migrate **duckdb-concurrency, homepage-assets, demo** from traces to insights
- Migrate **complex-project and integration** from traces to insights
- Migrate **docs-examples** (125 files) from traces to insights
- Fix mismatched names + collisions surfaced by `visivo run` after the migration

The storage / thumbnail code changes are split into a separate PR (#404) so this one can be reviewed as data-only.

## Test plan

- [x] `poetry run pytest tests/` — 1722 passed
- [ ] `visivo run` against each migrated test project — verify no DAG resolution or query errors
- [ ] Spot-check a few dashboards visually in the sandbox

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #404, opened 2026-05-01
Backend storage by name + dashboard thumbnail capture-on-view

tgsoverly:
## Summary

Two related backend/frontend cleanups, split out from the longer `tim-migrate-projects-to-insights` branch so the YAML migration (#403) can be reviewed independently.

### 1. Backend file storage by name (was hash)

Insights, inputs, model parquets, and dashboard thumbnails are now stored under their **clean names** on disk instead of `alpha_hash(name)`. Traces and source schemas already used names; this brings the rest in line.

- `target/<run_id>/insights/<name>.json` (was `<hash>.json`)
- `target/<run_id>/inputs/<name>.json`
- `target/<run_id>/files/<name>.parquet` for models, `<name>_<key>.parquet` for inputs
- `target/<run_id>/schemas/<name>/schema.json` (model schemas were under singular `schema/` — now plural to match source schemas)
- `target/dashboards/<name>.png`

`name_hash` (alpha-hash) **stays** in metadata responses — it's still the DuckDB SQL table identifier on the frontend (`useInsightsData`, `useInputsData`, `usePivotData`). This refactor is purely the storage layer.

**Drive-by bug fix**: `DashboardCard.jsx` previously computed `md5(name)` for the thumbnail URL while `dashboard_views.py` checked `alpha_hash(name)` — the two never matched, so the metadata endpoint always reported the thumbnail as missing and every page refresh re-triggered generation. Both endpoints now use the clean name; this collapses the mismatch and fixes "second load isn't instant."

Also writes `schema.json` for `CsvScriptModel` and `LocalMergeModel` (they were missing it).

### 2. Dashboard thumbnail: capture-on-view (replaces offscreen queue)

Previously the `/project` cards page tried to bulk-generate thumbnails by mounting every dashboard offscreen. That hit shared DuckDB-wasm catalog races, queue wedges, slot leaks (external dashboards), and stale-state captures (loading spinners).

The new flow is much simpler: when the user actually opens a dashboard, the live `<Dashboard>` is already rendered in the browser. A small helper waits for the above-fold spinners to clear (max 5s), runs html2canvas-pro on the rendered element, downscales to 800x500, and POSTs to `/api/dashboards/<name>.png/`. The `/project` cards page just shows whatever's on disk — no generation logic at all.

Net diff in the thumbnail subsystem: **−550 / +30** lines of removal, plus a focused ~120-line `captureDashboardThumbnail.js` helper.

- Deleted `viewer/src/components/project/DashboardThumbnail.jsx` (offscreen render)
- Deleted `viewer/src/components/project/thumbnailQueue.js` (slot/queue plumbing)
- Simplified `DashboardCard.jsx` (no slot acquire, no `<DashboardThumbnail>` mount, no `md5`)
- Added `captureDashboardThumbnail.js` + capture-on-view `useEffect` in `Dashboard.jsx`

## Test plan

- [x] `poetry run pytest tests/` — 1722 passed
- [x] `poetry run black --check visivo/` — clean
- [x] `cd viewer && yarn lint && yarn test` — 1555 passed, eslint clean
- [x] Sandbox round-trip:
  - `rm -rf test-projects/docs-examples/target/`
  - `bash scripts/sandbox.sh restart`
  - Visit `/project/annotation-and-shapes` → POST fires, file saved as `annotation-and-shapes.png`
  - Reload `/project` → 0 POSTs, cached thumbnail loads instantly
- [x] Big-dashboard case (`color-palettes-in-visivo`, 5900px tall, 27 charts): capture lands at t+6s instead of timing out; PNG renders sharp.

## Migration notes

- Existing `target/` directories will have orphaned hash-named files. Easiest path is to delete the run-id subdirectories under `target/` once and let the next `visivo run` regenerate them.
- The `dist_phase.py` static-mode writers also moved to name-based filenames (matching server mode). Trace URLs in dist mode still use hashes to avoid changing the static URL contract — called out as a future follow-up.


tests/commands/test_deploy_phase.py
```diff
@@ -194,14 +194,14 @@ def test_deploy_with_insights_and_inputs_success(requests_mock, httpx_mock, caps
...
-            "name": f"{insight.name_hash()}.json",
+            "name": f"{insight.name}.json",
             "id": "id4",
             "upload_url": "http://google/upload/id4",
         },
     ]
     input_file_starts = [
         {
-            "name": f"{input_obj.name_hash()}.json",
+            "name": f"{input_obj.name}.json",
```
jpjsonic, line 204: Nice, this will be simpler

viewer/src/components/project/Dashboard.jsx
```diff
@@ -127,6 +129,22 @@ const Dashboard = ({ project, dashboardName }) => {
     throwError(`Dashboard with name ${dashboardName} not found.`, 404);
   }
 
+  // Capture-on-view: once the user actually opens a dashboard and it has
+  // finished rendering, snapshot it for the cards listing. This replaces the
+  // old offscreen-render queue that tried to bulk-generate thumbnails for
+  // every dashboard at once. Only runs when there's no existing thumbnail.
```
jpjsonic, line 135: Nice, this a better process

viewer/src/components/project/Dashboard.jsx
```diff
@@ -127,6 +129,22 @@ const Dashboard = ({ project, dashboardName }) => {
...
+  useEffect(() => {
+    let cancelled = false;
+    captureDashboardThumbnail({
+      dashboardName,
+      getElement: () => dashboardRootRef.current,
+      isCancelled: () => cancelled,
+    });
+    return () => {
+      cancelled = true;
```
jpjsonic, line 144: Does this account for a quick page load where the charts dont render before the user clicks away?

jpjsonic (approved):
This is great! This going to make navigation so much smoother

viewer/src/components/project/Dashboard.jsx
tgsoverly: It just doesn't do anything.

PR #412, opened 2026-05-07
fix(snowflake): resolve ${env.X} on account/warehouse/role/timezone (B5)

jpjsonic:
## Summary
- Type `account` / `warehouse` / `role` / `timezone` as `Optional[StringOrEnvVar]` so the parser wraps `${env.X}` in `EnvVarString` (matches `private_key_path` / `private_key_passphrase`).
- Add `get_account` / `get_warehouse` / `get_role` / `get_timezone` resolver methods mirroring the `get_username` / `get_password` / `get_host` / `get_database` pattern on the base `Source` class.
- Route `url()` through the new getters so resolution is encoded once per field.
- Regenerate `visivo_project_schema.json` to reflect the field-type change.

## Bug
Every Snowflake query failed with:
\`\`\`
(snowflake.connector.errors.HttpError) 290404 (08001):
404 Not Found: post \${env.SNOWFLAKE_ACCOUNT}.snowflakecomputing.com:443/session/v1/login-request
\`\`\`
when \`account\`/\`warehouse\`/\`role\` were YAML-set with \`\${env.X}\` substitutions. Compile passed (the field types accepted the literal string), so the bug only triggered on the actual connect.

Two distinct issues, both required:
1. Field types \`Optional[str]\` kept \`\${env.X}\` as a literal string (not wrapped as \`EnvVarString\`).
2. \`url()\` read \`self.account\` / \`self.warehouse\` / \`self.role\` / \`self.timezone\` directly without going through \`_resolve_field\` — so even after wrapping, resolution would have been skipped.

\`username\` / \`password\` / \`database\` / \`host\` already worked because they're typed \`StringOrEnvVar\` AND go through the base-class getters in \`url()\`.

## Affected versions
- **Broken:** 1.0.79 → 2.0.0 (every version since \`\${env.X}\` was introduced).
- Pre-2.0 had \`{{ env_var(...) }}\` jinja workaround. 2.0 removed jinja, so this is the only path.

## Test plan
- [x] \`poetry run pytest tests/models/sources/test_snowflake_source.py\` — 19/19 pass (8 new parametrized + getter tests).
- [x] Full suite: \`poetry run pytest tests/\` — 1555 passed.
- [x] Black clean: \`poetry run black --check --target-version py312 visivo/models/sources/snowflake_source.py tests/...\`.
- [x] Schema regen: \`python -m visivo.generate_project_schema_json\` — schema reflects \`oneOf: [string]\` (StringOrEnvVar) on all four fields.
- [ ] Manual smoke against a real Snowflake project with \`\${env.X}\` on every connection field. (Empora can verify after dropping their local \`.venv\` patch.)


tgsoverly (approved):
Nice

PR #413, opened 2026-05-07
fix(server): /api/files/ falls back to alpha_hash(stem) lookup (B16)

jpjsonic:
## Summary
- Patch `serve_file_data_by_hash_with_run` in `visivo/server/views/file_views.py` so a 404 from the direct `<hash>.parquet` match triggers a fallback scan that serves the parquet whose stem hashes to the requested hash.
- Mapping is cached per-`run_id` with a thread lock; cache refreshes on miss so newly-written parquets are picked up without restarting the server.
- 9 new tests in `tests/server/views/test_file_views.py`.

## Bug
Every dashboard table built with the 2.0 `${ref(model).column} as Header` syntax 404'd its data fetch:
\`\`\`
404 GET /api/files/<32-char-alpha-hash>/main/
\`\`\`

Three components disagreed on filename convention post-1.0.82:
| Component | Behavior |
|---|---|
| Parquet writer (`jobs/parquet_io.py`) | Writes `<output_dir>/<run_id>/files/<clean_name>.parquet` |
| Insight URL builder (`insight_jobs_views.py`) | Reads filename from disk → embeds clean name in URL ✓ |
| Table data URL (viewer JS) | Computes `alpha_hash(model_name)` → asks for `<hash>.parquet` ✗ |
| Flask route (`file_views.py`) | Exact-match only on `<hash>.parquet` → 404 ✗ |

The route was correct before 1.0.82 (matched the writer's hash filenames). The 1.0.82 commit that switched the writer to clean names didn't update the table path on either the frontend or the route handler.

## Fix approach
Backend-only fallback (Option A from the spec). When `<hash>.parquet` doesn't exist, the route scans the run's files dir and matches by `alpha_hash(stem) == hash`. This:

- Lands as a one-file change on the request path (low blast radius).
- Is fully backwards-compatible — legacy hash-named parquets still serve via the direct lookup.
- Unblocks every consumer immediately without requiring a viewer rebuild.

The longer-term Option B (move table data URLs to clean-stem on the frontend, matching insights) can ship in a follow-up. The route fallback is cheap insurance against future divergence either way.

## Cache behavior
- Per-`run_id` `{hash → filename}` mapping built lazily on first request for that run.
- Refreshed on miss so a freshly-written parquet is picked up immediately.
- Thread-locked because Flask serves concurrently.
- O(1) amortized after warm-up; an O(N) scan over `files/` only runs on the first request and on cache-misses.

## Affected versions
- **Broken:** 1.0.82, 2.0.0
- **Working:** 1.0.81 and earlier (hash-named parquets — direct match wins)

## Test plan
- [x] `poetry run pytest tests/server/views/test_file_views.py` — 9/9 pass.
- [x] Full suite: `poetry run pytest tests/` — 1555 passed.
- [x] Black clean.
- [ ] Manual smoke: spin up an empora-dbt-style project with `${ref(model).column} as Header` tables, drop the local `.venv` patch, confirm tables render rows.
- [ ] Backwards-compat smoke: a project on 1.0.81 → 2.0.x upgrade with mixed hash and clean-name parquets in `target/` (e.g. user did `visivo run` on 1.0.81 then upgraded without `rm -rf target/`) still serves both via the route fallback. (Covered by `test_mixed_hash_and_clean_names_in_same_run`.)

### DIFF
--- a/tests/server/views/test_file_views.py
+++ b/tests/server/views/test_file_views.py
@@ -1,0 +1,174 @@
+"""Tests for /api/files/<hash>/<run_id>/ — covers the B16 hash-of-stem fallback.
+
+B16: post-1.0.82 the parquet writer started naming files with the model's
+clean name (e.g. `metrics_table.parquet`) instead of `<alpha_hash>.parquet`.
+The frontend still computes `alpha_hash(model_name)` client-side, so every
+table data fetch hits a 404 unless the route falls back to scanning the dir
+and matching by `alpha_hash(stem)`.
+"""
+
+import os
+import shutil
+import tempfile
+
+import pytest
+from flask import Flask
+
+from visivo.models.base.named_model import alpha_hash
+from visivo.server.views.file_views import register_file_views
+
+
+# Smallest possible parquet payload: real PAR1 magic bytes + footer. We don't
+# care about the parquet contents in these tests — only that send_file streams
+# the file we expect — so a stub byte string with the magic header is fine for
+# asserting "the right file was served."
+def _write_stub_parquet(path, payload=b"PAR1stub-payload-PAR1"):
+    os.makedirs(os.path.dirname(path), exist_ok=True)
+    with open(path, "wb") as f:
+        f.write(payload)
+
+
+@pytest.fixture
+def output_dir():
+    tmp = tempfile.mkdtemp()
+    yield tmp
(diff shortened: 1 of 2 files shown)


tgsoverly (approved):
Can we remove this eventually?

PR #414, opened 2026-05-07
New onboarding flow

jpjsonic:
It's what the people want


viewer/src/components/explorerNew/SQLEditor.jsx
```diff
@@ -106,6 +106,10 @@ const SQLEditor = ({
 
     setShowError(true);
     executeQuery(sourceName, queryText.trim());
+    // Tap for the onboarding "Run the query" sub-step. Lazy import.
+    import('../onboarding/onboardingState').then(({ recordOnboardingAction }) => {
```
tgsoverly: We get any performance gains from this?

viewer/src/components/onboarding/onboardingApi.js
```diff
@@ -0,0 +1,75 @@
+/* Thin wrappers over the existing Flask endpoints used by the legacy
```
tgsoverly: Based on these descriptions, can we not use the legacy wrapper?  Seems like we could use the new API's directly

visivo/templates/samples/github-releases/data/releases.csv
```diff
@@ -0,0 +1,46 @@
...
+visivo,2024-05-13,v1.0.49,398,10
+visivo,2024-05-27,v1.0.50,425,8
+visivo,2024-06-10,v1.0.51,461,11
+visivo,2024-06-24,v1.0.52,492,9
+visivo,2024-07-08,v1.0.53,524,12
+visivo,2024-07-22,v1.0.54,561,10
+visivo,2024-08-05,v1.0.55,592,13
+visivo,2024-08-19,v1.0.56,628,11
+visivo,2024-09-02,v1.0.57,664,14
+duckdb,2024-01-15,v0.9.2,3120,42
```
tgsoverly, line 20: A way to update?

pyproject.toml
```diff
@@ -97,3 +97,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.black]
 line-length = 100
+# Pin to py312 so Black doesn't emit Python 3.13-only formatting that fails
+# the AST safety check on CI's Python 3.12 runtime. requires-python allows up
+# to 3.14 for downstream installers, but our CI lane is py312.
+target-version = ["py312"]
```
tgsoverly, line 103: We should get a ticket to upgrade to 3.14

tgsoverly (approved):

pyproject.toml
jpjsonic: I'll create a linear ticket.

visivo/templates/samples/github-releases/data/releases.csv
jpjsonic: Not currently- maybe we should use a different project that's more static. I didn't want to bring the edge case of a csv script model into the example projects because it does create a little friction with our existing designs.

viewer/src/components/explorerNew/SQLEditor.jsx
jpjsonic: No- I doubt it. I'll replace these with static imports.

viewer/src/components/onboarding/onboardingApi.js
jpjsonic: Yeah that's a good point

PR #416, opened 2026-05-12
Make visivo API consistent and work with changes in core

tgsoverly:
# Summary

  Extends the visivo viewer + Flask backend so the same per-resource API contract works against both
  visivo Studio (local Flask) and core (Django). Adds a pluggable auth scheme so a single viewer
  codebase can authenticate to either backend, unifies the deploy/job-response shapes between local and
   cloud, and fixes a handful of perf/correctness bugs surfaced along the way.

#  What changed

 ## API contract
  - Phase 0 audit + fixes: every per-resource detail GET now returns the canonical {id, name, status,
  child_item_names, config} envelope (locked with regression test + contract doc under
  docs/contracts/).
  - List GETs return {<resource_plural>: [...]}; run-status returns the flat {run_id, status, progress,
   ...} shape.
  - Project loader cleanup — fetchProject now means "fetch one project by id"; the legacy bulk-blob
  fetcher is renamed fetchProjectBlob for callers that still need it.
  - Dashboard child_item_names now recurses through anonymous containers (Dashboard.rows → Row.items →
  leaf chart/table/markdown/input), so the local Flask response matches what core returns.

  ## Pluggable auth for the shared viewer
  - viewer/src/api/utils.js exports setAuthHeaderProvider(fn) + apiFetch(url, init). Visivo Studio
  leaves the default no-op in place (still unauthenticated); core registers a JWT provider on boot.
  Same fetch sites work in both worlds, no fork.
  - All project-view api modules (dashboards, dashboard, charts, tables, markdowns, inputs, insights,
  defaults, project, models, insightJobs, inputJobs) and the parquet loader (duckdb/queries.js) now use
   apiFetch.

  ## Job-response unification
  - deploy_phase.py now posts each insight/input JSON content alongside the file refs, so core's
  /api/{insight,input}-jobs/ can serve the same fully-inlined record visivo Flask serves from disk — no
   client-side adapter needed.
  - Deploy POSTs model records at the renamed /api/model-jobs/ endpoint.

 ##  Viewer fixes
  - useModelsData builds /api/files/<name>/<run_id>/ URLs (no alpha_hash in URLs anywhere; the hash
  stays only as DuckDB's table identifier).
  - ProjectNew defensively optional-chains dashboard.config?.… so a single bad row can't take down the
  listing.
  - useInsightsData dropped the 'initial' sentinel that caused a duplicate insight-jobs fetch on
  metadata discovery — pure waste for input-less insights.

  Companion change in core

  A matching set of changes lives in the core/app repo (Django ModelJob rename, per-resource view
  ORM-backing, /api/dashboards/ requires project_id, FileByHash streams parquet, shape_job_response
  embeds direct storage URLs, etc.). Both PRs need to land together for a clean cutover.


visivo/commands/deploy_phase.py
```diff
@@ -273,7 +275,7 @@ async def create_model_records(batch, project_id, json_headers, host, progress):
     async with semaphore_3:
         try:
             async with httpx.AsyncClient(timeout=60) as client:
-                url = f"{host}/api/models/"
```
tgsoverly, line 276: Going to have to think through how to release this, causing deploy failure.

visivo/server/managers/object_manager.py
```diff
@@ -273,31 +274,51 @@ def _serialize_object(
...
+
+        def _add(name_):
+            if name_ and name_ not in seen:
+                seen.add(name_)
+                child_names.append(name_)
+
+        def _collect(child):
+            """Walk a child_items() entry, collecting leaf names.
+
+            Containers like ``Dashboard.rows`` / ``Row.items`` don't carry
```
tgsoverly, line 287: Not sure that will be true forever.  One of the reasons we added the name was for selectors.

docs/contracts/api-conventions.md
```diff
@@ -0,0 +1,241 @@
...
+  "name": "<resource_name>",
+  "status": "new|modified|published|deleted|null",
+  "child_item_names": ["dep_a", "dep_b"],
+  "config": { <full Pydantic dump, exclude_none=True, exclude={file_path, path}> }
+}
+```
+
+| Field | Notes |
+|---|---|
+| `id` | Object identifier. **Flask backends** (visivo CLI / Studio) emit `id == name` because there is no database — only one project is loaded at a time and names are unique within it. **Django backends** (core / cloud) emit the row's UUID primary key for `id` and keep `name` as the human-readable label. Viewer code must treat `id` as opaque and never assume it equals `name`. |
```
jpjsonic, line 26: This is really helpful. Are you thinking that we should incorporate this into our documentation or just leave here more as a developer note?

docs/contracts/api-conventions.md
```diff
@@ -0,0 +1,241 @@
...
+| `name` | The user-facing name of the object — what the user typed in YAML. |
+| `status` | One of `new`, `modified`, `published`, `deleted`, or `null`. Cloud should always emit `"published"` (no draft cache concept exists in deploys). |
```
jpjsonic, line 28: We might want a moved status as well for moving the definition of an object from one file to another.

docs/contracts/api-conventions.md
```diff
@@ -0,0 +1,241 @@
...
+
+## Error response shape
+
+Every non-2xx response uses:
+
+```json
+{ "error": "<human-readable message>" }
+```
+
+Always the key `error`, never `message` / `detail` / `reason`. Today some
```
jpjsonic, line 187: ❤️

viewer/src/api/project.js
```diff
@@ -1,54 +1,80 @@
 import { getUrl } from '../contexts/URLContext';
+import { apiFetch } from './utils';
 
-// ========== Old project endpoint (for backward compatibility) ==========
-export const fetchProject = async (projectId = null) => {
+// ============================================================
+// Legacy bulk-blob endpoint
```
jpjsonic, line 5: We'll delete this in clean up right?

viewer/src/api/utils.js
```diff
@@ -3,3 +3,62 @@ import { json } from 'react-router';
...
+export const setAuthHeaderProvider = provider => {
+  authHeaderProvider = typeof provider === 'function' ? provider : () => provider || {};
+};
+
+export const authHeaders = () => {
+  const result = authHeaderProvider();
+  return result && typeof result === 'object' ? result : {};
+};
+
+const looksLikeOurApi = url => {
```
jpjsonic, line 34: I wonder if there's a more deterministic way to do this.

jpjsonic (approved):
Looks good! Nothing blocking just left some questions and comments

docs/contracts/api-conventions.md
tgsoverly: Developer note and for the robots.

viewer/src/api/project.js
tgsoverly: Correct.

viewer/src/api/utils.js
tgsoverly: Will keep an eye on this.

PR #507, opened 2026-06-27
Local serve: run-on-save + run indicator; remove insight preview

tgsoverly:
## What & why

Brings the cloud **"run on save"** paradigm to local `visivo serve` and removes the parallel **insight/chart preview-run** path. Previews now just show the already-built `main` data; a save triggers a run that rebuilds it — and a toolbar **run indicator** shows the status.

The run-on-save backbone already existed (`run_phase` → `FilteredRunner` → `target/main/`) and the whole cloud viewer run-stack (`runStore`, `useRunPolling`, `runDataVersion` soft-refresh, Runs view) was already built — it was just **gated off locally**. This change mostly *enables + repoints* rather than building anew.

## Backend (`visivo/server`)
- **`RunManager`** — in-memory run registry in the cloud `Run` shape (one per process).
- **`request_run` + `save_run_executor`** — debounced; rebuilds the changed DAG slice into `target/main/` from the **cached-injected** project (shared `project_injection` helper), capturing build output as the run's logs. Coalesces rapid saves.
- **`after_request` hook** fires a run on any successful save/delete of a **data-producing** resource (sources/models/insights/inputs/dimensions/metrics/relations). Presentation resources (charts/tables/markdowns/dashboards) don't run — the viewer re-renders them from refreshed config + built data, mirroring the cloud.
- **`GET /api/projects/<id>/run/`** and **`GET /api/runs/<id>/logs/`** in the cloud `RunSerializer` shape; **`/api/project/` reports `status:"draft"`** — the one signal the viewer's run-poller gates on.
- **Removed** the insight preview executor + `POST`/`GET /api/insight-jobs/<run_id>` endpoints. **Kept** the built-data GET reader, `model-query-jobs` (SQL editor), and `source-schema-jobs` (they share `RunStatus`/`PreviewRunManager`).

## Frontend (`viewer`)
- **Mount `useRunPolling`** in the local shell (`Home`) and add a toolbar **`RunIndicator`** (running / failed) — the poller was never mounted in the viewer before, so locally `latestRun`/`runDataVersion` never updated. Now the full loop works: save → run → poll → soft-refresh (no full page reload).
- **Reimplement the preview hooks** to read built `main` data via `useInsightsData` (no preview run, no `__preview__` namespace); delete `usePreviewJob`.

## Tests
- Backend: new `tests/server/test_run_views.py` (`RunManager`, the run/logs endpoints, `status:"draft"`, debounced `request_run` rebuilding into `main`, coalescing). Full server suite green (501 passed).
- Frontend: rewritten preview-hook tests + run-indicator wiring. Affected viewer suites green (TopNav/Home, new-views/common, explorerNew, project/items rendering).

## Notes
- Everything is in the `visivo` repo (server + viewer); **no `core` changes**.
- The committed viewer bundles (`visivo/viewers/dist|local`) are **not** included — they're regenerated by the build step from the viewer source in this PR.
- Minor follow-up: `ExplorerChartPreview` still builds a `context_objects` payload the new hook ignores (harmless dead overhead).
- Manual smoke worth doing: `visivo serve` → edit an insight → run appears in the indicator, data refreshes on success, no preview fired, SQL/schema editors still work.


tgsoverly:
Screenshot 2026-07-06 at 2 36 48 PM

viewer/src/hooks/useRunPolling.js
```diff
@@ -30,24 +31,16 @@ export const useRunPolling = () => {
...
-      const active = ACTIVE_RUN_STATES.includes(latestRun?.state);
-      const withinWindow = Date.now() < (until || 0);
-      // Keep polling while a run is in flight (it sits queued through the cold
-      // start, then running) or we're still bridging to the first poll;
-      // otherwise stop until the next edit re-arms us.
-      if (active || withinWindow) {
-        timer = setTimeout(tick, active ? 2000 : 4000);
-      }
+      const active = ACTIVE_RUN_STATES.includes(useStore.getState().latestRun?.state);
+      timer = setTimeout(tick, active ? 2000 : 4000);
```
jpjsonic, line 35: Any downside to polling more frequently?

visivo/server/jobs/save_run_executor.py
```diff
@@ -0,0 +1,100 @@
...
+from visivo.jobs.filtered_runner import FilteredRunner
+from visivo.logger.logger import Logger
+from visivo.server.jobs.project_injection import inject_cached_objects
+from visivo.server.managers.run_manager import RunState
+
+# Coalesce rapid saves (e.g. one editor action touching several rows) into one run.
+_DEBOUNCE_SECONDS = 0.5
+
+_pending_names = set()
+_pending_lock = threading.Lock()
```
jpjsonic, line 24: I guess we could eventually do locks based on if there's any dag overlap. We could run dag slices that dont touch each other independently

viewer/src/stores/runStore.js
jpjsonic, line 30: What happens when the user makes an uncommitted change locally in code, and then runs visivo serve. I dont think we'd pick that up as an uncommited change right?

visivo/server/views/run_views.py
```diff
@@ -0,0 +1,59 @@
...
+    r"insights|inputs|dimensions|metrics|relations"
+    r")/([^/]+)/$"
+)
+
+
+def register_run_views(app, flask_app, output_dir):
+    @app.after_request
+    def run_on_save(response):
+        try:
+            if request.method in ("POST", "DELETE") and response.status_code < 400:
```
jpjsonic, line 112: How will this delete work? Do we have a flow yet for deleting objects and how it impacts downtream dependencies?

viewer/src/hooks/usePreviewData.js
```diff
@@ -1,297 +1,119 @@
-import { useState, useEffect, useMemo, useRef, useCallback } from 'react';
```
jpjsonic, line 1: Claude flagged this: 

## Explorer live chart preview no longer reflects unsaved / brand-new config

viewer/src/hooks/usePreviewData.js:72 (useChartPreviewJob) + explorer/ExplorerChartPreview.jsx

ExplorerChartPreview still builds a full previewRequest.context_objects payload (edited insight props, edited model SQL, input overrides) and passes it in, but the rewritten useChartPreviewJob uses only insight_names and reads insightJobs[name] from the built main run. context_objects is dropped, so:
- composing a new, never-saved insight → no main entry → the chart renders empty the whole time you're composing;
- editing an existing insight's query in the Explorer → shows stale pre-edit main data until a save→run completes (and per finding 1, the first save won't even refresh it).

The PR body calls this "harmless dead overhead," but it's a real interactivity downgrade in the Explorer right rail. This is worth an explicit product decision — it runs against the documented expectation that the Explorer right-panel preview reflects live insight/chart edits and interactions. At minimum, ExplorerChartPreview should stop constructing the ignored payload, and the empty-while-composing state should be handled intentionally (e.g. an "unsaved — save to preview" affordance).

jpjsonic (commented):
Added some comments & questions

viewer/src/hooks/useRunPolling.js
tgsoverly: Certainly not locally.  Cloud it could with lots of users.

visivo/server/jobs/save_run_executor.py
tgsoverly: That is a good idea.

viewer/src/stores/runStore.js
tgsoverly: I am not totally following.  Changes the files while serve is running or changes files and then runs `visivo serve`?

PR #422, opened 2026-05-21
[FEATURE] Visual Dashboard w/ Unified Editor & Lineage Experience 

jpjsonic:
# [FEATURE] Visual Dashboard Building — Unified Workspace, Canvas Editing & Lineage

This is the Dashboard Building v1 integration branch (Linear initiative `Dashboard Building v1`, Tracks A–O): a unified `/workspace` surface where dashboards are built visually on a canvas, every project object is editable in place, and the project graph is explorable — all backed by the draft-cache → **Commit** workflow.

### Workspace shell & navigation
- New `/workspace` routes with the shared TopNav; `/editor` and `/lineage` now redirect into the workspace (`?view=project`, `?view=lineage`).
- Left rail Library (all object types, context menus, inline create), tab strip with per-tab state (Track O: context menu + keyboard nav), right rail with universal edit routing (~500ms debounced auto-save).
- Commit cluster in the top bar: live pending-change count, "Saving…" activity, Commit / Discard, transient "Committed ✓", external-edit banner when a hot reload drops drafts (VIS-808).

### Canvas (Track D + follow-ons)
- Drag-and-drop dashboard building: cross-row moves, nested rows, empty-row targets, row templates, smart width, drop-on-slot fill.
- Affordance model: items grab via a selection border-ring frame, rows via a left gutter; row height snapping (S/M/L/XL + Shift for px) — `Row.height` accepts enum or pixel int (VIS-A1).
- View-mode flip cards (prefers-reduced-motion aware), ⋮ kebab for item actions, broken-ref placeholder card + reference picker (Track L).

### Per-object canvases (Canvas Object Surfaces, VIS-1001…1014)
- `objectCanvasRegistry` + `ObjectCanvasFrame`: N-way lens picker per object type with editable-lens contract.
- Source ERD (cached-schema Data tab, FK edges, right-click "create model from table"); semantic-layer ERD with relation nodes, cluster layout engine and drag/tidy; relations builder (drag column→column authors a Relation); markdown editor canvas; pivot playground Build lens with save-as-new; metric/dimension field lenses; insight/input/table previews.
- Model column schema as a run-phase job artifact (`/api/model-schema-jobs/…`), powering ERD columns + metric/dimension authoring.

### Explorer integration (Track J)
- Explorer overlay inside the workspace (round-trip with return chip), "Add to dashboard" slot picker, shared save path through the store (no parallel codepath).

### Backend (Flask, parity with core where applicable)
[...]

PR #514, opened 2026-07-07
Land the Right Rail Overhaul → main (integration)

jpjsonic:
## Land the Right Rail Overhaul → main (integration PR)

Merges the **`feature/right-rail-overhaul`** integration branch — the whole schema-driven-editor project — into `main`. This is the branch the phase PRs already landed on; it currently has **no** path to main, so this PR is that path.

### What it lands
- **Phase 0 + engine** — generic schema-form engine vs `$defs`, field widgets, unified-selection scaffolding, data-layer unification, tracePropCatalog (VIS-991/992/1018/1019).
- **Chart/insight props editor** — `TracePropsEditor` + `TypeSelector`, AJV validation, trace-group sidecars (VIS-1020).
- **Validation-as-save** — the `useRecordSave` gate, born-valid structure configs, run-failure loop-back, cloud read-only (VIS-993/1025).
- **The 5 remaining issues** (PR #513, once merged into this branch): VIS-994 selection routing, VIS-996 schema-driven leaf forms, VIS-1021 Field Finder, VIS-1000 brand cleanup, VIS-1026 metric preview.

> ⚠️ **Draft until [#513](../pull/513) merges into `feature/right-rail-overhaul`** and the two reconciliations below are handled. Mark ready once both are done.

### Reconciliation 1 — current main (#508)
main is 1 commit ahead of feature (#508 post-#422 follow-ups). The merge has **one** conflict: `viewer/src/components/views/workspace/RightRailEditPanel.jsx` — #508 removed the legacy `onItemWidthChange`/`onItemRefChange` item handlers; feature's VIS-993 work changed those same lines. **Resolution:** take #508's removal (drop `onItemWidthChange`/`onItemRefChange` + the now-unused `setItemWidth`/`applyLeafRef` imports), keep feature's `onItemChange` sole-channel. (This is the exact resolution already applied once on the working branch.)

### Reconciliation 2 — Tim's #507 (run-on-save / remove insight-preview)
**#507 targets `main` and is ready to merge.** It deletes the insight-preview run stack (`usePreviewJob`, `/api/insight-jobs/`, the `__preview__` namespace), rewrites `usePreviewData` to read main-run data (shape-compatible), and adds a `/runs` view + `RunIndicator`. The **feature branch still carries the old preview stack** (`usePreviewJob.js`, the pre-#507 `usePreviewData.js`), so whichever of {#507, this PR} lands second must reconcile:

- **Prefer #507's `usePreviewData.js`** (shape-compat) and **delete feature's `usePreviewJob.js`**; feature's surviving consumers (ExplorerChartPreview, common/workspace InsightPreview & ChartPreview, usePreviewInputDependencies) work unchanged against the new hook shapes.
- **VIS-1026 (in #513) already pre-migrated `MetricPlayground`** off the synthetic-insight path onto the surviving model-query-job + DuckDB aggregate, so it needs no further change here.
- **Update the `/api/insight-jobs/` e2e specs** (explorer-multi-insight-batched-preview, explorer-skip-rerun-on-static-only-changes, explorer-insight-remove-clears-inputs, canvas-metric-preview, canvas-field-lens) — they intercept the now-deleted endpoint. These were deliberately left in #513 because the endpoint still exists on this branch.

**Recommended order:** merge **#513 → feature** first, then merge **#507 → main**, then land this PR (resolving the #508 conflict + adopting #507's preview hooks + refreshing the e2e specs). Doing #507 first means the feature-side preview stack reconciles once, here, rather than twice.


viewer/src/components/styled/FormFooter.jsx
```diff
@@ -34,13 +34,18 @@ const FormFooter = ({
   saveLabel = 'Save',
   cancelLabel = 'Cancel',
   leftActions,
+  // VIS-993 auto-save mode: persistence is debounced through useRecordSave,
+  // so the footer renders NO Cancel/Save buttons — just Delete plus whatever
+  // the form passes as rightContent (typically a SaveStateIndicator).
+  autoSave = false,
```
tgsoverly, line 40: Always false?

viewer/src/components/views/common/buildTraceGroupSpec.js
```diff
@@ -0,0 +1,429 @@
+/**
+ * buildTraceGroupSpec.js — pure trace-prop grouping (VIS-1020 §3)
```
tgsoverly, line 2: Why "spec"? Makes me think of a test.

viewer/src/schemas/area.catalog.json
tgsoverly, line 1: How are these kept up to date?  Didn't we already have schemas in visivo?  Should we consolidate?

tgsoverly (approved):
The biggest question I have is around schemas, the structure seems straightforward.

viewer/src/components/styled/FormFooter.jsx
jpjsonic: Not always false — it's a defaulted prop. `SchemaLeafForm.jsx` passes `autoSave={isEditMode}` (true whenever you're editing a dimension/metric/relation in the right rail), and `RightRailEditPanel.jsx` passes it truthy for the Edit tab. When `autoSave` is true the footer renders **no** Cancel/Save buttons — persistence is debounced through `useRecordSave` — just the Delete button plus the `SaveStateIndicator`. So the branch is the live right-rail path, not dead.

viewer/src/components/views/common/buildTraceGroupSpec.js
jpjsonic: Fair point — "Spec" does collide with the test-file convention. It's a pure function that returns the ordered `groupSpec[]` the FieldGroup engine renders — "spec" as in the group *specification* (like a layout spec), not a test spec. Leaving the name as-is for now to avoid churning every import site, but noting it as a rename candidate (`buildTraceGroups`).

viewer/src/schemas/area.catalog.json
jpjsonic: Good question — traced it end to end:

**Two different sources, so there's nothing to consolidate into visivo:**
- The `.schema.json` sidecars (the schemas we already had in visivo) are auto-synced — `copy-schemas` runs `cp ../visivo/schema/*.schema.json src/schemas/` before every `start`/`test`/`build`.
- The **new** `.catalog.json` / `.groups.json` are *generated* from `tracePropCatalog.js` (curated tiers/labels/keywords) + plotly's `plot-schema.json` by `scripts/build-trace-{catalogs,groups}.js`. They carry viewer-only *presentation* metadata (tier + group bucket) that doesn't belong in the pydantic schema — hence separate from the `.schema.json` files rather than folded into visivo.

**On "kept up to date":** I regenerated both → 0 diff (in sync today), but you're right that nothing guarded against drift — the generators weren't wired into `copy-schemas` / package.json / CI. Fixed in this push: added a `traceGroups` drift-guard test that re-derives every `.groups.json` from the live `plot-schema.json` using the build script's own classifier and diffs, so a plotly.js bump that adds/removes/renames a leaf now fails CI. (`tracePropCatalog.coverage.test.js` already guarded the catalog side against the schema.)

jpjsonic: We'll have to update this as plotly makes new releases. I'm kinda just betting that there's not going to massive changes to trace configs that make our curated selections out of date

PR #515, opened 2026-07-10
Workspace UX polish: URL-addressable tabs, persistence, surface reopen + fixes

tgsoverly:
A batch of Workspace/canvas tweaks from a design-review thread — sharpening the new build-mode UX. Nine self-contained changes across the Library, the tab model, and a couple of bugs.

## Library & "+ New" menu
- **Removed the redundant inline "+ New X" buttons** in each library subsection — creation now lives solely in the header "+ New" menu.
- **Reworked the "+ New" menu**: grouped like the sidebar (Layout Items · Data Layer), dropped the redundant "New " prefix on every item, and **added "Relation"** — which opens the Semantic Layer (a relation can't be templated; its condition needs two real models).
- **Added a Project · Explorer · Semantic Layer row** at the top of the Library so those surfaces are reopenable after their tabs are closed ("I closed the project tab and it isn't obvious how to get it back"). Project reliably **resurrects** its closed tab.
- **Minimized-nav selection**: expanding the left rail now reveals the active object — it expands that object's section + subsection and scrolls the row into view, instead of leaving the selection hidden in a collapsed group.

## Tabs & navigation
- **The active tab is now in the URL** → the browser Back/Forward buttons walk tab history, and deep links / refresh restore the active tab.
- **Document title** reflects the active tab (`<tab> · <project>`), so the browser tab is informative.
- **Open tabs persist across refresh** (localStorage, keyed per project). The open *set* lives in the store; the *active* tab comes from the URL.

### Architecture: one clean URL↔store loop
The URL and the store aren't kept in sync with a two-way mirror. Instead:

> **set → the store action writes the URL → `useWorkspaceUrlSync` reads the URL into the store → components read the store only.**

You still call `openWorkspaceTab` / `switchWorkspaceTab` as before; they route the *active selection* through the URL (via a `navigate` the Workspace registers on mount). `activateWorkspaceTab` is the single URL→store writer. A **no-navigator fallback** activates in-store synchronously when no router is mounted, so all ~60 existing call sites keep working unchanged. New URL-addressable values are a small addition to `workspaceUrl.js` + the sync hook.

One deliberate behavior: `/workspace` == the project tab, so the project ("home") tab is always available there.

## Bugs
- **Continuous `/api/inputs/` polling**: on an input-less project `fetchInputs` writes a fresh `[]` each call, changing the `inputs` reference; the effect keyed on it re-fetched forever. Now lazy-loads once (ref-guarded).
- **Model editor "black panes" on rail resize**: the SQL editor is Monaco, whose `automaticLayout` only *polls* — under a flex-driven rail drag its `vs-dark` canvas is left unpainted (the black panes). Now forces `editor.layout()` whenever a rail width changes.

## Test plan
- New/updated tests for every item: the input-poll loop guard, library menu grouping + Relation, the Back-button round-trip (`openWorkspaceTab` → URL → store → Back → store), document title, persist/restore across a remount, the surfaces-reopen row, reveal-on-expand, and the Monaco relayout-on-resize.
- **Full viewer suite green: 315 suites / 5211 tests, 0 failures.** `yarn lint` clean.


tgsoverly:
Screenshot 2026-07-11 at 7 09 57 AM

There is a good place to hook in the explorer

PR #517, opened 2026-07-17
Explore 2.0 Phase 1: Exploration entity (Pydantic model, Flask CRUD, store slice)

jpjsonic:
## Summary

Phase 1 of the Explore 2.0 plan — the Exploration entity, backend + store slice only (no visible UI; that's Phase 2). Implements the frozen S3 wire contract exactly: [`07-exploration-api-contract.md`](../blob/feature/explorer-workspace/specs/plan/explorer-workspace-unification/07-exploration-api-contract.md).

## Contract conformance checklist

- [x] `visivo/models/exploration.py`: `Exploration`/`ExplorationDraft`/`ExplorationQueryDraft`/`PromotionRecord`/`ReturnToRef`/`SeedRef`, field-for-field per contract §Entity. Timestamps serialize with a `Z` suffix (not `+00:00`) to match the sample.
- [x] **Not part of the DAG/schema**: `visivo/models/exploration.py` is never imported by `visivo.models.project`; `python -m visivo.generate_project_schema_json` + `git diff --exit-code visivo/src/visivo_project_schema.json` confirmed a true no-op (verified locally; no schema diff to commit).
- [x] Repository: JSON documents in `<project root>/.visivo/explorations/<id>.json`; mkdir-on-first-use (construction never touches disk); atomic writes (tmp file + `os.replace`); server-minted url-safe id (`exp_<8 hex>`); list ordered by `updated_at` desc.
- [x] Endpoints, verbatim: `GET/POST /api/explorations/`, `GET/POST/DELETE /api/explorations/<id>/` (**POST, not PUT** for update), `POST /api/explorations/<id>/consume-return-to/`.
- [x] Name defaulting: "Scratch" for the first exploration, "Exploration N" after (resolved question 3).
- [x] Update is a full-document replace of the mutable fields (`name`/`draft`/`return_to`) *present in the patch*; `id`/`created_at`/`seeded_from`/`promoted` are immutable via this route regardless of what the patch contains.
- [x] 400 reserved for a malformed **top-level** body shape; draft *content* (`insights`/`chart`/`computed_columns`) is never validated at rest — only the wire-shaped `queries` sub-list is structurally typed.
- [x] `consume-return-to` atomically (and idempotently) nulls `return_to`.
- [x] **Run isolation**: exploration routes are absent from `run_views.RESOURCE_META`/`_RESOURCE_ROUTE_RE`; regression-tested that POST-update + DELETE schedule no run via a real integration client (not a vacuous PUT-based test).
- [x] Client: `viewer/src/api/explorations.js` (thin wire client) + `stores/workspaceExplorationsStore.js` (byId/order slice, debounced optimistic sync, flush-on-unmount primitive).

## S4 git-history recon (reused vs. diverged)

Recovered the removed 2.0-era exploration files (commits `00a83329`/`e172d91b`):
- **Reused structurally**: the repository class shape (constructed with a storage path; `create/list/get/update/delete` methods returning plain dicts/records), the `register_x_views(app, flask_app, output_dir)` view-registration convention, and the view test's "mount blueprint on a bare Flask app" pattern.
- **Diverged deliberately** (per `02-architecture.md` §2): the old repository was SQLite-backed at `target/explorations.db` — rejected here because `target/` is a rebuildable output dir and explorations are user workbench data that must survive `rm -rf target`. This repository instead writes one JSON document per exploration under a new project-root `.visivo/explorations/` directory (already covered by the generated `.gitignore` template; watcher-safe since `hot_reload_server.py` only reacts to `.yml`/`.yaml`).
- The old model was a SQLAlchemy `ExplorationModel` with a flat single-query-per-tab shape (`sql`, `source_name`, `left_panel_tab`, …) — replaced entirely by the Pydantic S3 contract shape (`draft.queries[]`, `seeded_from`, `return_to`, `promoted[]`), since the wire shape changed fundamentally between the old pre-2.0 Explorer and Explore 2.0.

## Run-isolation evidence

`tests/server/test_run_views.py::TestExplorationRunIsolation`:
- `"explorations" not in RESOURCE_META`
- `_RESOURCE_ROUTE_RE.match(...)` is `None` for both `/api/explorations/<id>/` and the `consume-return-to` sub-route
- Real `integration_client` POST-update, DELETE, and consume-return-to calls all assert `visivo.server.views.run_views.request_run` is never called

## Test counts

- Backend (Python): full suite `2082 passed, 2 skipped, 5 xfailed, 1 xpassed` (no regressions). New: 14 model tests, 42 repository tests, 26 view tests, 5 run-isolation regression tests. `black --check` clean.
- Frontend (JS): full viewer suite `317 suites / 5257 passed` (no regressions), `eslint` clean. New: 16 API client tests, 20 store slice tests.

## Deviations / judgment calls (flagged for the orchestrator)

1. **`promoted` mutability ambiguity in the frozen contract itself**: the endpoint table says `promoted` is immutable via the POST-update route, but the informative note directly below it says "the client then appends `promoted[]` records via the update route" — those two statements conflict. Phase 1 conservatively implements the literal endpoint-table rule (`promoted` fully immutable via `POST /api/explorations/<id>/`) since promotion itself is out of scope until Phase 4. Phase 4 will need to either add an explicit allowance/sub-route for appending `PromotionRecord`s, or this doc needs a follow-up PR to resolve the contradiction before that phase starts.
2. **Update-with-no-body is 400** (not treated as a no-op): an absent JSON body on `POST /api/explorations/<id>/` returns 400 ("nothing to update"), while an empty object `{}` is accepted as a valid (if pointless) no-op update. This wasn't explicitly specified in the contract; it mirrors the removed 2.0-era view's behavior and is a defensive guard against a client bug that forgets to send a body.
3. **Default-name counting**: "Scratch" → "Exploration 2" → "Exploration 3" counts *all* existing explorations (including custom-named ones), not just auto-named ones — e.g. creating "Churn dig" then a nameless one yields "Exploration 2", not "Scratch". This reads naturally from the resolved-question wording but wasn't spelled out; flagging in case product intent differs.
4. **`integration_app` test fixture** (`tests/server/conftest.py`) now passes an explicit `working_dir` (the same isolated tmp dir as `output_dir`) instead of implicit `None`. This was necessary so exploration writes in tests stay inside the temp folder rather than leaking `.visivo/explorations/*.json` into the real working tree when `working_dir` defaulted to `os.getcwd()`. Verified no other test relies on the previous `None` default.

## Nothing left blocking Phase 2

Phase 2 (Explorer joins the shell) can proceed against this backend + slice as-is — `objectTypeConfigs`, `MiddlePane` routing, and the Home gallery are all additive from here.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01YQYjb2XjBnqBdotFKCp9tX

### DIFF
--- a/tests/factories/model_factories.py
+++ b/tests/factories/model_factories.py
@@ -1,4 +1,5 @@
 import re
+from datetime import datetime, timezone
 
 import factory
 from visivo.models.alert import Alert
@@ -30,6 +31,14 @@ from visivo.models.inputs.types.single_select import SingleSelectInput
 from visivo.models.inputs.types.multi_select import MultiSelectInput
 from visivo.models.inputs.types.display import RangeConfig
 from visivo.models.base.query_string import QueryString
+from visivo.models.exploration import (
+    Exploration,
+    ExplorationDraft,
+    ExplorationQueryDraft,
+    PromotionRecord,
+    ReturnToRef,
+    SeedRef,
+)
 
 
 class SingleSelectInputFactory(factory.Factory):
@@ -393,3 +402,61 @@ class ProjectFactory(factory.Factory):
                 [factory.SubFactory(DashboardFactory, table_item=True) for _ in range(1)]
             ),
         )
+
+
+class ExplorationQueryDraftFactory(factory.Factory):
+    class Meta:
+        model = ExplorationQueryDraft
+
+    name = "orders_q"
+    sql = "SELECT * FROM orders"
+    source = "warehouse"
+
+
+class ExplorationDraftFactory(factory.Factory):
+    class Meta:
+        model = ExplorationDraft
+
+    queries = factory.List([factory.SubFactory(ExplorationQueryDraftFactory)])
+    insights = factory.List([])
+    chart = None
+    computed_columns = factory.List([])
+
+
+class SeedRefFactory(factory.Factory):
+    class Meta:
+        model = SeedRef
+
+    type = "model"
+    name = "orders"
+
+
+class ReturnToRefFactory(factory.Factory):
+    class Meta:
+        model = ReturnToRef
+
+    dashboard = "kpis"
+    slot = "r2-i1"
+
+
+class PromotionRecordFactory(factory.Factory):
+    class Meta:
+        model = PromotionRecord
+
+    type = "metric"
+    name = "churn_rate"
+    promoted_at = factory.LazyFunction(lambda: datetime.now(timezone.utc))
+
+
+class ExplorationFactory(factory.Factory):
+    class Meta:
+        model = Exploration
+
+    id = factory.Sequence(lambda n: f"exp_test{n:04x}")
+    name = "Scratch"
+    created_at = factory.LazyFunction(lambda: datetime.now(timezone.utc))
+    updated_at = factory.LazyFunction(lambda: datetime.now(timezone.utc))
+    seeded_from = None
+    return_to = None
+    draft = factory.SubFactory(ExplorationDraftFactory)
+    promoted = factory.List([])
--- a/tests/models/test_exploration.py
+++ b/tests/models/test_exploration.py
@@ -1,0 +1,128 @@
+"""Tests for the Exploration Pydantic model (Explore 2.0 Phase 1, S3 contract).
+
+Also asserts the model is NOT reachable from ``visivo.models.project`` — the
+guarantee that ``generate_project_schema_json`` is a no-op for this file (the
+actual regen + git-diff assertion is a manual protocol step; see the PR body).
+"""
+
+import json
+
+import pytest
+from pydantic import ValidationError
+
+from visivo.models.exploration import (
+    Exploration,
+    ExplorationDraft,
+    ExplorationQueryDraft,
+    PromotionRecord,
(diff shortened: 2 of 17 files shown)


jpjsonic:
Merged into `feature/explorer-workspace` (rebase, per-concern commits preserved). Orchestrator notes on the flagged deviations: (1) the `promoted` mutability contradiction was REAL — resolved in the contract (specs `496e233`→`record-promotion` commit): `promoted[]` stays immutable via the generic update route and Phase 4 adds an append-only `POST /api/explorations/<id>/record-promotion/` sub-action with server-stamped `promoted_at`. Your conservative immutable reading was correct. (2) 400-on-empty-update-body, (3) default-name counting across all explorations, and (4) the `integration_app` fixture `working_dir` isolation are all accepted as-is.

jpjsonic:
Linear backfill: this PR shipped **Phase 1 — Exploration Entity** = VIS-1041, VIS-1042, VIS-1043, VIS-1044, VIS-1045 (all Done) under the [Explore 2.0 initiative](https://linear.app/visivo/initiative/explore-20-explorer-workspace-unification-19033169abcd).

PR #518, opened 2026-07-17
Explore 2.0 Phase 0: destinations view switcher + structural fixes

jpjsonic:
## Summary

Implements Phase 0 of the Explore 2.0 plan (`specs/plan/explorer-workspace-unification/`): the three-destination view switcher (Project · Semantic Layer · Explorer, D1) pinned atop the LeftRail, replacing the Library's old per-surface buttons. Destinations leave the workspace tab model entirely — they are never `workspaceTabs` records, never dirty, never closable — with a one-time localStorage scrub for any pre-Phase-0 persisted tab records. Opening any document tab (deep link, Library, context menu) sets the active destination to that document's owning view via a single write path (`activateWorkspaceTab`), so closing the last open tab returns to *that* destination's Home rather than always Project.

No Explorer functionality changes — `explorer` registers a clean placeholder Home pane so the 3-row switcher is fully real and testable now; Phase 2 swaps in the real Explorer Home gallery behind the same registry entry.

## Bugs closed

- **B1** — Semantic Layer's TabStrip fallback icon vs. its pane's borrowed `relation` icon: both now resolve the same new `objectTypeConfigs.js` entries.
- **B2** — RightRail "coming soon" dead-end for a destination: `getTabs(null)` now returns no tabs at all (destinations have no Edit surface) instead of falling through to a dead Edit tab.
- **B3** — `useWorkspaceScope`'s semantic-layer fall-through to a nonsensical `item` scope with a `+semantic-layer` selector: destinations now resolve through the view registry's own `scope` field.
- **B7** — `api/explorer.js`'s 6 dead exports (grep-verified, no callers besides their own tests) deleted; `fetchDiff` kept and given a focused test.
- **B8** — `previewRegistry.js` (a one-release shim with no remaining production consumer) deleted.
- **B11** — dimension/metric canvas descriptors move off the dead `sourcesMetadata` live-introspect feed (returns zero databases for DuckDB) onto `sourceSchemaJobsList`, mirroring the source ERD's existing fix.
- **B15** (partially, ahead of its filed Phase 2) — the Library's Explorer button no longer bare-`navigate()`s out of the shell; it activates the same real workspace-view state as its siblings.

## Structural changes

- `higherLevelViews.js` — the view registry (key, HomePane, urlPath, scope), consumed by MiddlePane, `useWorkspaceScope`, `workspaceUrl.js`, and the keyboard shortcuts.
- `workspaceStore.js` — `workspaceActiveView` field + `activateWorkspaceView`/`openWorkspaceView` actions; `activateWorkspaceTab`/`openWorkspaceTab` transparently delegate view-typed payloads so existing call sites (ProjectEditor's "Semantic Layer" button, TabStrip's `+`, Cmd+T) keep working unmodified; deep-link owning-destination rule; tab-set scrub migration.
- `workspaceUrl.js` — simplified to a `{kind: 'view'|'tab'}` resolver; `/workspace`, `/workspace/semantic-layer`, `/workspace/exploration` (reserved for Phase 2) are the three view paths.
- `Workspace.jsx` — mount/URL-sync effect rewritten off the project-tab-hydration model; persists `workspaceActiveView` to localStorage alongside the tab set, with a first-sync guard so a reload at the ambiguous bare `/workspace` root prefers a restored non-Project view.
- `MiddlePane.jsx` — dispatch is now: no active document tab → the active destination's HomePane; dashboard → DashboardPane; else → PerObjectPane.
- `TabStrip.jsx` — drops the dead `project` icon special-case; no longer returns `null` on an empty tab list (that's the *default* state now, not a transient impossibility — the `[+]` affordance must stay visible).
- Keyboard: Cmd/Ctrl+1/2/3 → the three views (not `Cmd+Shift+3/4/5`, which are OS-reserved macOS screenshot chords); tab-position switching shifts down to Cmd/Ctrl+4-9 — documented as a deliberate reconciliation.

## Test evidence

- `bash scripts/viewer-check.sh` (full suite + lint): **317 suites / 5246 tests passing** (2 pre-existing skips), lint clean.
- `npx eslint <every new/changed src file> --max-warnings=0`: clean.
- New unit coverage: `higherLevelViews.test.js`, `ViewSwitcher.test.jsx`, plus new/updated cases in `workspaceStore.test.js` (view actions, deep-link rule, scrub migration, close-hands-off), `workspaceUrl.test.js` (view/tab URL resolution), `useWorkspaceScope.test.jsx` (B3), `MiddlePane.test.jsx` (destination Home dispatch), `RightRail.test.jsx` (B2), `LeftRail.test.jsx`, `Workspace.test.jsx`, `TabStrip.test.jsx`, `useWorkspaceTabShortcuts.test.jsx` (keyboard reconciliation), `objectCanvasRegistry.test.js` (B11), `api/explorer.test.js` (B7), `ObjectTypeFilter.test.jsx`.
- New e2e story `view-switcher.spec.mjs` (switch views + URL nav, tab park/restore, reload-restores-view, B1 icon parity, deep-link owning-destination) — **written but not run** (shared sandbox; per instructions the orchestrator runs e2e at merge).
- Updated 3 existing e2e stories whose assumptions the destination model legitimately breaks: `tabbed-navigation.spec.mjs`, `workspace-tabs-context-menu.spec.mjs`, `workspace-tabs-shortcuts.spec.mjs` (project-tab-locator assumptions, Cmd+1/2/3 reassignment, tab counts).

## Deviations / judgment calls

- **Explorer's Home is a clean placeholder** ("Explorer arrives with explorations"), not the real gallery — matches the task's explicit choice (render the row now, swap the pane in Phase 2) over hiding the row.
- **HomePanes are NOT `React.lazy`-wrapped at the registry level** — only their genuinely heavy inner content is (e.g. `SemanticLayerCanvas`), matching the pre-existing `ProjectPane`/`SemanticLayerPane` precedent. Wrapping the whole pane in `React.lazy` would force every consumer through an extra Suspense tick for no real code-splitting benefit and broke a pile of synchronous test assertions for no gain.
- **Keyboard reconciliation**: Cmd+1/2/3 claimed by views forced tab-position switching down to Cmd+4-9 — a real, unavoidable conflict on a 1-9 keyspace; documented inline in `useWorkspaceTabShortcuts.js` rather than silently changing behavior.
- **`useWorkspaceScope`'s VIS-835-style "project tab wins over a stale dashboard URL" test was retired**, not ported 1:1 — project can no longer be a tab, and there's no clean analogous race to assert without contradicting "URL is canonical when nothing more specific overrides it." Replaced with equivalent-intent coverage (view scope when no tab; document tab always wins over a lingering view).
- **`RightRail` and its rail body render literally zero tabs** (not even a disabled "Edit") while a destination owns the center, per B2's letter — the panel still shows a real empty state ("Select an object…"), never a dead end.

Left for a later phase: the real Explorer Home gallery (start-from-source tiles, recent/parked exploration cards) and the exploration entity itself (Phase 1-2 of the plan).


jpjsonic:
Linear backfill: this PR is **Phase 0 — Destinations & View Switcher** = VIS-1035–VIS-1040 (In Review) under the [Explore 2.0 initiative](https://linear.app/visivo/initiative/explore-20-explorer-workspace-unification-19033169abcd). Integration-gate e2e fixes are being applied before merge.

PR #527, opened 2026-07-19
Explore 2.0: Explorer → Workspace unification (destinations, first-class explorations, pill grammar, promote flywheel)

jpjsonic:
# Explore 2.0 — Explorer → Workspace Unification

The Explorer is no longer a separate page: it is now the third first-class **destination** of the Workspace (Project · Semantic Layer · Explorer), and **explorations** are first-class, backend-persisted objects with a full lifecycle — create, autosave drafts, park/resume as tabs, promote into the project, and pull project objects back in ("Explore this") to close the flywheel: query sources → build insights/charts → promote → create metrics/dimensions → explore again.

**71 commits · 307 files · +34,355 / −14,920 (net −8k in the legacy Explorer surface).** `main` is a clean ancestor — no conflicts.

Full plan, architecture, UX spec, API contract, and design walkthrough: **visivo-io/specs#6** (`plan/explorer-workspace-unification/`).

## What landed (per-phase PRs, each merged through the same gate)

| PR | Phase | Highlights |
|----|-------|-----------|
| #517 | 1 — Exploration entity | Pydantic `Exploration` model, file-backed repository, Flask CRUD (POST-update wire contract, `consume-return-to`, append-only `record-promotion`), `workspaceExplorations` store slice with debounced sync |
| #518 | 0 — Destinations | `higherLevelViews` registry, ViewSwitcher pinned atop LeftRail, project/semantic-layer leave the tab model, URL sync simplification, Ctrl/Cmd+1/2/3 |
| #519 | 2 — Explorer joins the shell | `exploration` document type, ExplorerHome gallery, `/workspace/exploration[/:id]` routes, legacy bridge |
| #520 | 3a — Library expansion & DnD merge | Library source→schema→table→column drill-down with drag-out (D9), query chips on the SQL editor header, unified `WorkspaceDndContext`, shared ref-walk + inline-rename utils |
| #521 | 3b — Build rail, pill grammar & cutover | Insight+Chart CRUD rebuilt in the RightRail exploration branch, **D10 Tableau-informed pill aggregation grammar** (typed pills — raw `?{}`/`${}` never rendered (D8); type-restricted presets incl. dialect-gated MEDIAN mirrored in `sqlglot_utils.py`), `/explorer` cutover redirect, 8 legacy components deleted |
| #522 | 4 — Promote & preview | Gated promote checklist modal (update-by-name semantics), draft insight preview via `/api/insight-compile-draft/` + DuckDB overlay, true discard snapshot/revert |
| #523 | Hardening | Per-id write queue, backend name-collision lock, deleted-remotely banner, StrictMode-safe mounts, 6 race e2e stories |
| #524 | 5 — Flywheel & polish | "Explore this" from Library/context menus/ERD, dashboard `+ New Chart` → exploration → **Place in dashboard** round-trip, staleness banners, flywheel telemetry (`exploration_created/opened/promoted`, `time_to_first_chart` — taxonomy doc updated) |
| #525 | 6 — Cleanup & acceptance | Deletions ledger executed, 44-spec legacy e2e ledger fully reconciled, onboarding pass, 11 armor stories (cross-tab, back/forward, phantom tabs), VIS-726/662 reverified |
| #526 | 6b — Delta closure | **Promoted-lane freshness rebuild** (durable promote-time signature per exploration+insight, covering referenced-model state — fixes silent stale-chart-after-promote incl. the remount hole), onboarding/modal guards, e2e honesty fixes, two-stage publish-spec serialization |

## How it was verified

- **Per-phase gate** (all 7 build phases + hardening + 6b): viewer units + lint, backend pytest, black, then the full exploration e2e suite on a live sandbox with **3 consecutive fully-green runs required**. The gate caught real integration bugs at every single phase.
- **Recurring adversarial review** after each merge (multi-lens finders + per-finding adversarial verification): 60+ confirmed findings across the initiative, every one fixed at source or explicitly dispositioned in `research/e2e-gap-review.md` (specs PR). The final pass on Phase 6 itself found the promoted-lane remount hole (HIGH) — closed in #526.
- **Final acceptance evidence** (tip `d9373f6a`): e2e stage A **112/112 ×3 consecutive** + stage B (publish project) green ×3; viewer **5,690** unit tests passed, lint clean; backend **2,134** pytest passed; black clean. Full report: `research/acceptance-report.md` in the specs PR.
- Playwright suite grew by ~40 exploration/workspace stories, incl. regression armor for every HIGH finding (e.g. VIS-1091/1092 cross-exploration promoted-lane isolation, deterministic dirty-close-dialog races, backend-verified-after-reload asserts).

## Known caveats / follow-ups (all tracked)

- **VIS-1096** (LOW, deferred to Phase 7): concurrent first-visit seed race on an empty project can mint two seed explorations — cosmetic; proper fix is a server-side idempotency key on the create contract.
- **VIS-1080**: pre-existing resolver global-name-first bug (`${ref(model).field}` resolves by global field name, ignoring the stated model) — repro-verified, pill grammar deliberately replicates it for consistency; awaiting triage.
- **Dashboard-building test topology**: ~46 `parallel`-project specs (canvas/project-editor/etc.) require that initiative's per-spec dedicated sandboxes and fail environmentally in the standard :8001/:3001 topology on *every* branch incl. `main` — not regressions. Their `state-mutating` project's `dependencies: ['parallel']` edge means those 3 specs never run in this topology either (same flaw we fixed for `workspace-publish` in #526); flagged to that track.
- **`www/` marketing claims**: audited report-only in #525 — no copy changes required (Explorer flows aren't named on the site).
- **Phase 7 — cloud twin** (parallel track, not in this PR): Django Exploration twin per the frozen S3 contract (`07-exploration-api-contract.md`), VIS-1032/1077–1079.

## Review guide

Start with the specs PR walkthrough (`design/README.md` screenshots), then: `visivo/models/exploration.py` + `server/views/exploration_views.py` (backend), `viewer/src/stores/workspaceExplorationsStore.js` (lifecycle), `viewer/src/components/views/workspace/ExplorationPane.jsx` + `ExplorationBuildRail` (surface), `viewer/src/utils/insightFreshnessSignature.js` + `ExplorerChartPreview.jsx` (preview lanes), `viewer/src/utils/pillGrammar.js` + `PillMenu.jsx` (D10). Run it: `bash scripts/sandbox.sh start` → http://localhost:3001/workspace/exploration.


jpjsonic:
**Hold for UX overhaul before merge.** First hands-on review by Jared surfaced serious composed-experience problems the behavioral gates never measured (dual right rails, raw catalog error surfaced pre-first-run, confusing banner copy, duplicated type selectors, editor layout dead space). A full adversarial UX audit (real-browser driving with per-step screenshots across five flow families on isolated sandboxes) is running now; findings + an overhaul plan will follow as an additional phase on this branch. Functionality/test evidence in the PR body stands — the experience layer does not yet.

jpjsonic:
## Status update — Phase 6c Wave 1 is merged into this branch

This PR remains **on hold for your review** (decision D13). Posting what has landed since it was opened, because the branch has moved substantially.

### Wave 1 merged (#531 → `a2ecf4cd`)

All three UX-overhaul tracks answering your "pretty unusable" feedback, plus two e2e harness bugs that had to be fixed before any of it could be trusted, plus a coverage push. 43 commits, 115 files.

**Your five screenshot defects, re-measured on the merged tip via a walkthrough driven as a user:**

| you reported | now |
|---|---|
| two right rails | **one** — asserted from the DOM, 360px, Build content |
| `Catalog Error: Table with name molcjj…` before running anything | gone; guided *"Nothing to preview yet"* state |
| `References input "model" — not yet promoted` | gone |
| two stacked Scatter/Line dropdowns | one |
| Save-to-Project vs promote confusion | single `Save to Project` CTA |

Also: no `0 of 1366 properties`, wells offer `Type @ to insert a reference` instead of a number spinner that mangled refs, the SQL editor no longer eats ~440px of dark void, and at 1100px the library auto-collapses to an icon strip with results legible rather than collapsing to a sliver. Zero console errors.

**Still visibly outstanding, and correctly scoped to Wave 2:** the scratch query chip renders as `model` and its insight as `insight` — the generic names that pollute a project on save (T3's naming step, D11) — and the left-rail footer still shows dashboard-canvas help text on the Explorer surface.

### Two harness bugs worth knowing about

Both were silently corrupting results rather than failing: 34 specs pinned the backend address to port 8001 regardless of which sandbox they were driving, so parallel agents were deleting each other's records and asserting against foreign projects; and 15 more drove the shared frontend via a base that ignored `PLAYWRIGHT_BASE_URL`. `scripts/check-e2e-sandbox-isolation.sh` now enforces both classes.

### Coverage

Wave 1 measured 92.6% statements / 81.4% branches, with `InsightBuildSection` at **56% branch**. Raising it found a real bug no audit persona and no e2e spec caught: **the first click on a collapsed-by-default Build-rail group header was a silent no-op** (VIS-1105) — a dead click on a section header in the panel this phase exists to fix. It also found four tests that could not fail, each confirmed by deleting the guard from the source and watching the test stay green.

### Gate

Stage A 113/113 on three consecutive runs; stage B 1 passed + 9 env-skips each; viewer 340 suites / 6,123 passing; lint clean.

Wave 2 (T3 vocabulary/D11, T5 lifecycle & entry points) is in progress on its own branches and will come through as separate PRs into this branch. **`main` remains untouched — this PR is yours to merge.**

viewer/src/api/insightCompileDraft.js
tgsoverly, line 1: This should have "draft" in the name to match execute

viewer/src/api/insightExecuteDraft.js
```diff
@@ -0,0 +1,66 @@
...
+ *   models: Array<{name:string, name_hash:string}>,
+ * }>} — `rows` ARE the final chart rows; bind them through `props_mapping`.
+ * @throws {Error} with `.errorType`:
+ *   - `'requires_client_lane'` (409): the insight is dynamic (references an
+ *     input), so the server can't bake it — the caller falls back to the client
+ *     DuckDB sample lane.
+ *   - `'model_not_run'` (422): a referenced scratch model has no schema.
+ *   - `'unavailable'`: draft preview isn't available in this deployment mode.
+ */
+export const executeDraftInsight = async ({
```
tgsoverly, line 32: Those are were the explorer gets its data for previewing?  So an exploration creates those jobs?  Should they be called draftExploration, and executeExploration?  Because they contain more than an insight.

viewer/src/api/explorations.js
```diff
@@ -0,0 +1,128 @@
...
+    method: 'DELETE',
+  });
+  if (response.status === 204) {
+    return true;
+  }
+  throw new Error(await parseErrorOr(response, 'Failed to delete exploration'));
+};
+
+/**
+ * Atomically null `return_to` — the placement intent has been consumed.
```
tgsoverly, line 99: I will need to look into these, how do they relate to core.

viewer/src/api/explorations.js
```diff
@@ -0,0 +1,128 @@
...
+  throw new Error(await parseErrorOr(response, 'Failed to consume return_to'));
+};
+
+/**
+ * Append-only promotion record (Explore 2.0 Phase 4,
+ * 07-exploration-api-contract.md's record-promotion sub-action). Server
+ * stamps `promoted_at`. `type`/`name` are the promoted object's kind
+ * ('model'|'metric'|'dimension'|'insight'|'chart') and name.
+ */
+export const recordPromotion = async (id, type, name) => {
```
tgsoverly, line 118: Does this save the items to the project?

tgsoverly (approved):

viewer/src/api/explorations.js
jpjsonic: Yeah were going to need to do some integration work with core once we ship this.

viewer/src/api/explorations.js
jpjsonic: No it just tracks that the changes were promoted into the cache from the explorer the cache store still handles writing back to the files

viewer/src/api/insightCompileDraft.js
jpjsonic: Yeah that makes sense to me.

viewer/src/api/insightExecuteDraft.js
jpjsonic: Well they require context of other objects, but they are only running the insight job. Overall this is the part of the build that doesn't feel quite right to me.

tgsoverly: Typing out some ideas here.

I think if we are truly running insight jobs, then we should combine this into the "run project dag changes" and then get these changes into the project dag.

So if an exploration references other items, that would get resolved if the project.explorations.insight was picked up by the dag.  That feels like the way to do this.  I would just say that explorations feel like they are never serialized to disk at this point, maybe that changes, but it could change.  The question main question is what is new vs modified in the project.  If explorations are all copies and then they get merged, that can be handled, but I feel like we need to be clear about when the life cycle of the object.

If we do it this way, then there is only two "outside of dag" jobs.  1, is the "test-connection" which is running before a source is ever saved.  The other is the model's "run" button.

viewer/src/api/insightCompileDraft.js
tgsoverly: Already true in the merged code — both exports carry `Draft`: `compileDraftInsight` and `executeDraftInsight`. Resolving.

viewer/src/api/explorations.js
tgsoverly: Answered: core has no exploration concept at all today — the Django twin is VIS-1077 (Phase 7 — Cloud Enablement), against the frozen S3 contract, with VIS-1032 still open to freeze it. Tracked there rather than here. Resolving.

viewer/src/api/explorations.js
tgsoverly: Answered in-thread — record-promotion only logs that the promotion happened; the cache store still writes back to files. Resolving.

viewer/src/api/insightExecuteDraft.js
tgsoverly: Picking this up — leaving the thread open, since it's a decision rather than a fix.

Agreed on the direction: explorations become part of the project DAG, and a change to one submits a run scoped to it. Two things found while scoping it that are worth recording here:

1. **It reverses a guarded decision.** `visivo/models/exploration.py` opens with "Deliberately NOT part of the Project DAG / project YAML… This module must never be imported by `visivo.models.project`", and `tests/models/test_exploration.py` asserts exactly that (`test_project_module_does_not_import_exploration`, `test_exploration_module_not_in_project_schema_defs`). Fine to reverse, but it needs to be a deliberate call with those guards rewritten, not a refactor.

2. **Execution today is entirely client-side.** A chip's SQL runs through `model-query-jobs` into an in-memory job (1h TTL), then into browser DuckDB-WASM, where computed columns and metrics are evaluated. Nothing is written to disk. Only promote → commit → run produces artifacts.

So the decision taken is **hybrid**: the DAG run generates the durable assets, and the in-browser path stays for typing-speed feedback. Routing every keystroke through a run — pool claim plus poll in cloud — would trade a live preview for a spinner.

On "only two outside-of-DAG jobs": that lines up with the async-source-ops work. test-connection is now create-then-poll on both servers (visivo#546 / core#311), which leaves it and the model's run button as the two.

Open questions going into a design spike, mostly yours from above: whether explorations serialize to disk, what counts as new vs modified, and whether the DAG node is the exploration or its insights.

Cloud side is tracked as VIS-1077 (twin) / VIS-1032 (contract freeze).

PR #528, opened 2026-07-20
Phase 6c-T1: preview integrity & states

jpjsonic:
## Summary

Closes the 6c-T1 slice of `ux-audit.md` (cold-start #1/#2/#3, promote-roundtrip #1/#2/#3, pills #3, plus the two ⚠ conflicts-with-e2e findings assigned to this track: promote-roundtrip #9, existing-objects #8). Territory: `useDraftInsightPreview`, `usePreviewInputDependencies`, `ExplorerChartPreview.jsx`, preview state components, `insight_compile_views.py` (reviewed, no bug found — see disposition below), plus the exploration-staleness/promote-offer machinery for finding 5.

### 1. Draft-preview execution gap (BLOCKER) — FIXED
`useDraftInsightPreview.js` now checks every `compiled.models[]` entry has actually-fetched rows (`modelStates[name].queryResult.rows`) **before** ever sending `post_query` to DuckDB. A draft whose props compile without a schema lookup (see `insight_compile_views.py`'s own docstring — this is a real, documented server-side behavior, not a bug there) used to hand DuckDB a query against a table this hook never created, surfacing a raw `Catalog Error: Table with name <hash> does not exist!`. The guard now falls back to the existing `model_not_run` guided-empty state instead. A belt-and-braces regex catch also maps any hashed-table DuckDB error that somehow still slips through.
Evidence: `03-chart-tab-before-run-BEFORE.png`, `08-chart-tab-after-mapping-AFTER.png` (no Catalog Error anywhere in the whole walkthrough), `09-no-misclassified-input-banner.png`.

### 2. Infinite spinner — FIXED
Root cause found by tracing the actual code path: an insight with zero data-bearing props never even attempts a compile (correctly — nothing to preview), but the hook used to report a bare "not loading, not blocked" status, which `Chart.jsx`'s own `hasAllInsightData` gate then waited on **forever** (data that was never going to arrive). The hook now reports an explicit `no_data_props` blocked reason; `ExplorerChartPreview` renders a guided "Nothing to preview yet — drag a column..." state instead of ever mounting `<Chart>`.
Evidence: `03-chart-tab-before-run-BEFORE.png` / `04-...-after-6s-still-no-spinner.png` (`chart-preview-loading` count = 0 after a 6s wait, confirmed via the walkthrough script's own assertions).

### 3. Unresolved-"input" misclassification — FIXED
`usePreviewInputDependencies` now excludes published (`state.models`) **and** caller-supplied draft/not-yet-promoted model names (`ExplorerChartPreview` passes its own `explorerModelStates` keys) before computing `unresolvedNames` — a model ref written by a Build-rail pill drop (`?{${ref(model).column}}`) is textually indistinguishable from a real input ref without this exclusion. Banner copy for genuine unresolved inputs also drops "not yet promoted" (D11) for "isn't defined in this project yet." Verified the banner is absent both **before and after** promote (previously it persisted post-promote, per the audit).
Evidence: `09-no-misclassified-input-banner.png` (before promote), `12-chart-tab-after-promote-banner-cleared.png` (after promote, incl. after a fresh navigation back to the tab).

### 4. Error translation — FIXED
New `translatePreviewError.js` maps known error shapes (hashed-table catalog errors, `"has no dependent models"` DAG jargon) to plain-language headline + hint; anything unrecognized gets an honest generic headline. The raw message is **never discarded** — it moves to a collapsed "Technical details" `<details>` in `ChartPreview.jsx`, never the prominent copy. Unit-tested directly (`translatePreviewError.test.js`) since triggering these specific engine errors live wasn't reliably reproducible in a walkthrough without deep DuckDB manipulation.

### 5. ⚠ conflicts-with-e2e — post-promote offers & staleness — FIXED, root causes below
**Post-promote offers (promote-roundtrip #9).** Root cause: **condition too narrow**, not a fixture-reachable-state gap or discoverability problem. Both existing offers are genuinely correct for their own precondition — `dashboard-newchart-roundtrip.spec.mjs` proves "Place in `<dashboard>`" works when `return_to` is armed via the Library's dashboard-scoped "+ New Chart" button; `save-as-metric.spec.mjs` proves "View in Semantic Layer" works after an explicit Save-as-metric. Neither precondition is **ever** true for the single most common promote outcome (Explorer home → source tile → query → chart → Save to Project, which promotes a MODEL + insight + chart, never a metric/dimension, and never arms `return_to`). Fix: broadened `promotedField` to include a promoted `model` row (equally semantic-layer-visible), and added a fallback "Add `<chart>` to a dashboard" offer that fires whenever a chart was promoted with no `return_to` armed, reusing the existing `placeChartInDashboardSlot` plumbing.
Evidence: `11-promote-success-with-fallback-offer.png`, `11b-...-scrolled-into-view.png`, `11d-fallback-offer-accepted-dashboard.png` (chart actually lands in a real dashboard, backend-verifiable).

**Staleness (existing-objects #8).** Root cause: **the existing e2e-verified feature is real but narrower than the audit's expectation, by design** — `explorationStaleness.js`'s own prior "Scope" docstring said so explicitly: dangling-ref detection was implemented, "changed-vs-deleted distinction would need a `ref_fingerprint` persisted on the Exploration record — noted as follow-up work, not implemented here." That follow-up is this fix: `SeedRef.content_signature` (backend) captures a stable hash of the seeded object's config at `createExploration` time (centralized once, so every "Explore this" call site gets it for free); `computeExplorationStaleness` recomputes and compares on resume, additive with (never replacing) the dangling-ref check.
Evidence: `13-explore-this-context-menu.png`, `14-explorer-home-no-drift-yet.png` (before edit), `15-staleness-drift-banner-AFTER.png` (after editing the original insight via the real `/api/insights/<name>/` endpoint and reopening — banner reads "insight changed in the project since this copy was made", no "promoted" jargon).

New specs added under the DOUBLE-REGISTRATION RULE (grep-verified 2 occurrences each in `playwright.config.mjs`): `exploration-promote-fallback-dashboard-offer.spec.mjs`, `exploration-staleness-drift.spec.mjs`.

### 6. D11 vocabulary — FIXED in this track's surfaces
"Save to project" title unchanged/confirmed (`10-save-to-project-modal.png`); no "promote"/"input"/"MODEL" in the guided empty states, translated errors, or the new drift banner copy. Did not touch the promote modal's pre-existing "Promote N selected" CTA or checklist vocabulary (T3 wave 2's territory) beyond the banner-clearing/offer-broadening explicitly in scope here.

## Disputed / out of scope (found, not fixed — belongs to other tracks)

- Reproduced live, during the walkthrough: **drag-drop hit-testing offset** (dropping a column on the x well lands it in y) and **wells defaulting to a static/number-spinner mode** — both exactly the pills-buildrail/promote-roundtrip findings already assigned to 6c-T4 ("DnD integrity & build-rail cleanup"). Worked around in my own walkthrough script (documented inline) purely so the preview pipeline itself could be verified end-to-end; did not touch DnD code.
- `insight_compile_views.py`: reviewed per the grounding anchor — no additional bug found. The fix for finding 1 is entirely client-side per the audit's own fix direction ("Close the gap client-side").

## Test plan
- [x] `bash scripts/viewer-check.sh` — 339 suites / 5726 tests passing, lint clean
- [x] `PYTHONPATH=$(pwd) ./venv14/bin/python -m pytest tests/ -x -q` — 2134 passed
- [x] `./venv14/bin/black --check --target-version py312 .` — clean
- [x] Schema regenerated (`python -m visivo.generate_project_schema_json`) — no diff, `Exploration` is deliberately excluded from the project DAG schema (asserted by `tests/models/test_exploration.py`)
- [x] Standalone Playwright walkthrough against an isolated sandbox (own project copy, ports 8071/3071), screenshots in the session scratchpad, reviewed with vision — see per-finding evidence above
- [ ] New/changed e2e specs run in this PR (per instructions: they can't run against custom ports; verified by careful read + the walkthrough script instead — orchestrator runs the full suite at merge)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,11 @@ env/
 venv/
 venv12/
 venv14/
+# ...and the bare-symlink form: a trailing-slash rule matches directories
+# only, so a `venv14 -> /path/to/poetry-env` symlink slipped past it and got
+# committed by a `git add -A` (Wave 1), clobbering worktree venvs on checkout.
+/venv14
+/venv12
 cli_venv/
 api_venv/
 ENV/
--- a/tests/server/views/test_exploration_views.py
+++ b/tests/server/views/test_exploration_views.py
@@ -84,7 +84,11 @@ class TestCreate:
             json={"seeded_from": {"type": "model", "name": "orders"}},
         )
         assert resp.status_code == 201
-        assert resp.get_json()["seeded_from"] == {"type": "model", "name": "orders"}
+        assert resp.get_json()["seeded_from"] == {
+            "type": "model",
+            "name": "orders",
+            "content_signature": None,
+        }
 
     def test_create_with_draft(self, client):
         resp = client.post(
@@ -195,7 +199,11 @@ class TestUpdate:
         data = resp.get_json()
         assert data["id"] == created["id"]
         assert data["created_at"] == created["created_at"]
-        assert data["seeded_from"] == {"type": "model", "name": "orders"}
+        assert data["seeded_from"] == {
+            "type": "model",
+            "name": "orders",
+            "content_signature": None,
+        }
         assert data["promoted"] == []
         assert data["name"] == "Renamed"
 
--- a/venv14
+++ b/venv14
@@ -1,1 +1,0 @@
-/tmp/poetry-987-cache/virtualenvs/visivo-A-6GcZOt-py3.12
--- a/venv14
+++ b/venv14
@@ -1,0 +1,1 @@
+/tmp/poetry-987-cache/virtualenvs/visivo-A-6GcZOt-py3.12
--- a/viewer/e2e/stories/exploration-lifecycle.spec.mjs
+++ b/viewer/e2e/stories/exploration-lifecycle.spec.mjs
@@ -468,6 +468,13 @@ test.describe('Exploration lifecycle (Explore 2.0 Phase 2)', () => {
     await expect(page.getByTestId('workspace-middle-semantic-layer')).toBeVisible({
       timeout: 15000,
     });
+    // Wave-1 composed gate: the middle pane can commit a frame before the URL
+    // write lands under load (observed 1-in-3), so sampling `page.url()` at
+    // this exact instant is racy. Wait for the URL to BECOME the bare view
+    // route, then still assert the pathname exactly — same proof, not a
+    // weaker one. (That the pane can lead the URL at all is logged against
+    // 6c-T2 for a look at whether the write should be synchronous.)
+    await page.waitForURL('**/workspace/semantic-layer', { timeout: 10000 });
     expect(new URL(page.url()).pathname).toBe('/workspace/semantic-layer');
     await expect(page.getByTestId(`workspace-tab-exploration:${id}`)).toBeVisible();
 
--- a/viewer/e2e/stories/exploration-promote-fallback-dashboard-offer.spec.mjs
+++ b/viewer/e2e/stories/exploration-promote-fallback-dashboard-offer.spec.mjs
@@ -1,0 +1,257 @@
+/**
+ * Story: the post-promote "add to a dashboard" offer, reached the way the
+ * adversarial UX audit's auditor actually walked the surface (Explore 2.0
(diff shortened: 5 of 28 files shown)


jpjsonic:

PR #529, opened 2026-07-20
Phase 6c-T4: DnD integrity + build-rail curation (D12)

jpjsonic:
## Summary

Phase 6c-T4 — DnD integrity & build-rail cleanup, closing the audit's most
trust-destroying findings: drops landing in the wrong slot, silent
rejections, and a build rail that read like a raw Plotly-schema dump.

Root-cause note: three separate-looking blockers (cold-start #3's "70px
auto-scroll drift", promote-roundtrip #3's "drop on x lands in y", and
pills #1's "static-mode silently rejects the drop") turned out to share ONE
root cause — dnd-kit's built-in auto-scroll firing on the Build rail's
`overflow-y-auto` body mid-drag. Fixing that at the source (a
`data-dnd-freeze-scroll` opt-out on the shared `<DndContext>`'s
`autoScroll.canScroll`) resolved all three simultaneously; this was
confirmed empirically (166px drift → 0px, wrong-slot drop → correct-slot
drop) before I touched anything else.

### DnD integrity

| # | Finding | Fix | Evidence |
|---|---|---|---|
| 1 | **[cold-start #3 / promote-roundtrip #3, BLOCKER]** Rail auto-scroll during drag moves drop targets; column dropped on x lands in y | `WorkspaceDndContext`'s shared `DndContext` gets `autoScroll.canScroll` excluding any `[data-dnd-freeze-scroll]` ancestor; `ExplorationBuildRail`'s scroll body opts in | `before-02-mid-drag-DRIFTED.png` (149px drift, wrong content under cursor) vs `after-02-mid-drag-STABLE.png` (0px drift); `before-03-after-drop-SWALLOWED.png` (both x/y empty) vs `after-03-after-drop-LANDED.png` (y correctly filled) |
| 2 | **[pills #1, BLOCKER]** Drop on a "Static value" slot silently rejected | Root-caused to the SAME auto-scroll drift above — once frozen, the drop lands and (see #4 below) auto-switches to field mode | same before/after pair as #1; isolated confirmation in `PropertyRow.test.jsx` ("a droppable, query-capable EMPTY slot defaults to query mode…") |
| 3 | **[cold-start #3 / pills #1]** No visible feedback on a rejected/no-op drop | `InsightBuildSection.handleDropField` fires the shared workspace toast for every drag payload it can't act on (whole-table drop, unrecognized payload) | `13-rejected-drop-toast.png` — green "Can't drop a whole table here — drag a column instead." toast |
| 4 | **[pills #4]** No drag ghost; results-grid column drags mapped to nothing in `mapDragStartData` | Added `column`/`pill` branches + `ColumnDragPreview`/`PillDragPreview` overlays | `06-drag-ghost-results-grid.png` — a `# X` pill now follows the cursor (previously nothing rendered) |
| 5 | **[cold-start #3]** Eligible wells don't highlight on drag-start, only on hover | `PropertyRow` reads `useWorkspaceDrag()`; eligible droppable rows get a dashed ring the instant a compatible drag starts | visible in `06`/`07`-series screenshots (dashed ring around x/y while dragging) |
| 6 | **[pills #4]** Pills can't be dragged between slots | Pill is now also a drag SOURCE (`useDraggable`, `{source:'pill', sourcePath, raw}`); router + `handleDropField` move the resolved expression and clear the source slot | `09-pill-drag-mid.png` → `10-pill-drag-after.png`: x pill dragged onto y — x clears to the empty-well placeholder, y shows `SUM · model ▸ X` |
| 7 | **[pills #10]** Pill body click does nothing; only the 16px chevron works | Pill body is `role="button"`, click/Enter open the same `PillMenu` via an imperative `ref.open()` (chevron still works, `stopPropagation`s so the two don't double-fire) | `08-pill-menu-from-body-click.png` — menu opens from a click well outside the chevron |
| 8 | **[pills #5]** Preset menu opens off the bottom of the viewport | `PillMenuPopover` measures its own rendered height (`useLayoutEffect`) and flips to open upward when it would overflow | `14-pill-menu-collision-aware.png` at a 900px-tall viewport — full menu (8 presets + Custom aggregation + Save as metric + Remove) rendered entirely above the fold |

### Build-rail cleanup (D12)

| # | Finding | Fix | Evidence |
|---|---|---|---|
| 9 | **[grounding diagnosis #4]** Duplicate `Type` selector stacked over TracePropsEditor's own `TypeSelector` | Deleted the legacy `<Select>` in `InsightBuildSection`; `TypeSelector` is scoped `type-selector-${ownerName}` so it stays unambiguous with multiple insight sections stacked | `11-single-type-selector.png` — `type-selector` count 1, legacy `insight-type-select` count 0 |
| 10 | **[pills #2, BLOCKER]** Chart-type change silently wipes all pills | `preserveTraceProps` already carried forward compatible top-level props (x/y survive scatter→bar) — the wipe was specifically the deleted duplicate selector bypassing that path. Now also returns `dropped: string[]` so `TracePropsEditor` shows a dismissible warning naming what a genuinely incompatible switch (e.g. →Pie) removed | `12-type-change-warning.png` — "Switching to **pie** dropped this field not supported on that chart type: **X**." |
| 11 | **[pills #8/#9]** Raw schema counts ("0 of 1366 properties", "0/180", "+84 more") | `FieldGroup`'s badge hides at `presentCount === 0`; `SchemaEditor` gets an opt-in `hidePropertyCount` used by `ChartBuildSection`'s Layout Properties | `16-curated-sections-fresh.png` — no badges on any empty group header, no "0 of 1366 properties" line; confirmed via `bodyText.includes(...)` all `false` |
| 12 | **[D12]** "Chart: Untitled" header reads as a real, boring committed name | Unnamed-chart fallback renders italic/gray instead of solid text | visible in `16-curated-sections-fresh.png` header |
| 13 | **[promote-roundtrip #4 / pills #9]** Wells default to a number spinner that mangles a dropped/typed ref into "e1" | Droppable, query-capable EMPTY slots now default to query mode (the existing `RefTextArea` "Type @ to insert a reference" placeholder), not the static number `<input>` | `13-rejected-drop-toast.png` / `15-empty-wells-query-mode-default.png` — fresh x/y show the RefTextArea placeholder, not spin-button inputs |
| 14 | **[promote-roundtrip #5]** Raw JSON-schema regex shown as a validation message | `plotlyValidator.formatErrorMessage` recognizes the AJV `pattern` keyword and returns "Enter a query expression like `?{ref(model).column}`…" instead of the literal regex source | new unit test `plotlyValidator.test.js` validates against the REAL committed `scatter.schema.json` and asserts no `\?\{`, `^...$`, or `pattern "` substring ever appears |

Copy touched follows **D11** — "Save to project" (unchanged), no new user-visible "promote"/"input" language (toast/warning copy reviewed).

## Out of scope / left alone

Per the T4 charter I did not touch the rail's mount/outer container,
`ExplorationWorkbench`, `WorkspaceShell`, `RightRail`, or the preview hooks
— those are T2's and T1's territory respectively.

## Test evidence

- `bash scripts/viewer-check.sh` (from `visivo/`): **338 suites, 5708 passed / 2 skipped, 0 failed. Lint clean.**
- New/updated unit coverage: `WorkspaceDndContext.test.jsx` (column/pill drag mapping + overlay previews + property-zone pill-drop routing), `PropertyRow.test.jsx` (query-mode default for empty droppable slots, pill-as-drag-source, click/keyboard-to-open), `InsightBuildSection.test.jsx` (pill-to-pill move + source-clear, reject-toast on sourceTable/unrecognized payload, legacy-selector-removed regression guard), `TracePropsEditor.test.jsx` (scoped testid), `FieldGroup.test.jsx` (badge hidden at zero), `SchemaEditor.test.jsx` (`hidePropertyCount`), `plotlyValidator.test.js` (no raw regex leak against the real schema).
- UX walkthrough: standalone Playwright scripts driven with real `page.mouse` gestures (steps:15+, down/move/up — no `evaluate` shortcuts) against an isolated sandbox (`VISIVO_SANDBOX_NAME=uxt4`, ports 8074/3074, own `test-projects/integration` copy). Screenshots referenced above are in
  `/private/tmp/claude-501/-Users-jaredjesionek-visivo-Product/9b3f866e-0f25-4361-a48f-483a18803463/scratchpad/ux-fix-t4/`.
- `e2e/stories/exploration-build-rail.spec.mjs` and `trace-props-editor.spec.mjs` updated for the `type-selector-*` testid move (no assertions weakened — same control, scoped selector). Two of `pill-aggregation.spec.mjs`'s tests that don't depend on the hardcoded `:8001` backend-poll helper (`a confidently non-numeric column...`, `no raw ?{ or ${ syntax ever appears...`) were run directly against the sandbox and pass; the `waitForBackendDraft`-dependent tests in that file assume the standard 3001/8001 port pairing (pre-existing, unrelated to this change) and couldn't be exercised from my custom-port sandbox without violating the "never 8001/3001" sandboxing rule — the orchestrator's standard-port merge run covers them.

### DIFF
--- a/viewer/e2e/stories/exploration-build-rail.spec.mjs
+++ b/viewer/e2e/stories/exploration-build-rail.spec.mjs
@@ -217,7 +217,7 @@ test.describe('Exploration Build rail (Explore 2.0 Phase 3b)', () => {
     const insightName = await page.evaluate(
       () => window.useStore.getState().explorerChartInsightNames[0]
     );
-    const typeSelectTestId = `insight-type-select-${insightName}`;
+    const typeSelectTestId = `type-selector-${insightName}`;
     await pickSelectOption(page, typeSelectTestId, 'Bar');
     await expect(page.getByTestId(typeSelectTestId)).toContainText('Bar', { timeout: 5000 });
 
@@ -281,7 +281,7 @@ test.describe('Exploration Build rail (Explore 2.0 Phase 3b)', () => {
     // Switch to an indicator so `value` is a scalar-only slot (B13's
     // established fixture for this affordance, per
     // explorer-indicator-slice-authoring.spec.mjs).
-    await pickSelectOption(page, `insight-type-select-${insightName}`, 'Indicator');
+    await pickSelectOption(page, `type-selector-${insightName}`, 'Indicator');
 
     const tableRow = await expandSourceTable(page);
     const { locator: column } = await firstNumericColumn(page, tableRow);
@@ -325,7 +325,7 @@ test.describe('Exploration Build rail (Explore 2.0 Phase 3b)', () => {
     const insightName = await page.evaluate(
       () => window.useStore.getState().explorerChartInsightNames[0]
     );
-    await pickSelectOption(page, `insight-type-select-${insightName}`, 'Indicator');
+    await pickSelectOption(page, `type-selector-${insightName}`, 'Indicator');
 
     const tableRow = await expandSourceTable(page);
     const { locator: column, name: columnName } = await firstNumericColumn(page, tableRow);
--- a/viewer/e2e/stories/trace-props-editor.spec.mjs
+++ b/viewer/e2e/stories/trace-props-editor.spec.mjs
@@ -110,7 +110,7 @@ test.describe('Trace-Props Editor (VIS-1020)', () => {
     const editor = page.locator(EDITOR);
 
     // TypeSelector reflects the seeded type (react-select shows the label).
-    await expect(editor.getByTestId('type-selector')).toContainText(/Scatter/i, {
+    await expect(editor.getByTestId(`type-selector-${insightName}`)).toContainText(/Scatter/i, {
       timeout: 10000,
     });
 
@@ -138,7 +138,7 @@ test.describe('Trace-Props Editor (VIS-1020)', () => {
     // Drive the TypeSelector react-select: open its menu, type to filter to
     // "Bar", then click the option. The menu portals to document.body and its
     // options carry the brand classNamePrefix `vis-select__option`.
-    const typeSelector = editor.getByTestId('type-selector');
+    const typeSelector = editor.getByTestId(`type-selector-${insightName}`);
     await typeSelector.click();
     const combobox = typeSelector.getByRole('combobox');
     await combobox.fill('Bar');
@@ -148,7 +148,7 @@ test.describe('Trace-Props Editor (VIS-1020)', () => {
 
     // Type updated: the grouped title flips to the bar type, TypeSelector shows Bar.
     await expect(editor.getByText('Key fields (bar)')).toBeVisible({ timeout: 10000 });
-    await expect(editor.getByTestId('type-selector')).toContainText('Bar');
+    await expect(editor.getByTestId(`type-selector-${insightName}`)).toContainText('Bar');
 
     // Compatible prop preserved: `x` is valid in the bar schema, so its
     // PropertyRow is still rendered after the switch (preserveTraceProps).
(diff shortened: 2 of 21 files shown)


jpjsonic:

PR #530, opened 2026-07-20
Phase 6c-T2: single right rail (D6) + responsive shell

jpjsonic:
## Summary

Phase 6c-T2 — kills the "why are there two right rails?" bug (D6) and closes the 1100px responsive-shell BLOCKER, plus three related surface fixes assigned to this track.

### 1. D6 — the two-rails bug (code-grounding defect #1, shell-ia #1/#2) — CLOSED

**Root cause** (confirmed via code, matched the audit's own diagnosis): `RightRail`'s exploration branch never implemented D6 — `ExplorationWorkbench` mounted `ExplorationBuildRail` as an in-pane sibling of `CenterPanel`, while `WorkspaceShell` *also* unconditionally mounted the shell's own `RightRail`, which had no `exploration` branch and fell through to the generic "No editor for this object yet — coming soon" placeholder.

**Fix**: `RightRail`'s tab set is now derived per object type — dashboards get Outline+Edit, sources get Data+Edit, explorations get **exactly one tab, "Build,"** which mounts `ExplorationBuildRail` directly (threading `explorationId` from the active tab's `selectedItem`), and everything else stays Edit-only. `ExplorationWorkbench` no longer mounts `ExplorationBuildRail` at all — it's `CenterPanel` alone. The dead "coming soon" placeholder is now unreachable from any object type; a type with no real editor gets *no* Edit tab, not a placeholder one.

`ExplorationBuildRail`'s own internals (`InsightBuildSection`/`ChartBuildSection`/`TracePropsEditor`/`PillMenu` and their DnD) are untouched — only its outer container class changed (from an in-pane `w-96 border-l` panel to a `flex-1` fill of whatever the shell's resizable rail gives it), keeping that territory's testids and behavior intact for the parallel T4 builder.

| Before (1100px, audit evidence) | After (1100px) |
|---|---|
| before | after |

| After — rail open (1600px) | After — rail collapsed (1600px) |
|---|---|
| open | collapsed |

Exactly one right rail renders in every context I walked (project home, semantic layer, explorer home, open exploration, dashboard tab, chart tab, model tab, Runs) — verified from pixels, not testids.

### 2. Responsive shell — BLOCKER at 1100px (shell-ia #10, cold-start #2) — CLOSED

**Root cause**: the left rail (320px), the in-pane build rail (~360-384px), and the shell's own dead right rail (~360px) were all fixed-width with no floor on the canvas (`min-w-0`), so at 1100px the SQL editor/results/preview column was squeezed to a ~30px sliver with text shredded to one character per line.

**Fix**: `WorkspaceShell` now measures its own width (`ResizeObserver`) and gives the canvas a real, enforced minimum (480px, `CENTER_MIN_WIDTH`, via inline `style`, not a shrinkable Tailwind class). When both rails at their configured widths would push the canvas under that floor, rails auto-collapse — **library first, then the right rail** — until it fits. This is a one-way *nudge* reconciled against the existing manual collapse toggles (new `applyWorkspaceAutoCollapse` store action + `workspace{Left,Right}AutoCollapsedByShell` bookkeeping), so it never fights a rail the user opened or closed themselves, and widening the viewport back out un-collapses only what the shell itself collapsed.

Since removing the duplicate rail (finding #1) already reclaims ~360px on its own, most contexts now fit at 1100px with **zero** collapse — the exploration surface (single rail) needed only the library to auto-collapse; wider surfaces like the dashboard/model tabs didn't need to collapse anything at all at 1100px. See the 1100px column in every "after" screenshot above and below — none show a squeezed sliver.

Also fixed as a direct consequence: at true phone width (390px), the canvas now holds its real min-width instead of collapsing to zero (still not usable on a phone — the shell has no horizontal scroll, so most of it renders past the fold — but what *is* on-screen is legible instead of a near-zero sliver). Updated `ilc-mobile.spec.mjs`'s assertion mechanism accordingly (see Spec updates below).

### 3. SQL editor dead space (cold-start #7 / shell-ia) — CLOSED

**Root cause**: the editor's row always claimed 65% of the pane's height via a fixed flex ratio, regardless of content — for a one-line query that meant ~430-440px of empty dark void above a squeezed results area.

**Fix**: the top row now auto-sizes to the SQL's line count (`(lines + 2) * 19`, the same formula `ModelEditForm` already uses for its own Monaco instance), clamped between a sensible min and a 420px max, floored at 260px in wide mode so the chart preview beside the editor is never crushed by a trivially short query. Results get `flex: 1 1 auto` — the reclaimed space. Manually dragging the divider still works and takes over from the default the moment it's used. See the compact editor in the "after" screenshots above (~260px for an empty one-line query, vs. the audit's ~430-440px).

### 4. Tab-strip overflow (shell-ia #6) — CLOSED

**Root cause**: the `+` new-tab button lived *inside* the same `overflow-x-auto` container as the tabs, so once enough tabs opened to overflow the strip, `+` scrolled away with them and vanished — with no scroll affordance to discover the hidden tabs at all.

**Fix**: tabs now scroll in their own track; `+` is pinned outside it as persistent chrome (never scrolls away). Click-to-scroll chevrons appear at either edge exactly when there's more to see in that direction.

### 5a. Runs view dark-on-dark (shell-ia #7) — HARDENED (could not reproduce the literal bug live)

I could not visually reproduce a literal dark-on-dark render in my sandbox at this feature tip — direct URL load, SPA client-side navigation, and the narrow/compact TopNav variant all rendered correctly (light background, readable dark text) in both light and system-dark-mode Chrome. The app has no dark-mode media query or theme toggle anywhere in the viewer, so I could not find a code path that would literally invert to dark-on-dark. I'm flagging this honestly rather than claiming a fix for a bug I couldn't see.

What I *did* find and fix, defensibly regardless of root cause: the view depended entirely on `Home`'s ambient `bg-gray-50` wrapper for its background (no explicit background of its own at any entry point), and its destructive states used a hand-rolled Tailwind red instead of the shared `highlight` design token. Both are now fixed — every entry point (loading/error/loaded) sets its own explicit light surface, and the load-failure message, a failed run's badge, and its error label use `text-highlight`/`bg-highlight-100 text-highlight-700`.

runs

### 5b. Model tab dual-editor duplication (shell-ia #9) — CLOSED

**Root cause**: the model canvas rendered a second, read-only Monaco copy of the *exact same SQL* the right rail's `ModelEditForm` already edits — two editors, one of them locked, with no signal for a first-time user about which was real.

**Fix**: removed the duplicate editor from the canvas; it now shows a one-line hint ("Edit this model's SQL in the right rail — this pane runs it and shows results") pointing at the one real editing surface, keeping its source/fields chrome, Run button, and results table as its actual job.

model tab

### 6. D11 copy — no violations introduced

All new copy I wrote (the "Build" tab label, the model-tab hint, RunsView's unchanged strings) uses no "promote"/"input"/"MODEL" language. I did not touch the pre-existing "Promoted"/"Save to Project" copy inside `ExplorationBuildRail`'s internals — that vocabulary sweep is explicitly 6c-T3's job (Wave 2); touching it here would risk conflicting with that track's planned edit.

## Spec updates (structural moves this PR's fixes honestly changed)

- **`project-editor-narrow-pane.spec.mjs`**: pinned width moved from 1071px → 650px. With the responsive-shell fix, 1071px with default rail widths now leaves the middle pane ~650px+ wide — comfortably past ProjectEditor's own 620px/860px container-query breakpoints, so 1071px no longer exercises its narrow-layout behavior at all (the *good*, intended outcome of finding #2's fix). 650px is the new pinned width: even with BOTH rails auto-collapsed (the shell's own floor), the remaining container is still narrow enough to exercise the exact same container-query behavior this spec protects. Verified passing against my sandbox (4/4).
- **`ilc-mobile.spec.mjs`**: the "build canvas is not a phone surface" assertion mechanism changed from "canvas width < 40px" to "canvas holds its real min-width but extends past the 390px viewport with no scroll available" — the canvas no longer collapses to zero (it now honors `CENTER_MIN_WIDTH`), but is still genuinely unreachable past the fold on a real phone, just via a different mechanism. Verified passing against my sandbox (5/5).
- **`workspace-previews-visual-capture.spec.mjs`**: comment-only accuracy fix (non-assertive visual-capture tooling, no behavior change) — its docstring claimed the shell "does NOT reflow" on mobile, which is no longer true.

I also audited every other e2e spec with a viewport in the 620-1250px range (the range my auto-collapse logic could affect) and every spec pinned at the Playwright default (1280×720, ~120px of margin above the 480px canvas floor at default rail widths) — none of the others touch `/workspace/...` in a way my change affects; `table-wide-and-narrow.spec.mjs`'s narrow-width test uses the unrelated standalone `/project/...` route, not `WorkspaceShell`.

**A pre-existing, unrelated gap found (not fixed, out of scope)**: `right-rail-routing.spec.mjs`'s "project-chrome selection → DefaultsEditForm" test fails against this branch — traced to `openWorkspaceTab`'s `isWorkspaceView('project')` routing (views left the tab model in an earlier Phase-0 change I never touched), which the test predates. Confirmed via direct code tracing this is unrelated to any file in this diff.

## Verification

- `bash scripts/viewer-check.sh` (full suite): **339 suites / 5720 tests passing, 2 skipped, lint clean.**
- UX walkthrough: standalone Playwright script (`chromium`, `require('@playwright/test')`) against an isolated sandbox (ports 8072/3072, `VISIVO_SANDBOX_NAME=uxt2`), screenshotting every context at 1600px **and** 1100px — project home, semantic layer, explorer home, open exploration (rail open + collapsed), dashboard tab, chart tab, model tab, Runs. Judged from the rendered pixels: exactly one right rail everywhere, no dead panels, canvas usable at 1100px in every context, no dark-on-dark.
- Targeted e2e re-runs against the same sandbox: `right-rail-routing.spec.mjs` (7/8, 1 pre-existing unrelated failure — see above), `exploration-build-rail.spec.mjs` (4/6 — the 2 failures are a hardcoded-port artifact in the spec's own `apiBase` helper against my non-default sandbox ports, not app behavior; all UI-mechanics assertions passed), `project-editor-narrow-pane.spec.mjs` (4/4), `ilc-mobile.spec.mjs` (5/5).
- Full e2e suite intentionally **not** run — orchestrator gates that at merge per the standing instructions.

## Notes for reviewers

- `.github/pr-evidence/6c-t2/` holds the before/after screenshots referenced above (the "before" shot is the audit's own evidence at the same feature-tip commit); safe to delete post-merge.
- Territory respected: `ExplorationBuildRail`'s internals (`InsightBuildSection`/`ChartBuildSection`/`TracePropsEditor`/`PillMenu`) were not touched — only its outer container/mount, per the T2/T4 split.


jpjsonic:

PR #533, opened 2026-07-22
Create seeds and move csv script model into seeds and remove local merge model

tgsoverly:
## What & why

Visivo had three model types behind one `models:` list, discriminated by *shape*:
`SqlModel` (`sql:`), `CsvScriptModel` (`args:`), and `LocalMergeModel` (`models:`).

The two extra types cost far more than they earned. Each dragged a parallel job, a
byte-identical server manager, a byte-identical view module, six enumeration lists,
a Zustand store, an API module, an edit form, and a lineage node — and neither could
carry `metrics:`/`dimensions:`, so both were second-class.

This PR collapses everything to **one model type**:

- **`LocalMergeModel` is removed outright.** Its only real feature (a model→model DAG
  edge for cross-DuckDB joins) is unused in the repo. Models that share a source now
  join in plain SQL.
- **`CsvScriptModel`'s capability moves onto the *source* as a `seeds:` list** — a
  command whose stdout CSV is loaded into a named table before any model queries it.

## Why seeds live on the source, not the model

This is the decision that made the change small:

- Seeds run once **per source**, inside the existing source job, **before**
  `get_schema()` — so schema introspection sees the seeded tables.
- The Model→Source DAG edge already orders the source ahead of every model, so there's
  **no `depends_on`** and no new edge kind. Models sharing a source join its seeded
  tables in plain SQL.
- **`SqlModel` gains zero fields.**
- **No discriminator risk** — sources discriminate on `type:`, not shape.

`seeds` sits on `ServerSource` (all 8 database sources), *not* `Source`. `csv`/`xls`
file sources — which build a throwaway in-memory connection where a seed would silently
vanish — inherit only `BaseDuckdbSource` and reject the key via `extra="forbid"`.

```yaml
sources:
  - name: scratch
    type: duckdb
    database: target/scratch.duckdb
    seeds:
      - args: [cat, data/orders.csv]
        table_name: raw_orders
      - args: [python, fetch_items.py]
        table_name: raw_items

models:
  - name: orders
    source: ${ref(scratch)}
    sql: select * from raw_orders o join raw_items i on o.id = i.order_id
```

Adds `Source.write_dataframe`: DuckDB writes through its raw connection; SQLAlchemy
sources through SQLAlchemy Core (polars' own `write_database` needs pandas, which
Visivo doesn't depend on).

## Migration

`visivo migrate` now does **structural** YAML rewrites via ruamel (comments, quoting,
and block scalars survive), plus `--include-markdown` for fenced YAML in docs. A csv
script model becomes a per-model DuckDB source + a plain model — a 1:1 reproduction of
today's behavior. Local-merge models are **reported, not rewritten** (folding several
models' SQL is a judgement call).

This PR migrates the repo's own fixtures/docs: **136 csv script models across 127
project files** and **144 more across 48 docs pages**. The three local-merge models
were folded by hand.

```
$ visivo migrate
models.visivo.yml:
  model 'csv' (csv script) -> source 'csv-source' with a seed, plus a sql model
  ...
1 item(s) cannot be migrated automatically:
  model 'join_table' (local merge)
    -> rewrite as a sql model whose own SQL does the join, with every input
       seeded onto one shared source
```

## Two latent bugs fixed in passing

- **`project_injection.py`** was `setattr`-ing `csv_script_models` / `local_merge_models`
  onto `Project`, which has neither field and is `extra="forbid"` — reachable whenever
  a user edited one of these types in the editor. Removed.
- A **sourceless model in a project with no `defaults`** crashed with `AttributeError`
  while building the DAG, instead of reaching `ModelsHaveSourcesValidator`'s intended
  "does not specify a source" message. Guarded.

## Verification

- **Behavior is preserved, proven byte-for-byte.** Ran `complex-project` on pre-refactor
  `main` in a worktree and diffed the output: all six insights produce **byte-identical
  parquet**, including the two that previously ran through `LocalMergeModel`s.
- **Seeds land before introspection, confirmed in the live app.** The source canvas
  renders the seeded `coffee_export` table with introspected columns, zero console/API
  errors (Playwright against a local sandbox).
- **Python:** `2029 passed, 0 failures` (baseline `1994`, after deleting ~40 obsolete
  tests and adding new ones for `Seed`, `write_dataframe`, the source job, and the
  migration). `black --check` clean.
- **Viewer:** `314 suites / 5152 tests` passing, lint clean. New `SeedsEditor` wired into
  `SourceEditForm`.
- Docs build clean; all three committed schema snapshots + the mkdocs nav regenerated.

## Frontend

Deletes the two Zustand slices, two API modules, two edit forms, and two lineage nodes,
and prunes every registry that enumerated them. Adds a `SeedsEditor` on the Source form
(the model form needed no change) — gated on the same `ServerSource` boundary the backend
enforces, and omitting `seeds` from the saved config entirely when empty.

## ⚠️ Cross-repo coordination required before merge

`serializer.py` no longer emits the `csv-script-models` / `local-merge-models` deploy
segments — everything ships under `models`. **Confirm `core`'s deploy handler accepts the
merged type under `models`** before this merges. This is the one cross-repo risk.


mkdocs/concepts/model.md
```diff
@@ -41,14 +39,16 @@ insights:
 
 ## Model types
 
-| Type | Use when |
-|------|----------|
-| [SqlModel](../reference/configuration/Models/SqlModel/index.md) | You want to write a SQL query against one Source. |
-| [LocalMergeModel](../reference/configuration/Models/LocalMergeModel/index.md) | You want to join tables that live in **different** Sources. |
-| [CsvScriptModel](../reference/configuration/Models/CsvScriptModel/index.md) | Your data comes from a script, API, or other non-SQL process. |
+There is one Model type — [SqlModel](../reference/configuration/SqlModel/index.md) — a
```
jpjsonic: I wonder if we really need to clarfify that there's one model. I think that's what users will actually expect.

mkdocs/concepts/model.md
```diff
@@ -41,14 +39,16 @@ insights:
...
+SQL query against one Source.
+
+When your data comes from a script, an API, or another non-SQL process, give the Source a
```
jpjsonic, line 42: This also made me think that we probably need to update our onboarding.jsx to refelect these concept changes as well along with our seeded onboarding flow projects.

viewer/src/components/views/lineage/CsvScriptModelNode.jsx
jpjsonic, line 1: Don't we need to add a seed node?

visivo/models/sources/source.py
```diff
@@ -134,6 +149,10 @@ class ServerSource(Source):
     db_schema: Optional[StringOrEnvVar] = Field(
         None, description="The schema that the Visivo project will use in queries."
     )
+    seeds: List[Seed] = Field(
```
jpjsonic, line 162: Are you thinking that a seed will only be able to have one child source?

jpjsonic (approved):
Some questions in my comments, looks great though! This is a much clearer model

viewer/src/components/views/lineage/CsvScriptModelNode.jsx
tgsoverly: I guess I could break it out to its own job.  Right now it runs on the source.  Probably worth it.

mkdocs/concepts/model.md
tgsoverly: Good point.

visivo/models/sources/source.py
tgsoverly: For now, we can break it out if needed, pretty simply

PR #536, opened 2026-07-22
Phase 6c-T3: vocabulary & copy pass (D11) — Save to project, naming step, live chip naming

jpjsonic:
# Phase 6c-T3 — vocabulary & copy pass (D11)

Implements D11: the single user-facing verb for the exploration→project chain is
**"Save to project"** (CTA `Save to project…`; trail `Saved to project · pending
commit`). "Promote" is internal-only now (store actions, API routes, test ids —
unchanged). Closes the ux-audit.md findings assigned to this track: competing
verbs, no naming step in Promote, raw regex validation, dev-jargon errors, dead
query-chip controls, tiny update-existing signal, close-tab dialog naming by
internal id, duplicate PROMOTED ledger entries, unexplained disabled "Save as
metric", dead "Project" pill, context-blind footer text.

## What changed

- **`promoteNaming.js`** (new): pure functions (`isGenericPromoteName`,
  `suggestPromoteNames`) that recognize placeholder names (`query_1`/`model`,
  `insight`/`insight_N`, `chart`/`chart_N`) and propose real ones, anchored on
  the object's bound source or its sibling in dependency order.
- **`ExplorationPromoteModal.jsx`** — naming step: every brand-new row gets an
  editable name field, pre-filled with the suggestion above; manual rename on
  blur with human validation copy and collision handling. Vocabulary: `Save N
  to project ▸`, `Saved N objects to project.` + `Saved to project · pending
  commit` signpost. "updates existing" pill recolored amber with an icon +
  explainer banner.
- **`ExplorationBuildRail.jsx`** — CTA `Save to project…`; trail heading `Saved
  to project` with a pending-commit caption; PROMOTED ledger deduped by
  `type:name` (was showing the same object twice). **New in this pass**: a
  live auto-suggest effect (see "post-merge finding" below).
- **`TabCloseConfirmDialog.jsx`** — names the exploration by its real display
  name, never the raw `exp_<id>`.
- **`PillMenu.jsx`** — visible hint for why "Save as metric" is disabled.
- **`TopNav.jsx`** — tooltips on non-interactive Stage/Project capsule segments.
- **`library/Library.jsx`** — `libraryFooterHint(scope)`: context-aware footer
  copy (exploration / explorer / dashboard / semantic-layer / default) instead
  of the hardcoded dashboard-canvas line on every surface.
- Vocabulary sweep: `ExplorationCard.jsx`, `ExplorerHomePane.jsx`,
  `FieldSwapOfferBanner.jsx`, `ExplorerChartPreview.jsx`.

## Post-merge walkthrough finding — chip labelled "model"/"insight" during live editing

Wave-1 merged clean into this branch (all conflicts were "keep both" —
neither side's coverage tests were dropped; two duplicate `describe` titles
from the auto-merge were disambiguated with a `(coverage-push)` suffix,
tests unchanged). A walkthrough on the merged tip surfaced the single most
visible thing left in this phase: the naming-step suggestion only ran when
the Save-to-Project modal opened, so a query chip / insight sat on screen
literally labelled `model`/`insight` for the entire life of the draft up to
that point.

**Fix**: `ExplorationBuildRail.jsx` now runs the same suggestion logic live —
a `useEffect` that, on every relevant state change, renames any
placeholder-named model tab that has a bound source (`<source>_query`) or
placeholder-named insight whose model already has a real name
(`<model>_insight`), via the same `renameModelTab`/`renameInsight` store
actions the modal already used. Self-terminating (once renamed, the name is
no longer generic so the effect stops touching it) and safe by construction:
`renameModelTab`/`renameInsight` already refuse to touch anything that isn't
a brand-new draft (their own `isNew` guard).

**Territory note**: this reads `explorerModelTabs`/`explorerModelStates` and
calls `renameModelTab`/`renameInsight` — both already-existing, already-owned
by T5 store actions — but adds no new logic to `explorerStore.js` itself. The
new code lives entirely in `ExplorationBuildRail.jsx` (mine). Flagging per the
territory rule since "naming mechanics" is nominally T5's, even though no
T5-owned file was touched.

**Real bug this fix's first version had, caught by e2e, now closed**: the
live-naming effect could rename a cold-session model tab off its TEMPORARY
"first available source" fallback (VIS-1082) *before*
`useExplorerWorkbenchInit.js`'s `applyResolvedDefaultSource` rebind effect
corrects it — that rebind looks the tab up **by its original name**, so once
this effect renamed it first, the rebind's lookup silently missed and the
real project default was never applied. `explorer-cold-session-default-source.spec.mjs`
caught this in a focused e2e run against this branch's own sandbox (not
theoretical — an actual failing assertion, `source-selector` stuck on
`local-sqlite` instead of correcting to `local-duckdb`). Fixed by gating
model-tab renaming on `defaults` having already loaded — the same signal the
rebind effect keys off — with `defaults` also in the effect's own dependency
array so a tab whose fallback source coincidentally already matched the real
default (the common case; the rebind itself no-ops there, so nothing else
would otherwise re-trigger this effect) still gets suggested a name once
defaults land. Two regression tests pin both directions.

## FIXED — a latent PillMenu/SliceBadge overlap bug this PR's own longer names first triggered

`exploration-build-rail.spec.mjs`'s "toggling a preset (SUM -> AVG) on a slot
with an authored slice" test was failing on this branch: once a value pill's
referenced query name grew long (fails at 19 chars `local-duckdb_query`,
passes at 3 chars `abc`), clicking the pill's chevron never opened `PillMenu`
because `SliceBadge` rendered directly over the trigger — confirmed via
bounding-box comparison and `document.activeElement` after the click.

**Framing, corrected from an earlier version of this description.** The
underlying CSS weakness is older than this PR, but nothing triggered it while
scratch queries were named `model`: the D11 live auto-rename in *this* PR is
what makes names long during editing, so from a user's point of view this is a
regression this PR surfaces, not inherited debt. Verified both directions:
`exploration-build-rail.spec.mjs` passes 6/6 on `feature/explorer-workspace`
without these changes (name is `model`, nothing overflows) and failed on this
branch until the fix below.

**The fix (commit `a905dacf`).** It is a flex `min-width: auto` chain, and a
single `min-w-0` genuinely didn't resolve it — every level has to opt out:
1. `FieldPill`'s outer Tag → `min-w-0 max-w-full`;
2. the label span → `flex-1 min-w-0` (so it claims remaining space AND may
   shrink below its own text, which is what finally lets `truncate` engage);
3. `PropertyRow`'s value-slot column → `min-w-0` in place of `min-w-[180px]`.

Long-label regression tests were added to `FieldPill.test.jsx` (the docstring
had claimed such tests already existed; it did not — now it does), and they
were verified to falsify: dropping `flex-1 min-w-0` from the label fails the
truncation test, restored it passes. `exploration-build-rail.spec.mjs` now
passes 6/6 on this branch.

## Footer-copy finding — confirmed fixed on this branch; the discrepancy was which branch got tested

I initially reported this as "already fixed by the Wave-1 merge" after
tracing `useWorkspaceScope.js` and reproducing several trigger scenarios that
all resolved correctly. The coordinator re-measured on `feature/explorer-workspace @ a2ecf4cd`
and found the dashboard-canvas line unconditionally on both the Explorer Home
and an open exploration — correctly calling out that "couldn't find a trigger
path" and "isn't broken" are different claims.

Re-checked with `git diff a2ecf4cd..HEAD -- .../library/Library.jsx`: the
entire `libraryFooterHint` function (and the `data-testid="library-footer-hint"`
wiring) **does not exist at `a2ecf4cd` at all** — that commit still has the
old unconditional hardcoded line. It only exists on this branch, added in an
earlier session before any Wave-1 merge landed. My "already fixed by Wave-1"
attribution was simply wrong (Wave-1 never touched this file for this
reason) — but the fix itself has been present and correct on this branch the
whole time. The coordinator's repro was against the pre-my-PR base, which is
exactly what you'd expect to still show the bug.

Re-verified on this branch after a full sandbox restart (fresh dev server,
not reused/HMR'd) with `localStorage.clear()` before each navigation, reading
`[data-testid="library-footer-hint"]`'s real `textContent` from the DOM (not
a screenshot) on both surfaces:
- Explorer Home (`/workspace/exploration`, no tab open): *"Click a source
  tile to start exploring, or click an existing object to explore it."*
- An open exploration (`/workspace/exploration/exp_...`): *"Drag a column
  onto a chart field to map it. Click a data object to add it to your
  exploration."*

Neither contains "Drag a layout item onto the canvas." `useWorkspaceScope.js`
was not modified — the fix is entirely `Library.jsx`'s `libraryFooterHint`,
already in this diff.

## Coverage (shared `jest.cov.t3.json` config, `--runTestsByPath`, whole-file stmts+branches)

| File | Stmts | Branch | Status | Note |
|---|---|---|---|---|
| `promoteNaming.js` | 100% | 97.61% | ✅ | line 93: unreached defensive branch |
| `ExplorationPromoteModal.jsx` | 98.89% | 95.09% | ✅ | |
| `ExplorationBuildRail.jsx` | 100% | 100% | ✅ | |
| `TabCloseConfirmDialog.jsx` | 100% | 96.66% | ✅ | |
| `PillMenu.jsx` | 97.81% | 94.36% | ⚠️ below 95 | see below — only remaining sub-95 file |
| `TopNav.jsx` | 99.03% | 95.02% | ✅ | |
| `FieldSwapOfferBanner.jsx` | 97.36% | 96.77% | ✅ | |
| `homes/ExplorationCard.jsx` | 100% | 100% | ✅ | |
| `homes/ExplorerHomePane.jsx` | 100% | 100% | ✅ | was 87.8% branch — see "closed" note below |
| `library/Library.jsx` | 100% | 98.16% | ✅ | |
| `explorer/ExplorerChartPreview.jsx` | 100% | 100% | ✅ | |

**`ExplorerHomePane.jsx` — closed from 87.8% to 100% branch** (the
coordinator flagged this as the one number worth pushing further, since it's
the first surface a user lands on):
- The `stalenessById[exploration.id]?.danglingRefs || []` fallback was
  provably dead (the map is built by iterating the SAME
  `orderedExplorations` array this JSX renders from, in the same `useMemo`
  pass — every id rendered here is guaranteed an entry) — simplified away
  entirely rather than tested around, closing that branch by construction.
- The `creatingRef.current` double-click guards (real `fireEvent.click`
  calls can't outrun them in jsdom OR a real browser — RTL auto-`act()`-wraps
  each `fireEvent` call, so the DOM's own `disabled` attribute is already
  applied before a second click could ever be dispatched, in every
  environment) are now proven directly: a new pair of tests invoke the React
  `onClick` handler straight from the fiber, twice, in the same synchronous
  tick, bypassing DOM dispatch (and the `disabled` gate) entirely — the only
  way to exercise `creatingRef.current` itself rather than the DOM gate that
  normally makes it moot.
- The other previously-uncovered branch (both handlers' `if
  (seedPromiseRef.current) await seedPromiseRef.current;`) turned out to be
  a real, distinct, testable scenario I'd misread as part of the
  double-click guard: the lazy auto-seed effect's own in-flight
  `createExploration()` racing a manual "+ New exploration"/source-tile
  click on the same empty list. Two new tests drive that race for real
  (a controllable promise for the seed, click while it's pending, assert the
  manual create is parked behind it, resolve the seed, assert the manual
  create then fires) — genuine coverage, not a synthetic branch-count hack.

**`PillMenu.jsx` (94.36% branch) — the one file still below 95, honest
reasons, no padding:**
- `MenuRow`'s `disabled`/`disabledReason` props (lines 124/125/129): never
  actually passed `true` by either of the two call sites in this file —
  genuinely dead through the current UI, not just untested.
- Two `typeof window !== 'undefined'` SSR guards (lines 231, 320): `window`
  always exists in jsdom; the `else` branch is unreachable in this test
  environment by construction, not by gap in coverage.
- A `!containerRef.current` guard (line 319): the ref'd element always
  renders unconditionally before the effect runs, so the ref is always
  populated — same class as the SSR guards above.
- `showPreset('dimension')` (line 350) and the "neither restricted nor
  numeric-only" branch (line 351): `showPreset` is only ever called via
  `PRESET_AGGREGATIONS.filter(showPreset)`, and (a) `'dimension'` is never a
  member of `PRESET_AGGREGATIONS` (it renders via its own hardcoded
  `MenuRow`, not the filtered list), and (b) every member of
  `PRESET_AGGREGATIONS` is already in `RESTRICTED_AGGS` or
  `NUMERIC_ONLY_AGGS` — both branches are unreachable given the current data,
  not just untested paths.

No test in this diff asserts on a mock without a corresponding real
assertion — every "why not 95%" above is a specific code-level reason, not a
placeholder.

## Gates

- `bash scripts/viewer-check.sh` (run fresh, after the final commit, post
  both merges below): **341 suites / 6293 passing, 2 skipped, lint clean.**
  (One flake in `ExplorationPromoteModal.test.jsx` on one full-suite run,
  reproduced clean twice in isolation and clean on a second full-suite
  run — not a regression, matches that suite's own documented timing
  sensitivity under parallel load.)
- `bash scripts/check-e2e-sandbox-isolation.sh`: **exit 0** ("backend ports
  and frontend bases both honor the sandbox").
- **Merged both `jared/e20-computed-column-preview` (PR #534, per the
  coordinator's "don't block on it landing" call) and, once it landed,
  `origin/feature/explorer-workspace` @ `e42ca535`** — clean, no conflicts.
- e2e: focused, single-worker run against this branch's own `:3062` sandbox
  (the 9 specs most relevant to this diff — exploration-build-rail,
  exploration-lifecycle, exploration-promote×3, exploration-query-chips,
  explorer-cold-session-default-source, explorer-home,
  workspace-tab-close-dialog-navigation): **44 passed, 1 failed** — the 1
  failure was the PillMenu/SliceBadge overlap bug documented (now fixed in a905dacf)
  above (confirmed independent of this PR's changes), not a regression.
  An earlier multi-worker attempt was abandoned mid-run after machine load
  hit ~41 from several agents running suites concurrently — not treated as
  evidence either way, per the coordinator's guidance — but it's what
  surfaced the real VIS-1082 race in the first place, since re-running it
  cleanly is what would have hidden a genuine, since-fixed bug.
- UX screenshot walkthrough: hardened, throw-on-failure script (no
  screenshot-reading, no soft `.catch()` swallowing on anything asserted) —
  **24/24 assertions passed**, all against real DOM state (`textContent`,
  `inputValue`, live store reads), covering every D11 vocabulary/naming
  surface end to end: Explorer Home footer, live chip naming, in-editing
  footer, the Save CTA, the naming step (every field), the banned-word
  sweep, the submit button, the success + pending-commit signposts (two
  separate elements — confirmed both render), the trail, and the close-tab
  dialog's real display name. Also captured an honest before/after (briefly
  reverted `ExplorationBuildRail.jsx`'s live-naming effect to its pre-D11
  state, screenshotted a fresh exploration showing the literal
  "model"/"insight" chip, restored the fix, screenshotted the same flow
  again) rather than only screenshotting the fixed state.

## Do not merge

Opening against `feature/explorer-workspace` per instructions — not merging.


jpjsonic:

PR #537, opened 2026-07-23
Phase 6c-T5: lifecycle & entry points (Explore 2.0 Wave 2)

jpjsonic:
# Phase 6c-T5: lifecycle & entry points

UX-audit-driven pass over the Explorer/Workspace lifecycle and entry points (Explore 2.0 Phase 6c, Wave 2). Closes the audit's Lifecycle/Entry-points/Persistence findings, plus 9 additional bugs found while chasing down e2e failures and closing coverage gaps to 95%+ per touched file. Full gate results, per-file coverage breakdown, and a deferred-with-diagnosis root cause writeup below.

## Findings closed

**Lifecycle**
- [MAJOR] Clicking a source tile silently creates a new persistent exploration every time — `explorationLifecycle.js` (`isExplorationVisibleInGallery`); seeded creates stay hidden from the Home gallery, and are GC'd if their tab closes untouched (`workspaceStore.js`'s `closeWorkspaceTab`).
- [MINOR] Every source-tile click mints a new auto-named exploration — same fix; litter is now invisible and self-cleaning.
- [MINOR] Phantom "Scratch" exploration — lazy auto-seed removed entirely; `ExplorerHomePane` shows an honest empty state.
- [MAJOR] "Explore this" always spawns a new exploration — silent proliferation — same ephemeral-until-edited mechanism applies to every seed type.
- [MAJOR] Seeded exploration naming is incoherent — deterministic `"<seed> exploration"` default name (`seedDefaultName`), plus the pre-existing visible rename affordance (ExplorationCard kebab + ExplorationPane SubBar pencil) already covers "coherent naming + visible rename".
- [Addendum VIS-1102] Untouched exploration offered for saving — `promoteChecklist.js` now excludes scaffold insights/charts with no authored content; empty checklist → "No changes to save." reachable state restored in `exploration-promote-fallback-dashboard-offer.spec.mjs`.

**Entry points**
- [MAJOR ⚠ conflicts-with-e2e] No "Explore this" from Semantic Layer ERD — added header "Explore" button (always visible) + right-click context menu + left-click select. New spec: `explore-this-semantic-layer-erd.spec.mjs`.
- [MAJOR ⚠ conflicts-with-e2e] No "Explore this" from a dashboard chart's context menu — added to `CanvasContextMenu.jsx`. New spec: `explore-this-dashboard-chart.spec.mjs`.
- [MAJOR] "Explore this" discoverable only via right-click/kebab in Library — added hover-revealed "Explore" icon button, a peer of flip/kebab, in `LibraryRow.jsx`.
- [MINOR ⚠ conflicts-with-e2e] Source name click hijacks navigation — row body now expands in place; explicit "Open" icon button navigates. New tests in `library-source-drilldown.spec.mjs` driving the row body directly.
- [MAJOR] "+ Add Insight" creates blank instead of picking existing — `ExplorationBuildRail.jsx` now opens a searchable picker (existing insights primary, "New blank insight" secondary).
- [MINOR] Metric "Add to exploration" spawns another blank insight — `addObjectToActiveExploration` now extends the active insight and syncs the model tab strip.

**Persistence**
- [MAJOR ⚠ conflicts-with-e2e / MINOR dup] Reload discards recent SQL edits (debounce race) — `ExplorationPane.jsx` flushes both debounces on `visibilitychange`/`pagehide`/`beforeunload`. New spec: `exploration-reload-persistence.spec.mjs` (also covers the x/y-mapping-lost-on-reload finding under the same root cause).
- [MAJOR ⚠ conflicts-with-e2e] x mapping lost after reload — same fix + spec above (root-caused as the same debounce-race window, not a separate DnD bug).

**Deferred (with rationale — not silently dropped)**
- [MAJOR] Query results/preview don't restore on reopen — the codebase deliberately excludes query results from the persisted draft (`snapshotExplorerWorkingState`'s own docstring: "large, ephemeral... re-running is cheap"). A safe fix (auto-rerun-on-restore or a "Run to restore" prompt) touches the query-execution pipeline (`SQLEditor.jsx`/`useModelQueryJob`/`CenterPanel.jsx`) deeply enough that I did not want to ship it without full e2e verification, which was blocked for part of this session (shared-sandbox port bug, see below). Left for a fast follow-up.
- [MAJOR] View/destination switching discards working context (Cmd+1/2/3 always land on destination Home, not a return to the parked tab) — this is arguably intentional per 01-ux-spec.md §1's documented "Park" behavior (clicking a view releases the center to Home by design); the audit's complaint is narrower ("returning" to a destination where a tab is already open should restore it). Same reasoning as above — deferred rather than rushed.
- [POLISH] Explorer gallery cards have no visual preview / sort / search — skipped, time-boxed out of this phase.

## Note re: territory
- D11 vocabulary (Save-to-project copy, promote modal naming step, PROMOTED ledger copy) is T3's — I did not touch user-visible strings in that chain. My `promoteChecklist.js` change is CONTENT gating (which rows exist), not copy.
- D6 (RightRail exploration branch / two-rails fix) was already landed by T2 before this branch was based — confirmed via `RightRail.jsx` already having the `exploration` branch mounting `ExplorationBuildRail`.

## Gate results (final — post-merge of origin/feature/explorer-workspace @ e42ca535 AND the VIS-1107 fix @ fac3940c; all numbers below are from runs AFTER fac3940c, the actual final commit)
- `yarn lint`: 0 errors.
- `yarn test --watchAll=false`: 341/341 suites, 6377/6379 passed, 2 skipped (pre-existing, unrelated). Includes the new `ExplorationPane.jsx` VIS-1107 test (falsified against the old guard before trusting it) and its 36 sibling tests, all passing. One `WorkspaceDndContext.test.jsx` timeout was seen ONCE mid-session, traced to a concurrent, unrelated Playwright session the user was running by hand against their own `:3001` sandbox (confirmed via `ps aux`, not assumed) — reproduced clean in isolation twice (124/124 both times) despite load averages of 34–50 at the time; a contention flake, not a regression (see "Known session incidents" below for the standard held throughout before calling anything a flake).
- `bash scripts/check-e2e-sandbox-isolation.sh`: exit 0.
- Coverage (whole-suite, shared cov-config, all 15 diff files) — see the full per-file, per-branch table below for the current numbers and every remaining gap's specific reason.
- **e2e (`exploration-mutations` project, 122 specs) against my own sandbox :8072/:3072: 122/122 GREEN (13.3m), zero failures.** This is the genuinely fresh run taken after the VIS-1107 fix landed — a prior run (same scope, pre-fix) surfaced exactly the `:129` failure VIS-1107 diagnoses, plus one contention flake in an unrelated spec (`exploration-promote-fallback-dashboard-offer.spec.mjs:173`, re-confirmed 2/2 clean in isolation before being called a flake). An EARLIER run than that was deliberately stopped mid-flight at 46/122 (all green) the moment the origin/feature/explorer-workspace re-merge was requested, rather than let it finish and then have to void it per the same "don't trust a run spanning a mid-flight merge" rule established earlier in this track.
- UX screenshot walkthrough: 16 screenshots reviewed pixel-by-pixel (scratchpad/t5-walkthrough/) from the original pass, PLUS a second hands-on pass after the Wave-1 merge specifically re-exercising the rename flow (found the Save/Cancel button bug below), PLUS a third pass (2 more screenshots, same directory) against a temporary second sandbox (:8073/:3073, spun up via `scripts/sandbox.sh` with different ports so it wouldn't contend with the live e2e run on :8072/:3072) specifically re-verifying PR #534's PropertySearch.jsx changes: searching "gap" ranks `bargap`/`bargroupgap` sensibly; searching "title" groups the exact match first with a compact "N selected" badge, replacing the old confusing "0 of 1366 properties selected" counter entirely — no dark-on-dark/light-on-light, no clipping, 0 console errors. One minor Add-Insight-picker viewport-clipping observation left as polish, not blocking.

## Additional bugs found while chasing down e2e failures and the coverage sweep (not part of the original UX audit)
- **`ExplorationCard.jsx` whole-card not clickable** — only the small "Open" button opened the exploration; the whole gallery card is now clickable (`role="button"`, `onClick`/`onKeyDown`), matching every other card-like surface in the app.
- **`InlineRenameInput.jsx` Enter-key propagation** — `onKeyDown` never called `stopPropagation()`, so pressing Enter to commit a rename also bubbled up as "Enter pressed" to ExplorationCard's new whole-card `onKeyDown` above, firing `handleCardOpen()` mid-rename. Fixed at the shared component so every caller benefits, not just the gallery card.
- **`InlineRenameInput.jsx` Save/Cancel button click propagation (found via a SECOND, post-merge manual walkthrough of the same rename flow)** — the Enter-key fix above didn't cover the mouse path: the commit (✓) and cancel (✕) buttons' `onClick` never stopped propagation either, so clicking Save renamed correctly but the click ALSO bubbled to ExplorationCard's whole-card handler and navigated into the exploration — the existing e2e coverage only ever committed via `input.press('Enter')`, never by clicking the checkmark, so it never caught this sibling gap. Fixed the same way.
- **`e2e/helpers/workspace.mjs` ignored `PLAYWRIGHT_BASE_URL`** — its `BASE` constant only read the (undocumented-outside-a-few-docstrings) `VIS_CANVAS_BASE` var. All 16 specs importing `openWorkspace()` from here silently navigated to the shared `:3001` sandbox regardless of an agent's own isolated `PLAYWRIGHT_BASE_URL`, while their backend assertions (via `helpers/sandbox.mjs`) correctly hit the isolated backend — the exact mismatched-pairing bug `origin/jared/e20-e2e-apibase-isolation` fixed in `sandbox.mjs`, just never migrated to this second, older helper. Root-caused via Playwright trace inspection (the recorded `page.goto` call showed `:3001` despite `PLAYWRIGHT_BASE_URL=:3072`) after two manual repros against the live sandbox succeeded while the real spec kept failing deterministically. Fixed here; Jared's own merge of Wave 1 independently picked up the fix (`2963ebf9`) and extended `check-e2e-sandbox-isolation.sh` to catch the whole class going forward.
- **`explore-this-dashboard-chart.spec.mjs` test fixture bug** — the "container item never offers Explore this" test right-clicked a hardcoded `{x:4,y:4}` on `row.0`, assuming a fixed horizontal CSS-grid layout. `Dashboard.jsx`'s `shouldStack(width)` flips that same row to a stacked column layout once the canvas narrows (true inside the Workspace shell with both rails open), so the fixed offset landed on the item itself, not the row's own chrome. Rewrote to compute the real gap between `item.0`/`item.1` from their live boxes and pick whichever axis (row or column) actually has one.
- **`explorer-home.spec.mjs` "duplicate" test isolation gap** — unlike every sibling test in the file, it never seeded its own exploration; it only ever passed by accident when an earlier spec in a full suite run happened to leak state into the shared gallery. Running it in isolation hung forever on the first card locator. Now mints its own via "+ New exploration", matching the rest of the file.
- **`exploration-promote-fallback-dashboard-offer.spec.mjs` stale "gate correction"** — a merge-conflict resolution during the Wave-1 merge briefly left a test comment claiming the Save-to-Project checklist "can never be empty" for a fresh source-tile exploration. Re-verified live against the fully merged tree (three manual scenarios: blank exploration, source-tile exploration, and a model with real SQL typed) — the checklist correctly reads "No changes to save." with Promote disabled in the untouched cases, and offers exactly the one authored row when real content exists. Restored the precise assertion instead of leaving the weaker one in place.

- **`ExplorationPane.jsx` stale-snapshot clobber on reload during a cross-tab race (#19, found on a SECOND pass, after the Wave-1 + PR #534 merges, via a fresh full e2e run)** — the reload-persistence flush fix above (`flushNow`, wired to `visibilitychange`/`pagehide`/`beforeunload`) unconditionally re-POSTed this pane's *current* client snapshot on every hide/reload, even when nothing was actually pending because the debounce had already fired and synced. Harmless when this tab's snapshot is the freshest thing around — not on the losing side of a cross-tab last-write-wins race: reloading the losing tab silently re-pushed its now-stale snapshot and clobbered the winner's already-persisted draft, at the exact moment the page reloaded to go read it back. Root-caused via `exploration-cross-tab-concurrency.spec.mjs:112`, which failed deterministically (SQL reverted from the winning context's marker back to the losing context's) in both a full run and an isolated re-run — the other 3 failures in that same run (`exploration-cross-tab-dnd-isolation.spec.mjs`, `exploration-preview.spec.mjs:530`, `exploration-promote.spec.mjs:292`) did NOT reproduce in isolation and were confirmed contention flakes under that run's elevated load (avg 10-29). Fix: track "is a debounced write genuinely still outstanding" via a new `liveSyncPendingRef` — `liveSyncTimerRef` alone isn't a reliable signal (deliberately never nulled when a timer fires naturally, only when a new edit's debounce supersedes it). Falsified before trusting it: with the guard temporarily removed, the new regression test failed as expected; restored, it and all 34 sibling `ExplorationPane` tests pass, and the e2e test itself now passes in isolation against the live sandbox.
- **VIS-1107 (Urgent): `flushExplorationSync` gated on the outer live-sync ref, silently dropping the SEPARATE ~1s backend-persist debounce (found via a THIRD pass, on a fresh full e2e run against the merged tree after 90a7ce52)** — the `liveSyncPendingRef` guard above (90a7ce52's own fix) only tracks `ExplorationPane`'s own 600ms client debounce; it does not track `workspaceExplorationsStore.js`'s independent ~1s backend-persist debounce (`_pendingSyncTimers`) that `updateExplorationDraft` (re)arms on every call, including from the 600ms timer's own natural-fire callback (which deliberately hands off without force-flushing — a normal edit is meant to ride out its own ~1s window). Sequence that broke: edit → 600ms timer fires naturally → hands off to the ~1s backend debounce → `liveSyncPendingRef` flips back to false → reload lands before the inner ~1s timer completes → the guard read "outer layer idle" as "nothing to flush" and skipped `flushExplorationSync` entirely, silently dropping the still-queued write. This is exactly why `exploration-reload-persistence.spec.mjs:97` (a single SQL edit, reloads before the outer timer ever fires) passed while `:129` (two edits ~800ms apart — drag-and-drop overhead alone exceeds the 600ms window — so the outer timer fires naturally before the final drop) failed 3/3 in isolation on a quiet sandbox. Root-caused with an instrumented live repro against a real sandbox (temporary console logging + Playwright network-request tracing), not guesswork: the old code produced zero network requests around the reload in the x/y case vs. a completed POST in the SQL-only case — confirmed from the wire. Fix: `flushExplorationSync` is already a safe no-op in the real store when nothing is queued (checks `_pendingSyncTimers` internally), so it's now called unconditionally on every flush; only the "capture a fresh snapshot + call `updateExplorationDraft`" step stays gated on `liveSyncPendingRef`, since that's the part that risked a stale-snapshot clobber. Filed and resolved as Linear VIS-1107 (Urgent) — full diagnosis, the discrepancy answer, and verification are on that issue. Re-verified: `exploration-reload-persistence.spec.mjs` 3/3 clean (both `:97` and `:129`), `exploration-cross-tab-concurrency.spec.mjs` still 2/2 (no regression on the clobber fix), new unit test falsified against the old guard, 341/341 suites / 6377/6379 tests passing, lint clean.

All nine confirmed via isolated re-runs against a live sandbox after each fix — not just read from code or assumed fixed from a diff.

## Known session incidents (transparency)
- Mid-session I accidentally stopped my own sandbox (:8072/:3072) while a background Playwright run was still using it — `sandbox.sh stop` mid-flight, cascading ~32 of that run's 44 failures into `ERR_CONNECTION_REFUSED`/`ECONNREFUSED`. Separately, a `sandbox.sh start` invocation that exceeded the harness's foreground-command timeout got auto-migrated to a tracked background task; when that task was later torn down, its child server processes (started via a bare `&`, no `nohup`/`disown`) died with it, silently killing a sandbox I'd believed was still up.
- Later, while a full e2e run was going in the background, I did a manual Playwright-MCP walkthrough against the SAME live sandbox to re-verify the rename flow (finding the Save/Cancel button bug above) — the two browser sessions mutating the same backend `.visivo/explorations/` directory concurrently produced one contention-driven failure (`exploration-build-rail.spec.mjs:312`, a 30s draft-sync timeout). Stopped the background run, re-verified the isolated test cleanly on its own, and did not touch the sandbox manually again while any automated run was in flight.
- A third, later incident: a coverage run surfaced a `WorkspaceDndContext.test.jsx` timeout on a file never touched in this diff. `ps aux` showed the actual cause — a concurrent, unrelated Playwright script the user was running by hand against their own `:3001` sandbox (a completely different port from mine), driving real headless Chrome instances and pushing system load averages to 34–50. Re-ran the failing file in isolation twice (124/124 both times) before calling it a flake, per the same discipline as the other two incidents.
- Both classes of incident were diagnosed, not asserted away: every failure they produced was re-run against a verified-alive, uncontended sandbox before being called a pass, and every REAL (non-infra) failure that surfaced along the way is documented above with an actual fix.

**Cross-spec cleanup determinism (verified, not assumed):** a reviewer independently running `exploration-reload-persistence.spec.mjs` + `exploration-cross-tab-concurrency.spec.mjs` back-to-back on a sandbox already contended by a concurrent gate and a day of manual probes saw both specs fail together on repeat invocations — a plausible signal of test-authored state leaking between specs via the shared `.visivo/explorations/` repository. Diffed the exploration record count before/after each of 5 back-to-back iterations of both specs together on a clean, quiet sandbox: **0 leaked records after every single iteration, 15/15 individual test passes.** Also confirmed both specs' `afterEach` cleanup pattern (`request.delete(...).catch(() => {})`) is the SAME convention already used identically across ~25 other spec files in this suite — not something this track introduced. Conclusion: cleanup is deterministic; the observed cross-contamination was sandbox contention/staleness, not a leak in these specs.

## Checklist gating re-verification (Jared's note: "still visibly outstanding")
Jared's composed walkthrough flagged a fresh exploration's chip/insight showing the generic names "model"/"insight" and asked me to confirm the PROMOTE CHECKLIST gating (as opposed to naming, which is T3's) still only offers objects the user actually authored. Re-verified live against the fully Wave-1-merged tree in three scenarios (not read from code): a blank "+ New exploration", a source-tile-seeded exploration, and a model with real SQL typed in. In every case the Save-to-Project checklist correctly excludes untouched scaffold objects and, once real content exists, offers exactly that one authored row — `promoteChecklist.js`'s VIS-1102 gating holds. This also surfaced and fixed a stale test comment (see above) that had drifted from actual behavior during the merge.

## Filed for follow-up, found via the coverage sweep, not fixed here (VIS-1106)
Chasing `promoteChecklist.js`'s branch coverage to 97.89% surfaced a real status-resolution bug in the chart tier: `candidate.isNew` is built as `undefined` for every chart candidate (`buildPromoteChecklist`, the two places a chart candidate object is constructed), so `candidate.isNew === false ? 'modified' : 'new'` can never take the `'modified'` arm — a chart the user is actually *updating* silently gets labelled **"new"** in the Save-to-Project checklist whenever the diff endpoint returns a falsy-but-not-null `chart` value. That's the exact update-vs-create signal 01-ux-spec.md §3/D11's checklist exists to make legible, and it's currently always wrong for charts. Left unfixed deliberately: it's behavioral (status resolution, not gating), wants its own targeted test rather than riding in under a coverage commit where it'd be easy to miss in review, and whether it's reachable today depends on a backend contract (`fetchExplorerDiff`'s diff endpoint shape) owned by another module — worth confirming with that owner before deciding the fix (start tracking real chart `isNew` state vs. leaving the diff endpoint to always return null-or-string for `chart`).

## Coverage (whole-suite run, coverage measured only for files in this diff)

Command: `npx jest --config=/tmp/jest.cov.t5.json --coverage --coverageReporters=text` (config's `collectCoverageFrom` is exactly this PR's 15 touched source files; the WHOLE suite runs so tests living elsewhere still count).

**promoteChecklist.js explicitly, since it's the one that decides what gets written to the user's project on save: 100% stmts / 97.89% branch / 100% funcs / 100% lines.** Both remaining branch lines are named below with the specific mechanism, not asserted as "fine."

| File | Stmts | Branch | Funcs | Lines | Per-branch reason for anything left uncovered |
|---|---|---|---|---|---|
| **All files (aggregate, final run — after fac3940c)** | **98.88%** | **93.51%** | **99.32%** | **99.9%** | Current as of the final commit; per-file numbers below match this same run. |
| explorationLifecycle.js | 100% | 98.66% | 100% | 100% | Line 91, `insightStates[name] \|\| {}`: `name` always comes from `Object.keys(insightStates)` in the one caller, so the fallback can never see a missing key — unreachable by construction. |
| ExplorationCard.jsx | 100% | 100% | 100% | 100% | — |
| InlineRenameInput.jsx | 100% | 100% | 100% | 100% | — |
| workspaceExplorationsStore.js | 99.64% | 100% | 98.59% | 100% | — |
| workspaceStore.js | 100% | 100% | 100% | 100% | — |
| **promoteChecklist.js** | **100%** | **97.89%** | **100%** | **100%** | Line 51, `stripParentModel`'s `config \|\| {}`: every real candidate (model/field/insight/chart) always constructs a literal `config` object in `buildPromoteChecklist` — `validateRow` is never called with anything falsy via the public API. Line 226, the chart tier's `candidate.isNew === false ? 'modified' : 'new'` consequent: chart candidates are built with `isNew: undefined` at lines 115 and 156, so `candidate.isNew === false` is never true and the branch always yields `'new'` — **unreachable given the current diff contract, not unreachable outright**: whether it fires depends on whether `fetchExplorerDiff`'s backend endpoint can ever return a falsy-but-not-null `chart` value, a contract owned by another module. **Filed as VIS-1106** — a real bug (a chart the user is updating can get mislabeled "new" in the save checklist, the exact update-vs-create signal D11 exists to make legible), left unfixed here deliberately: it's behavioral, wants its own test, and smuggling a status-resolution change into a coverage commit would hide it from review. |
| ExplorationBuildRail.jsx | 100% | 95% | 100% | 100% | Line 22, `s.insights \|\| []`: forcing `insights: undefined` in a test to reach this hit an unrelated infinite-render loop elsewhere in the tree — proof this state is never naturally undefined in a mounted component; reverted that test rather than paper over a crash. |
| SemanticLayerErdModelNode.jsx | 98.55% | 97.14% | 100% | 100% | Line 48 only, the field-pill in-flight guard (`exploringRef.current.has(name)`): one specific double-click interleaving still untested after closing every other branch (unavailable-action guards on both the field pill and model button, the seed-override fallback, a failed create, selected-card chrome, empty-columns state). |
| ExplorationPane.jsx | 99.31% | 94.87% | 100% | 100% | Two lines. (1) `flushNow`'s `if (liveSyncTimerRef.current) {...}` false side: only reachable if `liveSyncPendingRef` is true while `liveSyncTimerRef` is already null — the two refs are only ever set together (both `true`/non-null in the same effect run) and only ever cleared together (this exact guarded block, or the natural-fire callback) — structurally unreachable given the current wiring, kept as a defensive guard against a future edit decoupling them. (2) `handleRecheckStaleness`'s `if (!current) return`: only reachable once the pane has already switched to its not-found render, which removes the very button a test would use to trigger the recheck — the two states are mutually exclusive in the UI. |
| LibraryRow.jsx | 98.76% | 98.23% | 95.83% | 100% | Line 163, `ContextMenu`'s `canAddToExploration = false` default: the component always passes this explicitly from its own props, so the default is never exercised. Line 213, a 3-way `&&` chain's middle evaluation slot: both outer conditions (`canExploreThis`, `canAddThisToExploration`) are independently tested true/false, but not the specific combination where the first is false and the second's own evaluation is what's being measured — low-value residual, same underlying condition already proven both ways individually. |
| LibrarySourceRow.jsx | 98.8% | 83% | 100% | 100% | `RowShell`'s drag/selected/hovered CSS-class ternaries (`isDragging ? 'opacity-40' : ''`, `draggable ? 'cursor-grab...' : ''`, etc.) across the table-row and column-row variants — each is visual-only and structurally identical to ones already closed elsewhere in the file; genuinely low value per-branch, not chased further this pass. The one BEHAVIORAL gap here (`!available`, schema browsing disabled in dist/cloud) is closed — see PR body for the mock restructuring that made it testable. |
| SemanticLayerCanvas.jsx | 100% | 89.23% | 100% | 100% | Remaining gaps are entirely in the pre-existing column-to-column connect/relation-authoring surface this diff doesn't touch (`onConnectStart`/`onConnectEnd`'s coordinate-fallback math, `groupFieldsByModel`'s collection fallbacks) — closed 8 "unavailable store action" tests plus 3 more found on a second pass (context-menu Open/Open-in-new-tab have their OWN copies of the setWorkspaceSelection/openWorkspaceTab guards, distinct from onNodeClick's copy — my first pass only closed the latter). |
| CanvasContextMenu.jsx | 92.73% | 89.02% | 100% | 98.58% | Verified-dead cluster (read every call site, not guessed): `isContainerItem`/`explorerSubjectAtKey`/`rowPathForItemKey`/`containerAncestorPath`'s own "not an item" early returns can never fire — every call site guards with `menu.kind === 'item' &&` first, and `menu.kind` is derived from the identical parsing rule these functions check internally, so the precondition always holds. Also dead: `MenuItem`'s unused `danger` prop (no current menu item passes it) and `setSelectedKey`'s truthiness check (a `useCallback` result, never conditionally assigned — always a function). One line (`root.contains(e.target)`'s false side, an event-bubbling edge case) is real but not cheaply falsifiable in jsdom. Lines 227-228 (the `nextConfig === dashboardConfig` no-op guard) — same shape as promoteChecklist's line 226: no reshape helper returns the identical object reference via any real menu action. |
| explorerStore.js | 98.93% | 88.41% | 98.67% | 100% | The single largest legacy file touched (only ~50 lines are actually mine). ~55 new tests this pass across `getAllKnownNames`/`assertNameUnique`/`NameCollisionError`, `renameModelTab`'s full ref-propagation matrix, every active-model selector, `selectHasModifications`/`selectDerivedInputNames`, `loadModel`/`loadChart`, `fetchExplorerDiff`, snapshot/restore, `recordPromotedInsightSignature`, `buildExplorationSeedState`'s not-found/skip paths, `validateExplorerExpression`'s response-shape fallbacks, `addObjectToActiveExploration`/`handleTableSelect`. Remaining ~55 branch-sides are further `\|\| []`/`\|\| {}` defensive fallbacks scattered across ~15 different small legacy selector/builder functions in `buildModelStateFromObject`/`transformInsightToUiState`/`buildExplorationSeedState`'s chart-seeding branch — each individually cheap but the SET is large; I ran out of time in this pass to close all of them and am naming that honestly rather than padding with a rushed batch. |

**Progress this pass (Jared's independent measurement matched mine): started at 97.35% stmts / 86.30% branch aggregate, 9 of 15 files under the 95% branch bar. Currently 9 files at or above 95% branch (promoteChecklist, ExplorationCard, InlineRenameInput, workspaceExplorationsStore, workspaceStore, ExplorationBuildRail, SemanticLayerErdModelNode, ExplorationPane, LibraryRow); 6 files still under, each with the specific per-branch reason above — a mix of "ran out of time, genuinely closable" (explorerStore.js, LibrarySourceRow.jsx's visual ternaries) and "verified structurally unreachable by reading every call site" (CanvasContextMenu.jsx's biggest cluster, promoteChecklist.js's two lines).**

**Honesty note:** every test added asserts real, specific behavior (no `expect(true).toBe(true)` filler, no render-smoke tests). Numbers re-measured fresh after every fix; none carried over unverified from an earlier run.

## Deferred finding, root-caused: "Query results and preview don't persist / restore"

**Broader than the audit described.** The audit reproduced this via reload; the actual
trigger is any RE-ACTIVATION of an exploration tab — switching away and back within the
SAME browser session loses results too, no reload required.

## Mechanism

1. Query results live ONLY in the legacy singleton Zustand slice, per-model:
   `explorerModelStates[name].queryResult` / `.queryError` / `.enrichedResult`.
   Written by `setModelQueryResult` (`viewer/src/stores/explorerStore.js:714-733`) —
   a pure in-memory write, never routed through `updateExplorationDraft`.

2. `snapshotExplorerWorkingState()` (`explorerStore.js:1758-1784`) is what
   `ExplorationPane` persists to the backend (via `legacyStateToDraft` /
   `updateExplorationDraft`). It builds each model's persisted shape explicitly:
   `{ sql, sourceName, sourceEdited, computedColumns, isNew }` — no `queryResult` field
   at all. The file's own docstring (`explorerStore.js:1745-1752`) says this is
   deliberate: "Deliberately EXCLUDED from both snapshot and restore... large,
   ephemeral query results... re-running is cheap."

3. `restoreExplorerWorkingState(snapshot)` (`explorerStore.js:1792-1818`) is a HARD
   RESET of `explorerModelStates`, built via `createEmptyModelState(...)` (defaults
   `queryResult: null`, line ~187) merged with the restored (results-less) fields.
   There is no code path that ever repopulates `queryResult` on restore.

4. `ExplorationPane.jsx`'s activate `useLayoutEffect` (lines 140-173, deps `[id, !!record]`)
   calls `restoreExplorerWorkingState` on EVERY activation of the pane for a given id —
   not just first mount.

5. `MiddlePane.jsx:148` renders `<ExplorationPane id={activeObject.name} />` with NO
   `key` prop — switching between exploration A and B (both type `exploration`) reuses
   the SAME component instance with a changed `id` prop, which is exactly what re-fires
   the effect in (4). So: open A, run a query, switch to B, switch back to A — A's
   `useLayoutEffect` cleanup already flushed+cleared, then the activate branch calls
   `restoreExplorerWorkingState` again for A, from the SAME results-less persisted
   draft — results are gone, with zero user action beyond tab-switching.

## Why this wasn't "just re-run it, it's cheap"

The direction note in the code is defensible for a SINGLE exploration held open the
whole session — but it does not account for park/resume (multi-tab use is the whole
point of the Workspace shell) or reload. Both round-trip through the exact same
snapshot/restore pair, so both lose results identically.

## Recommended follow-up fix (not implemented in this PR — scope/time)

Add a SESSION-LOCAL (never persisted to the backend) results cache keyed by
`(explorationId, modelName)`, mirroring the module-level Map pattern this codebase
already uses for exactly this kind of "ephemeral, per-record, non-serializable
bookkeeping, not reactive state" (see `workspaceExplorationsStore.js`'s
`_pendingSyncTimers` / `_openDraftSnapshots`). On restore, before defaulting
`queryResult` to null, check this cache first; only show "Run a query to see results"
when the model has genuinely never been run this browser session. This preserves the
"never persist large result sets to the backend JSON document" constraint the original
design protects, while fixing the actual complaint (results vanish on ordinary
park/resume, not just hard reload). A stretch alternative — auto-rerun on restore
when SQL is present and no cached rows exist — was considered and rejected for the
follow-up to design further: it fires a real network query on every tab switch, which
is a different cost/UX tradeoff than a client-side cache and deserves its own review.


jpjsonic:

PR #538, opened 2026-07-23
fix(explorer): chart silently dropped from Save-to-project when unnamed (VIS-1109)

jpjsonic:
## Summary

- **Bug (VIS-1109):** in the most common Explore 2.0 flow — new exploration → type a query → bind x/y → "Save to project" — the chart never gets a name, so `buildPromoteChecklist` gates it out of the checklist entirely (it requires `explorerChartName && chartHasContent`). The modal shows "Save 2 to project" (model + insight) with **no chart row and no notice** that the chart, which is rendering a real live scatter, was dropped.
- **Root cause:** `ExplorationBuildRail.jsx`'s live auto-naming effect (T3/D11) already carries a fresh build through the naming cascade `<source>_query` → `<...>_insight` (`promoteNaming.js`'s documented chain), but never took the last step to `<...>_insight_chart`. The chart stays `explorerChartName === null` ("Chart: Untitled") forever, even once it has real bound content.
- **Fix (preferred path, as scoped):** extend the same live auto-naming effect to the chart. It anchors on the first *meaningful* (non-scaffold) insight the chart references, falling back to the model anchor for a chart made valid by layout config alone — mirroring `promoteNaming.js`'s own save-time suggestion logic. It only fires once the chart is genuinely content-bearing, using the exact same "has content" gate `buildPromoteChecklist` itself uses (extracted into a shared `computeChartHasContent`/`computeMeaningfulInsightNames`/`isMeaningfulInsightState` trio in `explorerStore.js` so the effect and the checklist can never drift apart on what counts as "content"). A chart referencing only scaffold insights, with no layout of its own, is left unnamed and editable — same as an unbound model/insight already is.
- Did **not** need the fallback (explicit "chart isn't named and won't be saved" notice) — auto-naming doesn't conflict with any documented intent; `promoteNaming.js`'s own docstring already treats `_query → _insight → _chart` as the intended cascade, this effect just never completed it.

## VIS-1106 area (chart status labeling)

Read `promoteChecklist.js`'s chart status resolution (`candidate.isNew: undefined`, resolved via `diff.chart`) before touching anything nearby — left untouched; this PR only changes whether the chart becomes a *candidate* at all (via `explorerChartName` being set), not how its `new`/`modified` status label is resolved once it is one.

## Test plan

- [x] **Failing-then-passing unit tests** (`ExplorationBuildRail.test.jsx`, new `chart naming (VIS-1109 …)` describe block): asserts the chart auto-names itself to `<insightAnchor>_chart` (or `<modelAnchor>_chart` when only layout content exists) the moment it has real content; stays unnamed while scaffolding-only; never renames an already-named chart; disambiguates on collision; is idempotent. Confirmed these **fail** against the pre-fix effect (chart name stays `null` forever) and **pass** post-fix.
- [x] **Integration unit test** (new top-level describe in the same file, `VIS-1109 — a bound-but-unnamed chart is no longer silently dropped from Save to project`): renders the real `ExplorationBuildRail`, lets its real naming effect run, then calls the real `buildPromoteChecklist` (only the three low-level HTTP/AJV validation gates are mocked, same boundary `promoteChecklist.test.js` itself uses) and asserts a chart-tier row now appears — proving the fix end-to-end at the unit level, not just that an internal field changed.
- [x] **VIS-1102 regression pin** (same file): an untouched exploration (scaffold model/insight/chart, nothing bound) still produces an **empty** checklist — confirmed auto-naming never starts offering scaffolding.
- [x] **`promoteChecklist.test.js`**: all 34 existing tests still pass unchanged after refactoring `chartHasContent`/the insight scaffold-guard to use the new shared helpers (behavior-preserving refactor, verified before touching the UI effect).
- [x] **E2e** (`exploration-chart-build-updates.spec.mjs`, extended — already registered in both the `parallel` testIgnore and `exploration-mutations` testMatch lists in `playwright.config.mjs`, grep-verified twice, no config change needed): new test types a query, binds x/y via real drag-and-drop, never renames the chart by hand, then opens "Save to project" and asserts the chart row exists and is checked, actually saves, and confirms the chart object round-trips to the real backend project via `GET /api/charts/<name>/`. Verified this **fails** pre-fix (`explorerChartName` poll times out) and **passes** post-fix, run against an isolated sandbox (`:8092`/`:3092`).
- [x] Ran `exploration-promote.spec.mjs` (4 tests) and `exploration-build-rail.spec.mjs` against the same sandbox — all pass, no regression.
- [x] Verified 8 unrelated e2e failures seen in a broader sweep (`exploration-preview.spec.mjs` ×7, `save-as-metric.spec.mjs` ×1) reproduce **identically** with this fix reverted (stashed) — confirmed pre-existing, not caused by this change.
- [x] `bash scripts/viewer-check.sh` (full unit suite + lint) clean after the final commit: 342 suites / 6546 tests passing, 0 lint errors.


jpjsonic:

PR #539, opened 2026-07-23
fix(explorer): ephemeral-lifecycle data loss + spurious dirty dialog (VIS-1108/1110)

jpjsonic:
## Summary

Two confirmed bugs at the T3 (live model auto-rename) × T5 (exploration
lifecycle) seam, fixed coherently in one series since they share the same
root mechanism.

### VIS-1108 (URGENT, DATA LOSS)
Typing SQL into a source-tile browse then closing the tab within
`ExplorationPane`'s ~600ms live-sync debounce silently deleted **both** the
backend record and the typed edit. `closeWorkspaceTab`'s GC
(`workspaceStore.js`) read the PERSISTED `workspaceExplorations.byId[id].draft`
— which only reflects what the debounce has already pushed via
`updateExplorationDraft` — and, seeing a still-blank draft, concluded the
exploration was untouched. The GC runs synchronously inside the tab-strip's
click handler, well before React gets to `ExplorationPane`'s own deferred
unmount-cleanup flush, so the stale draft always "won" and the record (and
the only copy of the typed SQL) was deleted.

**Fix:** the GC now additionally takes a synchronous snapshot of the live
legacy `explorerStore` working-state — only when the closing tab was the
active one (the only case where that state can be hot for this exploration)
— and never deletes unless BOTH the persisted draft and the live snapshot
agree there's nothing meaningful yet. Purely additive: it only ever makes the
GC *more* conservative, so the existing "GC a genuinely untouched browse"
contract is unchanged.

### VIS-1110 (HIGH)
A source-tile browse seeds a generic model name (`query_1`) that T3's live
auto-rename effect (`ExplorationBuildRail.jsx`, D11) renames to
`<source>_query` the instant a source anchor is available. That mutation
touches the exact same `explorerModelTabs`/`explorerModelStates` fields
`ExplorationPane`'s live-sync effect watches, so it armed the same debounce a
real edit would — eventually flipping `syncStatus` to `'saving'` (the tab's
dirty dot) and firing a redundant backend POST purely from opening and
looking at a browse tile. Closing during that window popped a spurious
"unsaved changes?" confirm dialog for a tab the user never touched.

**Fix:** the live-sync effect gates on the same `isExplorationVisibleInGallery`
predicate the Home gallery and the GC already use as the single source of
truth for "is this still just a browse gesture", evaluated against a fresh
live snapshot. A record that's already visible (no seed, already renamed,
already promoted, or already has real content) is untouched by the gate —
every ordinary edit on an ordinary exploration arms the debounce exactly as
before — and a genuine first real edit on a browse seed makes the gate pass
the moment it lands, so real content is never dropped, only ever correctly
recognized.

## Falsifiability (failing-then-passing tests)

- `viewer/src/stores/workspaceStore.test.js` — new `describe('VIS-1108 — the
  GC must consult the LIVE legacy working-state, not a store draft known to
  lag')`: the "fast close-after-type" test fails on pre-fix code
  (`deleteExploration` called once when it should never be) and passes after.
  Two companion tests guard against regressing the existing GC contract and
  against misattributing a background tab's close to whatever exploration
  happens to be hot in the legacy store.
- `viewer/src/components/views/workspace/ExplorationPane.test.jsx` — new
  `describe('VIS-1110 a pure rename-of-a-generic-seed must not dirty an
  untouched browse')`: the auto-rename test fails on pre-fix code
  (`updateExplorationDraft` fires when it shouldn't) and passes after. Two
  companion tests prove the gate never blocks a real edit or an already-visible
  exploration.
- `viewer/e2e/stories/exploration-lifecycle.spec.mjs` — two new stories against
  a real sandbox, both verified to fail on pre-fix code (one on the missing
  SQL in the persisted record, the other on `syncStatus` genuinely flipping to
  `'saving'` from the rename alone) and pass with the fix. Already
  double-registered (this file was registered in both `parallel`'s testIgnore
  and `exploration-mutations`'s testMatch before this PR — `grep -c
  "exploration-lifecycle.spec.mjs" playwright.config.mjs` = 2), unchanged by
  this PR since no new spec file was added.

## Regression verification (sandbox :8093/:3093)

All green post-fix:
- `exploration-lifecycle.spec.mjs` — 11/11 (9 pre-existing + 2 new)
- `exploration-reload-persistence.spec.mjs` — 2/2 (VIS-1107 gate)
- `exploration-cross-tab-concurrency.spec.mjs` — 1/1
- `explorer-home.spec.mjs` — 8/8 (including the pre-existing "GC an untouched
  browse" contract test)
- `exploration-build-rail.spec.mjs` — 6/6
- `exploration-concurrent-rename-and-draft-sync.spec.mjs` — 2/2
- `workspace-back-forward-exploration.spec.mjs` — 2/2
- `workspace-tab-close-dialog-navigation.spec.mjs` — 3/3
- `workspace-tab-phantom-exploration.spec.mjs` — 2/2

Plus the full unit suite (`bash scripts/viewer-check.sh`): 342 test suites,
6544 passed / 2 skipped, lint clean.

## Test plan
- [x] Unit tests fail before the fix, pass after (both bugs)
- [x] e2e tests fail before the fix, pass after (both bugs)
- [x] Full `yarn test` (342 suites) + `yarn lint` clean
- [x] `exploration-reload-persistence.spec.mjs` + `exploration-cross-tab-concurrency.spec.mjs` still green (VIS-1107 non-regression)
- [x] Existing "GC an untouched browse" contract still green (no regression)
- [x] `check-e2e-sandbox-isolation.sh` clean (no hardcoded ports in new specs)


jpjsonic:

PR #540, opened 2026-07-23
test(e2e): force the VIS-1091 same-name collision explicitly (T3 unique-naming)

jpjsonic:
## Summary

- `exploration-preview.spec.mjs:382` (VIS-1091) was failing because its PRECONDITION assumed two independently-created explorations' first insights would auto-collide on the same default name (`insightNameB === insightNameA`). T3's live auto-rename (Wave 2) now scopes `generateUniqueName` globally (`getAllKnownNames` unions every cached collection, project-wide) — once exploration A promotes, exploration B's own auto-namer sees A's real name in `state.insights` and correctly uniquifies away from it. Valid naming improvement, invalid test precondition — not a regression to work around by weakening the guarantee.
- Restructured the setup to force the same-name collision explicitly instead of relying on a natural collision, then re-run the existing (unchanged) isolation predicate.
- **How the collision is forced**: I first tried the real rename affordance the task pointed at (`InsightBuildSection`'s inline-rename UI, and the `renameInsight` store action it calls — same code either way). Confirmed by live reproduction against the sandbox that **both refuse**: `assertNameUnique` treats a real, already-promoted insight name in `state.insights` as taken regardless of which exploration promoted it (correct, intentional, separately unit-tested behavior — not specific to this bug). The UI shows an inline "already in use by a insight" error and leaves the name unchanged; the store action throws `NameCollisionError`. So the rename affordance is genuinely unreachable for this exact scenario. I did **not** stop there, because a different, real path still reaches the required state without touching that guard: `exploration-promote.spec.mjs`'s `seedDraftWithExistingNames` hit the identical wall for its own same-name scenario and resolved it by seeding the store directly with the target name in place — "what the (documented, still out-of-scope) 'Explore this' flow would eventually produce." I mirrored that precedent with a new `forceInsightNameCollision` helper that remaps only the insight's key, leaving its bound props/content exactly as exploration B independently created them — so the forced collision is a NAME match over genuinely independent content, precisely the scenario VIS-1091 guards against.
- The isolation predicate itself (the actual guarantee: an exploration must never treat its own draft as promoted just because a DIFFERENT exploration published a real insight under the same name) is untouched.

## Falsifiability proof

Temporarily dropped the exploration-scoping in the test's predicate (`isRealInsight && (true || isPromotedByThisExploration)`) to simulate the original bug — the test failed red (`Expected: false, Received: true`, timeout waiting on the predicate). Restored the predicate and confirmed green again. Also temporarily broke the equivalent real guard in `ExplorerChartPreview.jsx`'s `promotedNames` filter and reverted it — worth noting this test's predicate is a store-level replica (reads `state.insights` + `workspaceExplorations.byId[...].promoted` directly) rather than exercising that component's memo, a pre-existing property of the test unrelated to this fix, unchanged here.

## Test plan

- [x] `VIS-1091` spec run individually against sandbox :3091 — 3× consecutively, all green
- [x] Full `exploration-preview.spec.mjs` file run against sandbox :3091 — all 9 tests pass
- [x] `yarn lint` clean (no src files touched in the final diff — a temporary source edit used only for the falsifiability proof was reverted; `git diff` confirms zero change to `ExplorerChartPreview.jsx`)
- [x] Falsifiability: predicate reproduces the original bug and fails red when the exploration-scoping is dropped


jpjsonic:

PR #542, opened 2026-07-24
test(e2e): cover dimension creation in the chart-build loop

jpjsonic:
## Summary

- `exploration-chart-build-updates.spec.mjs` covered the chart-build loop's drag → aggregate → computed column → layout prop → **metric** creation → type-switch path, but never dimension creation — Jared explicitly flagged this as the one uncovered piece.
- **Real affordance, not an invented one**: `PillMenu.jsx` has no "Save as dimension…" durable-promote action at all (only "Custom aggregation…", "Save as metric…", "Remove" live below the "Use as" preset list — confirmed by reading the full component, no `onSaveAsDimension` prop exists anywhere). Dimensions aren't named/durable-promoted the way metrics are in this surface. The real, only user-facing "make this a dimension" action is the **"Use as: Dimension" preset** (`pill-menu-preset-dimension`) — the exact same preset `save-as-metric.spec.mjs`'s own last test ("'Save as metric…' stays disabled for a dimension (non-aggregate) pill") already uses to produce its disabled-pill fixture.
- Extends the existing chart-build test: pushes the x pill (bound to the computed `less_than_4` column, a string CASE expression) into an aggregate (`COUNT`, since SUM/AVG/MEDIAN are hidden for non-numeric columns) and then explicitly back to Dimension via the preset — mirroring how the file already exercises SUM → "Save as metric" for y. Asserts a live chart at every step, the pill reflecting the bare dimension with no raw `?{` syntax (D8), and that the dimension conversion (not merely the column name) survives the later scatter → bar chart-type switch.

## Verification

- Falsifiability proven: temporarily skipped the Dimension-preset click and reproduced a failure at exactly the new guard assertion (`no longer wrapped in the COUNT aggregate`) before restoring the real flow.
- Ran the extended spec 4x consecutively against an isolated sandbox (`:8091`/`:3091`) — all green, consistent ~33-37s total per run (no contention signal).
- `yarn lint` (project's actual `eslint src` scope) clean — e2e specs aren't in scope, unaffected.
- Only file changed: `viewer/e2e/stories/exploration-chart-build-updates.spec.mjs` (already registered in both the `exploration-mutations` `testMatch` and `parallel` `testIgnore` in `playwright.config.mjs` — unaffected by this change).

## Test plan

- [x] `PLAYWRIGHT_BASE_URL=http://localhost:3091 npx playwright test exploration-chart-build-updates --reporter=list` — 3x consecutive green runs (plus an initial 4th)
- [x] Falsifiability check: guard assertion fails when the real Dimension-preset click is removed
- [x] `yarn lint` clean


jpjsonic:
Superseded by the clean re-base (single dimension-coverage commit cherry-picked onto the current feature tip). #542 was authored on the pre-squash integration branch, so it diffed as +9k/61 files against the squashed feature tip; the actual change is one spec file. Closing in favor of the clean PR.

PR #543, opened 2026-07-24
test(e2e): cover dimension creation in the chart-build loop (VIS-1111)

jpjsonic:
Closes the dimension-coverage gap Jared asked for — the one piece of the chart-build loop not yet exercised. Extends `exploration-chart-build-updates.spec.mjs` (already covers drag/drop, computed columns, layout props, **metric** creation, scatter↔bar).

This is the clean re-base of #542: that PR was authored on the pre-squash Wave-2 integration branch, so it showed a spurious 61-file/+9k diff against the squashed feature tip. Its single real commit (dimension coverage, +46/-1 to one spec) is cherry-picked here onto `feature/explorer-workspace`. #542 will be closed.

## What it covers
The build surface has no "Save as dimension" action (dimensions are the non-aggregate counterpart to metrics). The user path is the **"Use as: Dimension" preset** (`pill-menu-preset-dimension`), which `pillGrammar.js` serializes to a bare `?{${ref(model).column}}` with no aggregate wrapper. The test aggregates the x pill (via `COUNT`, since SUM/AVG are hidden for the non-numeric `less_than_4` column), asserts a live chart + `COUNT` label, then converts it back to a dimension and asserts: the chart stays live, the pill shows the bare column with **no** `COUNT` label and **no** raw `?{` (D8), and the dimension state survives the subsequent scatter→bar switch.

## Verification
- Falsifiability proven (agent): removing the `pill-menu-preset-dimension` click fails exactly at the "no longer wrapped in COUNT" guard.
- **3 consecutive green runs on the feature-based tip** (~35s each) after the clean cherry-pick.
- Already double-registered (existing spec file); `check-e2e-sandbox-isolation.sh` clean; `yarn lint` (scoped to `src`) unaffected.

https://claude.ai/code/session_01UoqyLSV4REMNep8pDEqKzi


tgsoverly (approved):
Once this is green again. :shipit:

PR #544, opened 2026-07-28
Choose when runs happen: automatic or manual, with a staged-changes surface and cancel

tgsoverly:
Two related changes to how runs are controlled: you can **cancel** one that's in flight, and you can choose whether runs fire **on save or on a button**.

---

# 1. Choosing when a run happens

Today a run is a side effect of typing. Every data-affecting save schedules one — debounced, but never asked for. Locally that's fine, since a run is a sub-second in-process rebuild. In cloud it isn't: a run claims a sandboxed pod. And in neither case can you answer *"what would a run build right now?"* — the information exists, but only ever as a boolean.

So: a per-user setting, **Automatic** or **Manual**. Local defaults to automatic, cloud to manual. In both modes the Run view lists the changes waiting to run, the Runs tab carries a dot when that list is non-empty, and a Run button builds exactly what's listed.

### The staged set is what a run would build — not what a commit would publish

These are different questions and the UI now treats them as such. A chart colour or a dashboard layout tweak is uncommitted but needs no run, so it appears in the commit panel and *not* in the staged list. Listing it beside a Run button would promise work that button doesn't do.

The scope of the run is derived from the same set that's displayed, so **the list and the run cannot disagree** — locally the staged registry produces both the list and the `dag_filter`; in cloud `run_dag_filter` is derived from `unbuilt_changes`.

### Where the mode lives

Each server answers for itself, so the viewer needs no local-vs-cloud branch — the doctrine `stores/branchingStore.js` already states:

> There is NO local-vs-cloud branching here — behavior is driven entirely by what the endpoints return.

| | stores it in | default |
|---|---|---|
| local | `~/.visivo/config.yml` | automatic |
| cloud | `User.run_trigger` | manual |

Both expose `GET`/`PUT /api/me/preferences/`.

`~/.visivo/config.yml` already held `telemetry_enabled` but **nothing had ever written it** — users hand-edited it. So the writer round-trips through ruamel in `rt` mode: comments, key order, and unrecognised keys all survive. Reads never raise; a missing, unreadable, or half-edited file falls back to defaults rather than stopping `visivo serve`. Telemetry now reads through the same helper, so the two settings can't disagree about where the file is or how a broken one behaves.

### Tracking un-run changes locally

Cloud answers "does this need a run?" from two columns per resource (`data_hash` vs `last_built_data_hash`). Local serve had no equivalent — `run_on_save` fingerprinted a resource before and after each write, noticed a difference, and fired, recording nothing. That was enough while every data edit ran immediately; a manual trigger needs somewhere to *hold* the change.

`StagedManager` mirrors those two columns with a `current` and a `built` map. Keeping both is what gives local the same revert behaviour cloud gets for free: **edit a model and change it back, and it un-stages** rather than leaving the Runs tab lit forever.

### Two local boundaries, stated so nobody "fixes" them later

- **Manual gates the editor's save path only.** The file watcher (`serve_phase.on_project_change`) and the commit-triggered rebuild stay automatic. The watcher also swaps `app.project`, the whole serve loop depends on it, and a CLI-first user editing YAML in an editor expects it to rebuild. Those runs clear the staged set so state stays consistent.
- **There is no local commit gate and there must not be one.** "Commit blocked until you run" is cloud-only. Locally, commit means "write YAML".

### Changes

| file | |
|---|---|
| `visivo/server/user_config.py` | new — read/write `~/.visivo/config.yml`, ruamel round-trip |
| `visivo/server/managers/staged_manager.py` | new — the un-run set, `current` + `built` |
| `visivo/server/views/preferences_views.py` | new — `GET`/`PUT /api/me/preferences/` |
| `visivo/server/views/run_views.py` | always stage; fire only when automatic. New `POST /api/projects/<id>/run/` |
| `visivo/server/views/commit_views.py` | `/changes/` also returns `staged`, `staged_dag_filter`, `run_trigger` |
| `visivo/server/jobs/save_run_executor.py` | `run_now()` shared by the debounce and the button; unstage on success |
| `visivo/commands/serve_phase.py` | clear staged after a watcher rebuild |
| `visivo/telemetry/config.py` | read the user config through the shared helper |
| `viewer/.../RunsView.jsx` | staged panel: the list, the Run button, the mode toggle |
| `viewer/.../RunsToolIcon.jsx` | amber dot when changes are waiting |
| `viewer/.../commitStore.js` | staged state, `runTrigger`, `setRunTrigger`, `commitAction` |
| `viewer/.../runStore.js` | `triggerRun`; refresh staged when a run succeeds |
| `viewer/.../CommitModal.jsx` | a `run_required` refusal now names the Runs tab |

### Two details worth reviewing

**The Run button is never disabled for an empty staged set** — it becomes **"Run all"**. Sending an explicit empty `dag_filter` means "rebuild everything", and it's the only way back when outputs are missing or corrupt but the fingerprints claim they're built. Absent and empty are deliberately different requests; don't collapse them.

**`staged` rides on `/changes/`, not on the run endpoint.** That endpoint returns an object, so new keys are invisible to an older client — and the viewer ships on its own schedule at both ends (cloud vendors it at a lagging release tag; local upgrades when the user does). Changing the run endpoint from array to object would silently kill the run indicator on any viewer older than the server. `/changes/` is also already called after every resource write, so the staged list and the tab dot update **on save, with no new request and no polling latency**.

---

# 2. Cancelling a run

A run that wedges leaves the editor spinning and the commit gate shut, with no way out short of editing the database — which is what we had to do by hand for a run stuck mid-build. This gives its owner a button.

- Rendered only for **queued and running** runs. Cancelling a finished one is meaningless and the endpoint 409s on it. Queued matters as much as running — that's the state users actually get stuck in.
- What the button promises is the **state change**: the run goes `canceled`, the spinner stops, the commit gate reopens.
- Stopping the compute is best-effort server-side (a Kubernetes Job is deleted; a warm-pool run finishes in the background with its results discarded). Hence "Cancel run" rather than anything implying a kill.
- A 409 is deliberately **not** an error — it only means the run terminated between render and click, and the refetch shows that.

| file | |
|---|---|
| `viewer/.../RunsView.jsx` | `CancelRunButton`, shown in `RunDetail` while the run is active |
| `viewer/.../api/branching.js` | `cancelRun` — returns `{status, body}` rather than throwing |
| `viewer/.../config/urls.js` | `runCancel` route |

**Not implemented locally.** `FilteredRunner` runs on a daemon thread with no cancellation primitive, and faking it would be worse than its absence — so `POST /api/runs/<id>/cancel/` is cloud-only for now, and the button will fail under `visivo serve`. Worth hiding it when the endpoint isn't available; called out rather than quietly left.

---

## Dependencies

Needs the matching core work — `POST /api/runs/<id>/cancel/`, `User.run_trigger` + `/api/me/preferences/`, and `unbuilt_changes` on `/changes/`. **Merge and deploy core first**, or the button and the toggle call endpoints that don't exist. Core then picks the viewer up via `yarn copy`.

Note the cloud default is manual, so after that deploy an edit stages instead of running until someone presses Run. The staged list and the tab dot are what make that legible — ship the API and the UI together rather than in separate deploys.

## Testing

- `poetry run pytest tests/` — 2064 pass. New: `tests/server/test_user_config.py` (defaults, round-trip, comments and unknown keys preserved, corrupt YAML falls back). Extended `tests/server/test_run_views.py` with trigger-mode, staged-set, run-endpoint and preferences coverage.
- `yarn test` — 5181 pass across 314 suites. New coverage for the staged panel, the tab dot, `triggerRun`, and the store hydration — including the case where the **server is older than the viewer and sends no staged keys**.
- `yarn lint` clean; `black --check visivo/` clean.

### DIFF
--- a/viewer/src/api/branching.js
+++ b/viewer/src/api/branching.js
@@ -116,6 +116,26 @@ export const fetchRunLog = async runId => {
 };
 
 /**
+ * Stop a run in flight. POST /api/runs/<id>/cancel/.
+ *
+ * Returns the raw {status, body} rather than throwing, because 409 isn't an
+ * error the user needs shouting about — it just means the run reached a terminal
+ * state before their click landed, and a refetch will show that.
+ *   200 {state, execution_stopped}  — canceled
+ *   409 {detail}                    — already finished
+ */
+export const cancelRun = async runId => {
+  const response = await apiFetch(getUrl('runCancel', { runId }), { method: 'POST' });
+  let body = null;
+  try {
+    body = await response.json();
+  } catch {
+    body = null;
+  }
+  return { status: response.status, body };
+};
+
+/**
  * Commit (publish) a draft. POST /api/projects/<id>/commit/ {message}.
  *
  * Returns the raw {status, body} so the caller can branch on the gates the
--- a/viewer/src/api/branching.js
+++ b/viewer/src/api/branching.js
@@ -103,6 +103,70 @@ export const fetchRuns = async projectId => {
 };
 
 /**
+ * Launch a run. POST /api/projects/<id>/run/.
+ *
+ * Send NO dag_filter to build what's staged — the button's normal action, scoped
+ * to exactly the list the Run view is showing. Send an explicit `dag_filter: ''`
+ * to rebuild everything ("Run all"), which is the only way back when outputs are
+ * missing but the fingerprints claim they're built. Absent and empty are
+ * genuinely different requests; don't collapse them.
+ *
+ * Returns the raw {status, body} like cancelRun/commitDraft, because 409
+ * ("a run is already in flight") is a state to render, not an exception.
+ */
+export const triggerRun = async (projectId, { dagFilter } = {}) => {
+  const response = await apiFetch(getUrl('projectRun', { projectId }), {
+    method: 'POST',
+    headers: { 'Content-Type': 'application/json' },
+    body: JSON.stringify(dagFilter === undefined ? {} : { dag_filter: dagFilter }),
+  });
+  let body = null;
+  try {
+    body = await response.json();
+  } catch {
+    body = null;
+  }
+  return { status: response.status, body };
+};
+
+/**
+ * The requesting user's own preferences. GET /api/me/preferences/ ->
+ * {run_trigger}. Returns null where there's no such endpoint (dist), so the
+ * caller can simply not render the control.
+ */
+export const fetchPreferences = async () => {
+  try {
+    const response = await apiFetch(getUrl('mePreferences'));
+    if (response.status === 200) {
+      return await response.json();
+    }
+  } catch {
+    // No URL configured for this environment — same as "not available".
+  }
+  return null;
+};
+
+/**
+ * PUT /api/me/preferences/ -> the saved {run_trigger}. Returns null on failure;
+ * the caller reverts its optimistic update.
+ */
+export const savePreferences = async preferences => {
+  try {
+    const response = await apiFetch(getUrl('mePreferences'), {
+      method: 'PUT',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify(preferences),
+    });
+    if (response.status === 200) {
+      return await response.json();
+    }
+  } catch {
+    // fall through
+  }
+  return null;
+};
+
+/**
  * A single run's captured log. GET /api/runs/<id>/logs/ -> {state, logs,
  * error_json}. The runner streams the log live while the run executes (the
  * editor tail-polls this), then it settles into the final static log.
--- a/viewer/src/api/branching.js
+++ b/viewer/src/api/branching.js
@@ -2,13 +2,17 @@ import { getUrl } from '../contexts/URLContext';
 import { apiFetch } from './utils';
 
 /**
- * Branching API (core/Django only) — draft / branch / commit / run endpoints.
+ * Branching API — the shape of the project you are editing: capabilities,
+ * draft, branch, discard, the pending-changes set, and commit.
  *
- * These endpoints exist only in the cloud (core) backend, not in local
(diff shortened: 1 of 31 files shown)


viewer/src/api/branching.js
```diff
@@ -102,6 +102,70 @@ export const fetchRuns = async projectId => {
...
+ * Send NO dag_filter to build what's staged — the button's normal action, scoped
+ * to exactly the list the Run view is showing. Send an explicit `dag_filter: ''`
+ * to rebuild everything ("Run all"), which is the only way back when outputs are
+ * missing but the fingerprints claim they're built. Absent and empty are
+ * genuinely different requests; don't collapse them.
+ *
+ * Returns the raw {status, body} like cancelRun/commitDraft, because 409
+ * ("a run is already in flight") is a state to render, not an exception.
+ */
+export const triggerRun = async (projectId, { dagFilter } = {}) => {
```
jpjsonic: does it make sense to have this api call in the branching file?

visivo/server/user_config.py
```diff
@@ -0,0 +1,87 @@
...
+"""
+
+import os
+from pathlib import Path
+
+import ruamel.yaml
+
+from visivo.logger.logger import Logger
+
+# Local serve runs on your own machine, where a run is a sub-second in-process
```
jpjsonic, line 21: Do we want to have different behavior? I guess we can always make these toggles that the user can set on their project too.

jpjsonic (approved):
Nothing blocking looks good! Just left a few questions / comments

visivo/server/user_config.py
tgsoverly: Yea, probably will need to have global properties too, but we can see.

viewer/src/api/branching.js
tgsoverly: Not really.  :-)

PR #545, opened 2026-07-29
Let the host set where the viewer is mounted

tgsoverly:
Gives a host a way to say **where the viewer is mounted**, so the viewer's own navigations land there first time.

## The problem

Studio serves the viewer at the router root, so its internal navigations are written root-absolute — `navigate('/explorer')`, `navigate('/workspace/dashboard/…')`, `to="/runs"`, ten in total.

A host that mounts the viewer deeper had no way to say so. The cloud app mounts it at `/:account/:stage/:project/…`, so it caught those escaped paths at its own router root and redirected them back into the project prefix (`ViewerEscapeRedirect` + `projectPrefix.js`, remembering the prefix in `sessionStorage`).

That works, but every such click costs **two navigations**, and in between the URL is outside the project route — so the host unmounts the entire project subtree and remounts it. In core that is what makes a `<Loading>` placeholder flash on each sidebar click.

Repairing the escape after the fact is the wrong shape. Being told the base removes the round trip.

## What this adds

**`contexts/viewerBase`** — `setViewerBase` / `getViewerBase` / `viewerPath`.

**`hooks/useViewerNavigate`** — `useNavigate` with root-absolute paths resolved against the base.

Module state rather than a React context, matching `setAuthHeaderProvider` in `api/utils.js`: navigation also happens from stores and callbacks that aren't inside the tree — the workspace tab loop already registers its `navigate` this way — and a context would only be reachable from components.

**Studio is unchanged.** The default base is `''`, so `viewerPath` returns its argument and `useViewerNavigate` is `useNavigate`.

## Changed call sites

Nine components move from `useNavigate` to `useViewerNavigate`; `RecordRunStatus`'s `to="/runs"` becomes `to={viewerPath('/runs')}`.

`Workspace` already derived a mount base by slicing the path for a `workspace` segment. It now prefers the injected base and keeps that derivation as a fallback — the derivation can only work once you are *already* on a `/workspace` URL, which is precisely when it is least needed.

## Two details worth reviewing

**`viewerPath` is idempotent.** Call sites get converted piecemeal, so one that rebases a value its caller already rebased must not double the prefix. Covered by a test, along with the near-miss case (`/acmex` under base `/acme` is still rebased).

**`useViewerNavigate` forwards only the arguments it was given.** My first version always passed a second argument, turning `navigate('/x')` into `navigate('/x', undefined)`. Identical at runtime, but it broke eight callers' test assertions — a leaky drop-in rather than a transparent one. Worth knowing because it is the kind of thing that looks harmless in review.

## Testing

16 new tests across the two modules — rebasing, idempotency, relative and absolute paths left alone, history deltas, location objects, and the trailing-slash / non-string base cases.

Full suite: **316 suites, 5170 tests, green**, with no existing test modified.

## Follow-up in core

Core sets the base in `ProjectLayout` and deletes `ViewerEscapeRedirect` + `projectPrefix.js` and the five escape routes. That change needs this vendored first, so **merge this before the core PR**.

### DIFF
--- a/viewer/src/components/explorer/ExplorerOverlay.jsx
+++ b/viewer/src/components/explorer/ExplorerOverlay.jsx
@@ -1,11 +1,12 @@
 import React, { useCallback, useEffect, useState } from 'react';
-import { useNavigate, useParams, useSearchParams } from 'react-router-dom';
+import { useParams, useSearchParams } from 'react-router-dom';
 import { PiX, PiArrowsClockwise } from 'react-icons/pi';
 import useStore from '../../stores/store';
 import { getTypeColors } from '../views/common/objectTypeConfigs';
 import { emitWorkspaceEvent } from '../views/workspace/telemetry';
 import { ExplorerRoundTripProvider } from './ExplorerRoundTripContext';
 import ExplorerPage from './ExplorerPage';
+import { useViewerNavigate } from '../../hooks/useViewerNavigate';
 
 /**
  * ExplorerOverlay — VIS-778 / J-2.
@@ -33,7 +34,7 @@ const slotLabel = slot => {
 };
 
 const ExplorerOverlay = () => {
-  const navigate = useNavigate();
+  const navigate = useViewerNavigate();
   const { dashboardName } = useParams();
   const [searchParams] = useSearchParams();
   const slot = searchParams.get('slot') || 'new';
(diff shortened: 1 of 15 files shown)


jpjsonic (approved):
Looks good, I'll rebase my pr to this.

PR #548, opened 2026-07-30
Extract DashboardExplorerRedirect out of LocalRouter

tgsoverly:
Small refactor, but it unblocks core's CI — see *Why now*.

## The problem

The cloud app (core) mounts `/workspace/dashboard/:name/explorer` too, and after Explore 2.0 that route is `DashboardExplorerRedirect`. It could only be imported from `LocalRouter.jsx` — which core cannot import:

```js
export const localURLConfig = createURLConfig({ environment: 'server' });
setGlobalURLConfig(localURLConfig);          // module scope
const LocalRouter = createBrowserRouter(...) // module scope
```

Importing it from core would silently repoint **every API call in the cloud app** at the local-serve URL config, and stand up a second browser router. The build succeeds either way — this only surfaces at runtime — which is exactly why it's worth extracting rather than leaving a comment telling the next person not to do it.

## The change

`DashboardExplorerRedirect` moves to its own module and is re-exported from `LocalRouter`, so existing callers are untouched. The hooks it was the sole user of (`useEffect`/`useRef`/`useState`/`useStore`) come out of `LocalRouter` with it.

`LocalRouter.test.jsx` passes **unmodified** — that's the check that the re-export preserved the contract, not an assurance.

## Why now

core#311 is red on `build-app`:

```
Could not resolve "./viewer/components/explorer/ExplorerOverlay" from "src/App.jsx"
```

Core vendors the viewer from visivo `main` at build time, so #527 deleting `ExplorerPage`/`ExplorerOverlay` broke it. Core's router now follows the same cutover — Explorer as a Workspace destination, `/explorer` as a redirect — and mounts this component for the dashboard round trip. It needs to exist on `main` for that build to pass.

Separated from visivo#546 deliberately: that PR is about async source ops, this is a standalone refactor, and core is blocked on only this half.

Lint clean; 14 `LocalRouter` tests green.

### DIFF
--- a/viewer/src/LocalRouter.jsx
+++ b/viewer/src/LocalRouter.jsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, useState } from 'react';
+import React from 'react';
 import { Navigate, Route, createBrowserRouter, createRoutesFromElements, useParams } from 'react-router-dom';
 import { futureFlags } from './router-config';
 import { loadProject } from './loaders/project';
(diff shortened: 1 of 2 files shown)


tgsoverly:
Not needed — closing.

The whole justification was that core had to import this component for the `/workspace/dashboard/:name/explorer` round trip, and couldn't get it from `LocalRouter` because that module calls `setGlobalURLConfig` and `createBrowserRouter` at module scope.

Core doesn't mount that route at all: the Explorer has never been deployed in cloud, so there are no live links into the dashboard round trip or the old standalone `/explorer` page, and nothing to keep resolving. core#311 drops both routes instead of porting the redirect.

The `LocalRouter` module-scope side effects are still real, and would still bite anyone importing that file from the cloud app. But with no consumer, extracting this component is churn — worth doing when something actually needs it.

Reverted from visivo#546 too, so its net diff is unchanged.

PR #566, opened 2026-08-04
Deploy refuses to run against output that isn't there

tgsoverly:
Closes VIS-1129.

## The failure it prevents

Every collector guards its files with `os.path.exists` and drops what's missing. Harmless while a miss meant one genuinely absent file — but when the run's layout moved (VIS-1128), deploying against a `target/` built by an older version made **every guard fire at once**. The deploy uploaded no parquet and reported success.

The damage surfaced three layers away: core had no file for the ref, so it left `signed_data_file_url` as the author's local path; the browser requested that and got the SPA's `index.html`; DuckDB then reported

```
Table with name mevqdxvdsldydvswuhleavelgkica does not exist!
```

Nothing in that chain points at the cause.

## The check

`verify_run_output` runs **before anything uploads** — a half-finished deploy is worse than one that refuses to start.

It checks exactly what the collectors intend to collect, so it can't demand a file the deploy wouldn't have sent anyway:

- every insight has its envelope
- a **static** insight has its precomputed parquet
- a **dynamic** insight's dependent models have theirs (it owns none itself)
- every input has its envelope

```
Missing 2 asset(s) that `visivo run` produces. Run it, then deploy. Set STACKTRACE=true to list them.
```

The list is diagnostic rather than a to-do — the reader acts on the remedy, not on individual paths — so it sits behind `STACKTRACE` alongside the rest of the CLI's detail.

## Two fixture gaps it found

Both real, and both would have been silent:

- the deploy test never wrote the parquet a static insight produces;
- it never wrote anything for the insight `ProjectFactory` inlines into its dashboard, which a real run builds and uploads.

## Testing

Passes when the run produced everything; short message by default; full list under `STACKTRACE=true`.

**2312 passed.** The 3 failures in `tests/templates/test_samples.py` are pre-existing and environmental (`pyenv: visivo command not found`) — identical on `main`.

### DIFF
--- a/visivo/commands/deploy_phase.py
+++ b/visivo/commands/deploy_phase.py
@@ -872,6 +872,11 @@ def verify_run_output(project, output_dir):
     table under a name_hash nobody can trace. Checking here turns that into one
     sentence naming the remedy.
 
+    Scope is deliberately narrow: whether the files the run RECORDED are where
+    the deploy will look for them. Not whether every insight built — that is
+    the run's business, and an insight with no envelope is a pre-existing
+    condition this has no opinion on.
+
     Checks the run's OWN record of what it built — each envelope's ``files[]``
     — rather than re-deriving which artifacts should exist. An earlier version
     predicted them from ``insight.is_dynamic(dag)``, which is "has any Input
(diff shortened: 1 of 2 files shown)


visivo/commands/deploy_phase.py
```diff
@@ -857,6 +857,76 @@ async def process_models_async(models, output_dir, project_id, form_headers, jso
     _raise_if_any_failed(response_items, "Failed to create model records")
 
 
+def verify_run_output(project, output_dir):
```
jpjsonic, line 860: This is a great idea

jpjsonic (approved):

PR #567, opened 2026-08-04
The whole-project read answers with an envelope, not the tree

tgsoverly:
Companion to core#334 (VIS-1113), which drops the `project_json` column from core. This removes the blob from the other two places it lived — the `visivo serve` payload and the `visivo dist` bundle.

## What changed

`/api/project/` and `data/project.json` both shipped the entire dereferenced project as one `project_json` blob, and every consumer dug through it for one or two fields. Both now return the same envelope:

```json
{"id", "name", "project_dir", "config": {"defaults": …},
 "dashboard_count", "source_count", "created_at", "status"}
```

Resource lists already have their own endpoints; nothing needed the tree.

## A latent bug this fixes

The viewer reads project defaults at `project?.config?.defaults`. The server put the defaults dict **at** `config`, not under it:

```python
"config": project_data.get("defaults", {}),     # before
```

So that read has been silently missing under `visivo serve` and falling through to `project_json.defaults` — which is why the fallback looked redundant but wasn't. Removing the blob without fixing the shape would have quietly broken dashboard filter defaults locally. The envelope is now nested the way `api/project.js` documented all along, and both servers agree.

## Consumers

Most already had the correct read first and the blob as a dead clause, so they only lose the clause:

- `project?.project_json?.name || project?.name` → `project?.name` — TabStrip, Workspace, useWorkspaceTabShortcuts, ProjectHomePane, ProjectEditor
- `?? project?.project_json?.defaults` dropped — Project.jsx, WorkspaceDndContext, ProjectEditor

**Onboarding was the real dependency** — `project_dir`, plus source and dashboard counts, with no other source. That is why the envelope gains those three fields rather than the blob simply being deleted:

- `OnboardingFlow` → `project?.project_dir`, `project?.name`
- `onboardingManifest` `connect_source` → `project?.source_count` (the `sourceStore` clause ahead of it is unchanged and still primary)
- `commonStore.evaluateIsNewProject` → `project?.name` + `project?.dashboard_count`

## Kept deliberately

`flask_app._project_json` stays. `matches_served_project()` compares it to tell a genuine external YAML edit (drops drafts) from a no-op save that recompiles identically (drafts must survive). That is a fingerprint, not a payload.

## Tests

- python `2318 passed`, 0 failures
- viewer `6659 passed`, 355 suites, `yarn lint` clean, `black --check` clean
- `test_serve` / `test_dist` now assert the envelope's shape and that `project_json` is **absent**, rather than that it is present

Note `ExplorationPromoteModal`'s "accepting the fallback offer…" and one `WorkspaceDndContext` case are flaky — they fail intermittently on a clean tree at this base too. Both pass on re-run; the green figures above are a full clean run.

### DIFF
--- a/tests/commands/test_dist.py
+++ b/tests/commands/test_dist.py
@@ -82,7 +82,13 @@ def test_dist_creates_dist_folder(setup_project, output_dir, dist_dir):
         data = json.load(project_json)
         assert "id" in data
         assert "created_at" in data
-        assert "project_json" in data
+        # The bundle ships the same envelope the server serves, not the
+        # whole dereferenced project tree.
+        assert "project_json" not in data
+        assert data["name"] == project.name
+        assert "defaults" in data["config"]
+        assert "dashboard_count" in data
+        assert "source_count" in data
 
 
 def test_dist_errors_without_data(setup_project, output_dir, dist_dir):
--- a/tests/commands/test_serve.py
+++ b/tests/commands/test_serve.py
@@ -59,7 +59,12 @@ def test_serve(output_dir):
     client = server.app.test_client()
     response = client.get("/api/project/")
     response_json = json.loads(response.data)
-    assert "project_json" in response_json
+    # The canonical envelope, not the whole project tree.
+    assert "project_json" not in response_json
+    assert response_json["name"] == project.name
+    assert "defaults" in response_json["config"]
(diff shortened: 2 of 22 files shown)


jpjsonic (approved):
🚀

PR #579, opened 2026-08-06
Only ask for the explorer diff where it exists

tgsoverly:
### What the endpoint does

`/api/explorer/diff/` compares the explorer's **unsaved working state** against the project's published objects, returning `new` / `modified` / `null` per object — that's what drives the status badges before anything is saved.

Answering that needs the project to compare against, which is why only `visivo serve` implements it. Cloud 404s, **on every explorer edit**.

### Why nothing appeared broken

The store already caught the failure and set `explorerDiffResult` to `null` — the same value a successful call would have produced in that situation. So the request bought nothing but a red line in the network panel. That's also why it survived: there was no symptom beyond the console.

### The gate

The call is now gated on the server advertising the endpoint.

`urls.js` couldn't express this: its `server` / `dist` split is about the deployed viewer, and **core is a server** — so declaring the endpoint there would still report it available in cloud. It goes through **capabilities** instead, which already differ per host (`local_filesystem` is `true` on serve, `false` in core) and which the client is meant to *ask* rather than infer where it's running. Serve advertises `explorer_diff: true`; cloud omits the key.

I added an explicit flag rather than reusing `local_filesystem`. They happen to agree today, but they mean different things — one is "the server can open the project's files", the other is "the server implements this endpoint" — and collapsing them is the kind of thing that bites the next person.

**Skipped rather than attempted-while-unknown.** The point is to stop firing a request that can't succeed, and `fetchExplorerDiff` re-runs on the next explorer edit once capabilities have loaded, so nothing is permanently lost on serve.

### Testing

**Viewer 6724** (358 suites) and **Python 2320**, lint clean, black clean.

New: the endpoint isn't called when the server doesn't offer it, and isn't called before capabilities load. The Flask capabilities shape test now asserts the flag, so the two halves can't drift. The three existing payload tests seed the capability — they drive the request itself, so they now have to stand on a server that offers it.

### What this does not do

Cloud still has no badges — which is where it already was. **Core could implement the comparison**: it holds both the draft and the published config for every resource. That's worth doing on its own merits, not smuggled in behind a 404. Happy to pick it up if you want it.

One failure in the full run: `ExplorationPromoteModal`'s "accepting the fallback offer…". I checked it against clean `main` rather than assume — it fails **2 of 3 isolated runs there**, so it's pre-existing.

### DIFF
--- a/tests/server/views/test_commit_views_coverage.py
+++ b/tests/server/views/test_commit_views_coverage.py
@@ -144,6 +144,10 @@ class TestProjectScopedContract:
         # than on secret_keys being empty.
         assert data["secrets_required"] is False
         assert isinstance(data["secret_keys"], list)
+        # serve owns the project files, so it can answer /api/explorer/diff/.
+        # Cloud omits the key and the client skips the call rather than firing
+        # a request that 404s.
+        assert data["explorer_diff"] is True
 
     def test_draft_echoes_project_id(self, client):
         data = client.post("/api/projects/proj1/draft/").get_json()
--- a/viewer/src/stores/explorerStore.js
+++ b/viewer/src/stores/explorerStore.js
@@ -1845,6 +1845,22 @@ const createExplorerSlice = (set, get) => ({
 
   fetchExplorerDiff: async () => {
     const state = get();
+
+    // `/api/explorer/diff/` compares the explorer's unsaved working state
+    // against the project's published objects — which needs the local project
+    // to compare against, so only `visivo serve` implements it. Cloud answered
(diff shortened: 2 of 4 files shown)


tgsoverly:
Superseded — we want the diff working in cloud rather than skipped in the viewer. Closing in favour of implementing `/api/explorer/diff/` in core.

Keeping the notes here since they're the reason the endpoint is worth implementing rather than dropping: it compares the explorer's UNSAVED working state against published objects to drive the new/modified badges, and cloud has both halves (draft config and published config) to compare.

PR #583, opened 2026-08-07
VIS-1191: lazy-load CLI subcommands (only import what's called)

tgsoverly:
`command_line.py` eagerly imported **all 14 subcommands** at module load, pulling the entire model / query / source-connector graph into **every** invocation — even `visivo --version`. A `LazyGroup` now imports each subcommand's module only when that command is invoked (or the group is listed), so the command that runs pays for itself alone.

This is the biggest startup lever and is **freezer-independent** — it helps the interpreted CLI, the current PyInstaller build, and the coming Nuitka build alike — and it cuts the import cost of **every `visivo run` a Celery worker spawns**.

## Measured (interpreted, `--version`)

| | modules imported | wall clock |
|---|---|---|
| eager (main) | 1840 | 2.62 s |
| lazy (this PR) | **542** (−70%) | **1.55 s** (−40%) |

`visivo run` likewise now imports only `run`'s graph, not deploy/serve/dbt/dist/etc.

## Tests

- Importing `command_line` no longer pulls in run/deploy/serve/dbt/dist (verified in a **fresh interpreter** so other tests can't mask a regression).
- Every command is still listed in `--help` and resolves lazily, with its CLI name matching the mapping key (guards against a command's own name drifting from the key).
- Full `test_command_line.py` + version/run/compile/list suites green; `black` clean.

Note: `--help` still imports all commands (Click renders each command's short help), but that path is human-only — the runner and scripted paths (`--version`, `run`, …) get the full win.

Closes VIS-1191.

### DIFF
--- a/tests/test_command_line.py
+++ b/tests/test_command_line.py
@@ -86,3 +86,46 @@ def test_generic_exception_still_reports_the_issue_url():
 
     assert "An unexpected error has occurred" in output
     assert "Click here to report this issue" in output
+
+
+def test_importing_command_line_does_not_eagerly_import_subcommands():
+    """The lazy-load contract (VIS-1191): importing the CLI entry must NOT pull
+    in every subcommand's module graph. A fresh interpreter is used so other
+    tests' imports can't mask a regression.
+
+    `visivo run` should pay the import cost of `run` alone — not deploy, serve,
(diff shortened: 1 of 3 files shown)


tgsoverly:
Just closing for now, the performance is better, but don't know if we can switch to nuitka for good yet.

PR #587, opened 2026-08-10
Enforce cli_version match on deploy only, not on every run

tgsoverly:
## Why

Running a project in the cloud failed with:

> Error: The project specifies 2.0.3, but the current version of visivo installed is 2.1.0. Your project version needs to match your CLI version.

`CliVersionValidator` runs inside `Project`'s `@model_validator`, so it fires on **every** `Project(**data)` construction — run, compile, serve — not just deploy, and it requires an **exact** match. The cloud runner installs one visivo (2.1.0), but a project deployed by an older CLI is *recorded* at its version (2.0.3), and core's assembler writes that into the run — so the run is blocked. Every existing cloud project breaks the same way the moment the runner is upgraded.

It never bit locally because a project's `cli_version` **defaults to the installed CLI's version**, so it always matched.

## Change

A run should use whatever visivo is installed, regardless of the version that authored the project (forward *and* backward). Version match only matters when **publishing to the shared cloud project**. So:

- Remove `CliVersionValidator` from the always-on `ProjectValidator` list (it no longer runs on construction/run/compile/serve).
- Call it **explicitly in `deploy_phase`**, right after the project is parsed — deploy still rejects a mismatch.

## Tests

- `test_cli_version_mismatch_does_not_block_construction_or_run` — a project recorded at another version now constructs without raising.
- `test_cli_version_validator_still_rejects_on_demand_for_deploy` — the validator still raises when called (the deploy path).
- Full project + validators + serializer suites green (63).
- **End-to-end:** a project with `cli_version: "2.0.3"` compiles + runs on 2.1.0 with no version error.

This is what unblocks cloud runs of already-deployed projects (pairs with #586, which unblocks the runner *build*).

### DIFF
--- a/tests/models/test_project.py
+++ b/tests/models/test_project.py
@@ -506,3 +506,25 @@ def test_named_child_nodes_keeps_duplicate_model_scoped_names_distinct():
     assert "model_a" in named_nodes
     assert "model_a.avg_total" in named_nodes["model_a"]["direct_children"]
     assert "model_b.avg_total" not in named_nodes["model_a"]["direct_children"]
+
+
+def test_cli_version_mismatch_does_not_block_construction_or_run():
+    """A run uses whatever visivo is installed, so a project recorded at another
+    CLI version must still construct (and run). Version match is deploy-only —
+    the cloud runner runs projects deployed by older CLIs."""
+    from visivo.version import VISIVO_VERSION
+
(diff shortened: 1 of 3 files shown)


tgsoverly:
Superseded by the major-version compatibility approach — the check stays on run (not deploy-only); only the comparison changes from exact-match to major-based.

PR #597, opened 2026-08-12
Scan every schema, batch column reflection, and stop hiding failures

tgsoverly:
A Postgres source reported **"no tables"** while sitting on a database with **213 of them**. Three separate defects stacked into that one symptom.

## 1. Only the default schema was scanned

`_get_available_tables_for_schema` looked at the connection's default schema, plus `db_schema` if set. The tables were in a named schema and `public` was genuinely empty, so a correct-looking zero came back:

| schema | tables |
|---|---|
| `rnacen` | **186** |
| `rnacen_django_test` | 23 |
| `public` | **0** ← the only one scanned |

Now every non-system schema is scanned. An explicit `db_schema` still wins and narrows to itself. Table keys keep their historical shape — bare in the default schema, `schema.table` elsewhere — because stored schema payloads and downstream code index on it.

## 2. Columns were reflected one table at a time

Measured against a remote warehouse: **~1.9s per table**, so a 186-table schema took **~6 minutes** and whatever was waiting gave up long before it finished — leaving an empty schema that looked like an empty database. Columns are now reflected per schema in a single round trip.

**Batching falls back rather than replacing the old path.** SQLAlchemy's multi-column query joins `pg_collation`/`pg_constraint`, and a locked-down server can deny those while still permitting ordinary per-table reflection:

```
ProgrammingError: (psycopg2.errors.InsufficientPrivilege)
permission denied for table pg_collation
```

Verified against exactly such a server. Batching without the fallback would have turned "slow" into "broken" on the source that prompted this.

## 3. Failures were swallowed

Every layer was `except: pass` or `except: return []`, so a permissions error, a timeout and an empty database were indistinguishable. `metadata` now carries `errors` and `scanned_schemas`, and zero tables always arrives with a reason.

That is what identified the underlying cause of the original report: the account could read table *names* but not `pg_collation`, so **no column could be reflected at all** — a server limitation the old code reported as an empty schema.

## Verification

- Full suite: **2343 passed** (was 2334), black clean.
- 9 new tests covering all three behaviours, including the fallback and the "empty must carry a reason" invariant.
- Exercised end to end against the live source: fast path confirmed on SQLite, fallback + error reporting confirmed against the restricted Postgres server.

### DIFF
--- a/tests/jobs/test_run_source_schema_job.py
+++ b/tests/jobs/test_run_source_schema_job.py
@@ -159,3 +159,93 @@ class TestSourceSchemaAction:
         source = seeded_source([Seed(table_name="raw", args=["false"])])
         result = action(source_to_build=source, output_dir=temp_folder())
         assert not result.success
+
+
+class TestSchemaBuildOutcomes:
+    """Three outcomes, because two were not enough.
+
+    Downstream jobs cannot run without a schema — SQLGlot needs it to expand
+    ``SELECT *``, and without it a model's only column is a literal ``*`` of
+    unknown type. So "connected but resolved nothing" has to fail HERE, next to
+    the evidence, rather than surfacing later as "Column 'id' not found" on a
+    job that is not the cause.
+
+    But a partial schema is genuinely useful, and an empty database is not an
+    error at all.
+    """
+
+    @staticmethod
+    def _sqlite_with(tables):
+        import sqlite3
+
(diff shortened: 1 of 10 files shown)


jpjsonic (approved):
Should just use the information schema rather than hitting each table directly? I think we do that for snowflake already

PR #604, opened 2026-08-16
VIS-1230: reopen an untouched source browse-seed instead of duplicating it

tgsoverly:
## Problem

Clicking a source tile in the Explorer always minted a **new** exploration seeded from that source — and because the name is derived from the source (`"<source> exploration"`), repeated clicks piled up **indistinguishable same-named explorations**. The existing guard (`creatingRef`, VIS-1084/1086) only blocks *rapid/concurrent* clicks; a **deliberate** second click (create tile → go back to Explorer Home → click the same source again) still created a duplicate.

## Fix

`handleSourceTile` now first looks for an existing **untouched browse-seed** for that source — `seededFrom` is `{type:'source', name}` and it isn't yet "real" per `isExplorationVisibleInGallery` — and **reopens** it instead of creating another. Clicking a source is a browse gesture, so a second click reopens the browse you already have.

A seed you've actually worked in *is* gallery-visible, so it's left alone and a genuinely new exploration is created — you can still keep several explorations of the same source once you've started using the earlier one. An untouched seed for a *different* source is never reused.

## Tests

Three new cases in `ExplorerHomePane.test.jsx`:
- reopens an existing untouched browse-seed (no `createExploration`, focuses the existing tab),
- still creates fresh when the only existing exploration for that source has been worked in (renamed away from the seed default),
- doesn't reuse an untouched seed belonging to a different source.

Full suite green (38 passed); lint clean.

Closes VIS-1230.

### DIFF
--- a/viewer/src/components/views/workspace/homes/ExplorerHomePane.jsx
+++ b/viewer/src/components/views/workspace/homes/ExplorerHomePane.jsx
@@ -144,6 +144,22 @@ const ExplorerHomePane = () => {
 
   const handleSourceTile = async sourceName => {
     if (creatingRef.current) return;
+    // VIS-1230: reuse an existing UNTOUCHED browse-seed for this source instead
+    // of minting another exploration with the identical seed-derived name
+    // (`<source> exploration`). Clicking a source is a browse gesture, so a
+    // second click should reopen the browse you already have, not pile up
+    // indistinguishable same-named copies. A seed the user has actually worked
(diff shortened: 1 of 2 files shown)


tgsoverly:
Superseded — the 'reopen existing seed' approach was replaced. Follow-ups: (1) source-click explorations should get unique space-numbered names like '+ New exploration' does, (2) fix leaked state where a source-seeded exploration also gets an extra default query. New PRs to come.

PR #609, opened 2026-08-19
VIS-1235: promote stops at the draft/cache layer, not commit

tgsoverly:
## VIS-1235 — Explorer promote force-commits: stop at the draft/cache layer

Promoting an exploration ("Save to project") saved each object to the draft
cache **and then force-called `commitChanges()`**, writing YAML on every
promote and swallowing any commit failure in a bare `catch {}` (promote still
reported `success: true`). This diverges from how every other workspace edit
works — save to the draft cache, review in the commit panel, then commit
deliberately.

### What changed
1. **Promote no longer commits.** Dropped the force-commit block in
   `promoteExploration` (`viewer/src/stores/workspaceExplorationsStore.js`).
   Promoted objects now land as **pending changes** in the commit panel; the
   Commit button writes them to `project.visivo.yml` once, deliberately.
2. **Failure is surfaced, not swallowed.** With the `catch {}` gone, the
   returned `success` reflects the per-row save results — a save that doesn't
   land is reported to the caller (already covered by the store's
   partial-failure tests).
3. **Dead endpoint removed.** `POST /api/project/write_changes/`
   (`visivo/server/views/project_views.py`) was a caller-less direct
   `ProjectWriter` YAML write that bypassed the draft layer — deleted along with
   its now-unused import and the `writeChanges` viewer URL key
   (`viewer/src/config/urls.js`).

### Acceptance
- Promoting leaves promoted objects as pending changes in the commit panel and
  writes **no** YAML.
- Committing afterward writes them once.
- A save that doesn't land makes promote report failure.

### Tests
- `workspaceExplorationsStore.test.js` — rewrote the two commit-behavior tests
  to assert promote **never** force-commits; the existing partial-failure tests
  already cover `success` reflecting per-row saves. **100 store tests pass.**
- `yarn lint` clean.
- `poetry run black --check` clean; **798 server tests pass** (no test covered
  the removed endpoint).

### DIFF
--- a/viewer/src/stores/workspaceExplorationsStore.js
+++ b/viewer/src/stores/workspaceExplorationsStore.js
@@ -879,28 +879,17 @@ const createWorkspaceExplorationsSlice = (set, get) => {
         });
       }
 
-      // Persist the just-promoted drafts to the project's YAML files. The
-      // per-type `saveX` actions above only write each manager's *cached*
-      // (draft) tier (POST /api/<type>/<name>/ -> save_from_config -> in-memory
-      // cache); they never touch YAML. The commit endpoint
-      // (/api/projects/<id>/commit/ -> ProjectWriter) is the ONLY path that
-      // writes project files under `visivo serve`, so without this "Save to
-      // project" reported success while project.visivo.yml was never written —
-      // a new-user blocker on a fresh, empty project. Gated on an actual save
-      // so an all-invalid promote never fires a no-op commit. commitChanges
-      // itself no-ops when there is no active project (e.g. trimmed test
-      // stores), so this is safe.
-      if (results.some(r => r.success)) {
-        // A commit (write-to-disk) failure must NOT abort the promote flow: the
-        // objects are already saved to the draft cache, commitChanges records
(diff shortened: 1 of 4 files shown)


viewer/src/stores/workspaceExplorationsStore.js
```diff
@@ -879,28 +879,17 @@ const createWorkspaceExplorationsSlice = (set, get) => {
...
-      // so an all-invalid promote never fires a no-op commit. commitChanges
-      // itself no-ops when there is no active project (e.g. trimmed test
-      // stores), so this is safe.
-      if (results.some(r => r.success)) {
-        // A commit (write-to-disk) failure must NOT abort the promote flow: the
-        // objects are already saved to the draft cache, commitChanges records
-        // its own error in `commitError`, and downstream steps (dashboard
-        // placement, return_to consumption) must still run.
-        try {
-          await get().commitChanges?.();
```
jpjsonic, line 899: Makes sense, this is probably how this should work

jpjsonic (approved):

PR #612, opened 2026-08-20
VIS-1224: Explorer panes reuse the standard edit panel (+ ⋮ Rename, insight-above-chart)

tgsoverly:
## VIS-1224 — Explorer panes reuse the standard edit panels

Brings the Explorer Build-rail insight & chart panes fully in line with the standard RightRail edit panels — same panel, shared internals — while keeping everything the Explorer needs (stacked collapsible insights, live write-through, drag-and-drop, Save-as-metric).

### Shared inner panels (rendered by both surfaces)
- **`InsightEditFormFields`** — Basic Information + Visualization Props (controlled, with DnD passthrough). Used by the standard `InsightEditForm` **and** the Explorer `InsightBuildSection`.
- **`ChartEditFormFields`** — Basic Information + insight-selection slot + Layout. Used by the Explorer `ChartBuildSection`.

### Explorer panes (colored sidebar → SelectionChip header)
- The bespoke panes are thinned to a neutral collapsible shell; the body is the shared standard panel. Headers now show the **SelectionChip identity** (type icon + name + TYPE), keeping collapse / status dot / ⋮ menu / remove.
- Name edits write through live (no Save button — "Save to project" stays the promote step); editable only for unsaved objects; the ⋮ Rename focuses the Basic Information field.
- Draft **DnD preserved**: property-zone, interaction-zone (insight), plus Save-as-metric + advisory validation.
- Insight panes stack **above** the chart; a reusable portaled **`PanelMenu`** backs the ⋮ menus.

### Insight-on-chart model
- **Explorer chart pane** no longer has an insight section — its insights ARE the stacked panes above it. The rail's **"+ Add Insight"** is a plain create.
- **Standard chart edit panel** drops the inline per-insight props editor and gains the shared **`AddInsightMenu`** ("+ Add Insight" → New blank / Find existing) to compose the chart's insights. Extracted `AddInsightMenu` to a shared component.

### Verification
- Full `workspace` / `explorer` / `common` sweep: **4517 tests pass**, lint clean.
- Sandbox e2e: reorder + "⋮ menu opens un-clipped" pass; a parity probe confirms both panes render "Basic Information" + name fields + neutral headers.
- DnD-drag e2e stories fail **identically on `main`** in this sandbox (pre-existing/environmental) — this change doesn't touch the DnD router.

### Follow-ups (not in this PR)
- Default-item naming: use `-` (not `_`) in the Explorer's auto-generated names, and give the chart a default name. (Reverted a first attempt — a blind `_`→`-` pass hyphenated JS object keys; needs a targeted change + test-fixture migration. Small, deferred to keep this PR green.)
- Fully DRY the standard `ChartEditForm`/`InsightEditForm` onto the shared `…Fields` components (Explorer already consumes them; `InsightEditForm` already shares).

### DIFF
--- a/viewer/e2e/stories/exploration-build-rail.spec.mjs
+++ b/viewer/e2e/stories/exploration-build-rail.spec.mjs
@@ -7,7 +7,7 @@
  * `ExplorerRightPanel`). Covers 03-delivery-plan.md's Phase 3b gate:
  *
  *   1. Rail CRUD — Add Insight stacks a new section; Chart always renders
- *      above the Insight sections.
+ *      below the Insight sections (VIS-1224: insight panes stack on top).
  *   2. Type-switch cache — TracePropsEditor's own preserveTraceProps carries
  *      x/y across a scatter -> bar -> scatter round trip.
  *   3. Drop targets on TracePropsEditor fields render D8 typed ref pills —
@@ -139,7 +139,7 @@ test.describe('Exploration Build rail (Explore 2.0 Phase 3b)', () => {
     }
   });
 
-  test('Add Insight stacks a new section; Chart section always renders above the Insight sections', async ({
+  test('Add Insight stacks a new section; the Insight sections render above the Chart section', async ({
     page,
   }) => {
     await gotoExplorerHome(page);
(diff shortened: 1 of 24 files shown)


jpjsonic (approved):
This is great- I think the style is much more consistent vs. the previous design.

PR #626, opened 2026-08-25
Stop the pill/code editor flip-flopping (VIS-1240)

tgsoverly:
Phases 1–3 of **VIS-1238**. Closes **VIS-1240**, **VIS-1241**, **VIS-1242**.

## Phase 1 — Stop the flip-flopping (VIS-1240)

> "I have seen behavior in two ways… I am not sure what causes it to switch back and forth."

It wasn't a mode you were choosing. `showPill` was re-derived on **every render**, so it flipped for three unrelated reasons:

| Cause | What you saw |
|---|---|
| **Mid-typing** | `${ref(q).gdp}` momentarily parses → pill replaces the editor, caret lost; ` / 100` flips it back |
| **Async field tables** | `metrics`/`dimensions` start `[]`, so a bare ref is opaque until they load → raw code, then a pill |
| **Surface** | pills were gated on `droppable`, so the *same value* was a pill in the Build rail and raw text in the right rail |

Fixed by freezing the representation while the raw editor has focus, holding the decision during a fetch, and making rendering a property of the **value** (dropping is still gated on `droppable`).

Also: the static/query toggle **used to lie** — `currentMode` ORs in `isQueryMode`, so on a `?{...}` value the "static" button depressed and snapped straight back. It can't demote (a `?{...}` in a number input mangles), so it now says why.

*Measured, not guessed:* removing the `droppable` gate broke **1 test out of 6866** — the one asserting the split by name.

## Phase 2 — The pill becomes a self-contained editor (VIS-1241)

The menu could only change the **aggregation**; the `ref ▸ column` header was static text, the index lived in a separate badge, and every click committed and closed.

It's now a form — **Property · Use as · Index · Modifier**, with Apply/Cancel, committing once:

```
Property  [ gdp   ▾ ]   →  ?{ sum(${ref(m).gdp}) / 100 }[0]
Use as    [ SUM   ▾ ]            └───┬───┘ └─┬─┘  └┬┘
Index     [ [0]   ▾ ]         aggregation modifier index
Modifier  [ / 100    ]                     (inside)  (last)
```

- **Property picker** from `POST /api/model-schemas/<model>/`, requested inside the popover (which mounts only when open) so a panel of pills doesn't fire a request per row. Stays editable while loading; falls back to free text when a model has no schema.
- **Modifier** — the manual escape valve you asked for, applied *inside* the braces so an index can still follow.

**The modifier forced grammar work planned for Phase 4**, because without it applying a modifier reparsed as `opaque` and the pill vanished on Apply — reintroducing the very flip Phase 1 removed. `parse` gained a head-match pass with a balanced-paren scan (`AGG_WRAP`'s `\)\s*$` anchor is precisely why `sum(x) / 100` matched nothing). It stays conservative: a second `${ref()}`, arithmetic *inside* the aggregate, and non-preset functions all remain opaque, verbatim.

⚠️ **A dangerous consequence, caught and closed:** once modified expressions parse, `pillFieldSwap` would have offered to rewrite `?{${ref(m).region} = 'west'}` to a bare `?{${ref(region)}}` — **silently dropping the `= 'west'`**. Modified slots are now excluded from both swap detectors.

Per your request, the index is **hidden at its default** — every slot used to carry an "All values" badge restating the default; it now appears only when there's a real index to show.

## Phase 3 — Drop a model, then configure it (VIS-1242)

Three blockers, the first fatal:

1. **Models weren't draggable at all** — in neither drag list, so `draggable={false}`. Added a separate `PROPERTY_DRAG_TYPES` rather than extending `EXPLORATION_DRAG_TYPES`, which also drives "Add to exploration" (a test caught that side effect).
2. **The drop had no type guard** — any payload with a `name` silently wrote `?{${ref(<activeModel>).<objectName>}}`, so dropping a source or insight produced a dangling ref with no message. Now an explicit allowlist that says why.
3. **An unbound model ref had no pill** — `${ref(model)}` parsed as `opaque`, suppressing the pill, so a dropped model had nothing to configure. New `modelRef` kind; metric/dimension names still win (backend's global-name-first order), unknown refs still opaque.

A dropped model lands as `<model> ▸ choose a property` and **opens the same editor** a configured pill uses — keyed on the value *changing* into that state, so loading an insight doesn't pop menus.

## Verification

Sandbox, end to end:
- Right rail renders `AVG · local_test_table ▸ y` as a pill while a sibling `CASE WHEN` correctly stays raw text — representation following the value, not the panel
- Property list populated from the real endpoint (`["x","y"]`)
- Menu stays open across three edits, commits once
- Applying `/ 100` + `[0]` leaves a pill that **re-opens with both intact** — the round-trip that matters
- Zero console errors throughout

**6887 tests passing** (359 suites), `yarn lint` clean.

## Not covered
The right-rail form's Save button wasn't exercised in the browser (no `form-footer-save` in that view to click), so the *persisted* string rests on unit tests plus the earlier check of the exact form against the Python `QUERY_STRING_VALUE_PATTERN`. Phase 4 (VIS-1243) — configurable chips in `RefTextArea`, structured property paths — remains.

### DIFF
--- a/viewer/src/components/explorer/AddComputedColumnPopover.jsx
+++ b/viewer/src/components/explorer/AddComputedColumnPopover.jsx
@@ -3,6 +3,17 @@ import { createPortal } from 'react-dom';
 import { PiPlus, PiX, PiCheckCircle, PiWarningCircle, PiSpinner } from 'react-icons/pi';
 import { recordOnboardingAction } from '../onboarding/onboardingState';
 import { getTypeColors } from '../views/common/objectTypeConfigs';
+import { refKindsFor } from '../views/common/fieldTypes';
+
+// A "computed column" is the Explorer's name for a MODEL-SCOPED metric or
+// dimension: it promotes with a `parentModel`, and `project_writer._new` nests
+// it under `model.metrics` / `model.dimensions`. Nesting is what makes
+// `sql_model.py` reject any `ref()` in the expression, so this field's declared
+// ref vocabulary is empty — which is why the editor below is a bare textarea
+// and not `RefTextArea`. Read it from the registry rather than asserting it
+// here, so if the rule ever changes this surface follows. See VIS-1253.
+const COMPUTED_COLUMN_REF_KINDS = refKindsFor('metric', 'expression', { nested: true });
+const REF_PATTERN = /\$\{\s*ref\s*\(/;
 
 const DEBOUNCE_MS = 750;
 
@@ -203,6 +214,10 @@ const AddComputedColumnPopover = ({
           <label className="block text-xs font-medium text-secondary-600 mb-1">
             SQL Expression
           </label>
+          {/* Deliberately plain: no ref-insert menu, no @ mention, no drop
+              target. `COMPUTED_COLUMN_REF_KINDS` is empty, so every one of
+              those affordances would offer an insertion the backend rejects on
+              save. Do not "upgrade" this to RefTextArea. */}
           <textarea
             value={expression}
             onChange={handleExpressionChange}
@@ -211,6 +226,16 @@ const AddComputedColumnPopover = ({
             className="w-full px-2 py-1.5 text-xs font-mono border border-secondary-300 rounded bg-white focus:outline-none focus:ring-2 focus:ring-primary-500 resize-none"
             data-testid="computed-col-expression"
           />
+          {/* Caught while typing rather than as a Pydantic error after Save. */}
+          {!COMPUTED_COLUMN_REF_KINDS.length && REF_PATTERN.test(expression) && (
+            <p
+              className="mt-1 text-xs font-medium text-highlight-600"
+              data-testid="computed-col-ref-warning"
+            >
+              References aren&apos;t available here — a computed column reads columns from its
+              model directly. Saving with a <code>ref()</code> will fail.
+            </p>
+          )}
         </div>
 
         {isValidating && (
--- a/viewer/src/components/explorer/AddComputedColumnPopover.test.jsx
+++ b/viewer/src/components/explorer/AddComputedColumnPopover.test.jsx
@@ -274,4 +274,40 @@ describe('AddComputedColumnPopover', () => {
     expect(onValidate).toHaveBeenCalledTimes(1);
     expect(onValidate).toHaveBeenCalledWith('SUM');
   });
+
+  // A computed column is a MODEL-SCOPED (nested) metric/dimension, and
+  // `sql_model.py` rejects any ref() in a nested expression. The editor must
+  // therefore offer no ref affordances — and say so before Save, not after.
+  describe('refs are not available here (VIS-1253)', () => {
+    const openPopover = () => {
+      renderPopover();
+      fireEvent.click(screen.getByTestId('add-computed-column-btn'));
+    };
+
+    test('the expression field offers no ref-insert affordance', () => {
+      openPopover();
+      expect(screen.getByTestId('computed-col-expression')).toBeInTheDocument();
+      // No RefTextArea machinery: no + ref menu, no chips, no drop target.
+      expect(screen.queryByTestId('ref-text-area')).not.toBeInTheDocument();
+      expect(screen.queryByTestId('add-ref-button')).not.toBeInTheDocument();
+    });
+
+    test('a pasted ref is flagged inline while typing', () => {
+      openPopover();
+      fireEvent.change(screen.getByTestId('computed-col-expression'), {
+        // eslint-disable-next-line no-template-curly-in-string
+        target: { value: 'sum(${ref(orders).amount})' },
+      });
+      expect(screen.getByTestId('computed-col-ref-warning')).toBeInTheDocument();
+    });
+
+    test('a plain SQL expression is not flagged', () => {
+      openPopover();
+      fireEvent.change(screen.getByTestId('computed-col-expression'), {
+        target: { value: 'SUM(amount)' },
+      });
+      expect(screen.queryByTestId('computed-col-ref-warning')).not.toBeInTheDocument();
+    });
+  });
+
 });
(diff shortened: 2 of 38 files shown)


jpjsonic:
## Browser acceptance pass

Sandbox on `638afda1` (note: that's one commit past Phases 1–3 — the VIS-1243 parse work is on the branch too), `test-projects/integration`, Chromium at 1600×1000, backend :8001 / frontend :3001.

### Confirmed working

**Phase 1 (VIS-1240)**
- **Mid-typing freeze holds.** Typed `${ref(local_test_table).y}` character-by-character into `x`'s raw editor. Once the value fully parsed the editor *stayed raw and kept focus* — no pill swap, no caret loss. Appended ` / 100`, still raw. Blurred → became the pill `local_test_table ▸ y / 100`. This is the flip that used to happen.
- **Representation follows the value.** On `aggregated-bar-insight`, `y` (`avg(...)`) renders as `AVG · local_test_table ▸ y` while its siblings `x` and `marker.line.color` (both `CASE WHEN`) stay raw text, in the same panel.
- **The static toggle stopped lying.** On `?{...}` values the button is genuinely `disabled` with `title="Clear the expression to use a static value"`; on static values (`marker.color`, `marker.line.width`) it stays enabled and pressed. No depress-and-snap-back.

**Phase 2 (VIS-1241)**
- Menu stayed open across three edits (SUM → index `[0]` → modifier `/ 100`); the pill still read `AVG …` throughout, then committed once on Apply → `SUM · local_test_table ▸ y / 100[0]`. Nothing flipped to raw.
- **The round-trip that matters holds.** Re-opened: property `y`, SUM checked, index `[0]`, modifier `/ 100` — all intact.
- Property list came from the real endpoint (`x`, `y`). Index badge is hidden at default and appears only once set.
- "Manually edit field…" shows `sum(${ref(local_test_table).y}) / 100` with the index preserved beside it as a separate control — nothing silently dropped.

**Phase 3 (VIS-1242)**
- All 14/14 model rows now carry drag handles.
- Dropping `local_test_table` on an empty slot lands `local_test_table ▸ choose a dimension` and auto-opens the editor. Apply commits `local_test_table ▸ x`.
- Loading an insight does **not** pop menus.

### Closing your "Not covered" note

The right-rail form *does* have a Save button, and the persisted string is correct. Edited `simple-scatter-insight`'s `y` to SUM + `/ 7` + `[0]` and hit Save:

```
POST /api/insights/simple-scatter-insight/  →  200
"y": "?{sum(${ref(another_local_test_table).new_y}) / 7}[0]"
```

Modifier inside the braces, index after — exactly the documented shape. Checked it against the Python pattern directly:

```
QUERY_STRING_VALUE_PATTERN.match("?{sum(${ref(...).new_y}) / 7}[0]")  →  True
```

Reloaded the page: the pill comes back as `SUM · another_local_test_table ▸ new_y / 7[0]`. Full round-trip through save + reload is clean.

---

### Findings

**1. The Phase 3 type guard is bypassed on empty slots — the silent no-op is still reachable**

This is the one I'd hold the PR for. The behaviour splits on whether the slot is already filled:

- Drop a **source** on a **filled** slot (a pill) → droppable resolves to `property-x`, `handleDropField`'s allowlist runs, toast: *"Can't use a source as a field value."* ✅
- Drop the same source on an **empty** slot → droppable resolves to `ref-text-:r3:` (dnd-kit's own announcement confirms it), so `WorkspaceDndContext.jsx:278` branch 2b handles it, returns `'ref_text_rejected'` — **no toast, no feedback, nothing.** ❌

The value is safe either way (no dangling ref is written), so blocker #2's *correctness* half is fixed. But the empty slot is the case a user hits first, and there the drop is still exactly the "no pill, no error, no animation" moment your own comment at `InsightBuildSection.jsx:200` calls "the single most trust-destroying moment the audit found." The `DROPPABLE_ON_PROPERTY` allowlist never runs on that path.

Suggested fix: branch 2b needs a rejection toast of its own — the allowlist can't be the only place that speaks.

**2. Dropping a model 400s the backend with a raw Python error**

The Phase 3 placeholder is POSTed straight to compile-draft before the user has configured anything:

```
POST /api/insight-compile-draft/
  {"props":{"type":"scatter","x":"?{${ref(local_test_table)}}"}}
  →  400  {"error":"'SqlModel' object has no attribute 'expression'"}
```

Nothing surfaces in the UI, but it's a console error on the headline Phase 3 path, so "Zero console errors throughout" doesn't hold once you drop a model. Two separate things worth fixing: the client shouldn't compile an unconfigured `modelRef` placeholder at all, and an `AttributeError` shouldn't be reaching the wire as an error string (CLAUDE.md's "clear, actionable error messages").

**3. "Save as metric…" now lies in the right rail — the same shape of bug Phase 1 just fixed**

On the AVG pill in `right-rail-edit-leaf-form`, the action is disabled under the hint *"Only an aggregate pill (SUM, AVG, …) can be saved as a metric."* It **is** an AVG pill — the header right above it says `Aggregate (AVG)`.

Cause: `saveAsMetricEnabled` (`PillMenu.jsx:273`) ands in `typeof onSaveAsMetric === 'function'`, and only `InsightBuildSection` wires that handler. So the disabled state is intentional, but the message blames the wrong thing.

The text predates this PR (#527), so this isn't a regression you introduced — but Phase 1 removing the `droppable` gate is what put pills into the right rail in the first place, which makes the wrong message *newly reachable* there. Worth splitting "not an aggregate" from "not available on this surface" while you're in here, since it's the same lying-affordance class the PR set out to kill.

**4. Nit — the static toggle explains itself only on hover.** The "says why" is a `title` attribute: no visible text, no `aria-describedby`. A user who clicks and gets nothing still gets nothing unless they hover and wait.

**5. Nit — dropped-model menu contradicts itself.** The pill reads `choose a dimension` while the Dimension select already displays `x` (first option preselected). Apply commits `x`, which is a fine default, but the label and the control disagree about whether a choice has been made.

### Not verified
- The `pillFieldSwap` exclusion for modified slots (the `= 'west'` data-loss case) — constructing it in-browser needs a metric/dimension whose name collides with the column. Resting on your unit tests.

### Unrelated, noticed in passing
`GET /api/model-jobs/?model_names=query_1&run_id=main` 404s when creating an exploration from a model row. Nothing in this diff touches it — flagging only because it's the other console error in the session.

🤖 Acceptance pass run with [Claude Code](https://claude.com/claude-code)

PR #629, opened 2026-08-25
Legible DuckDB lock errors and a quieter serve log (M11 + W9)

jpjsonic:
## What

Two small Error Legibility projects (dossier W2 + W9), independent of the Diagnostic contract PR (#628).

### M11 — the lock error becomes actionable
DuckDB's "Can't open a connection to same database file with a different configuration" (same-process config clash, or a DBeaver/duckdb-CLI hold) was wrapped into the generic *"Ensure the database exists and the connection properties are correct"* — advice that cannot be followed for a lock. `DuckdbSource.get_connection` now classifies lock conflicts (`duckdb.ConnectionException`, "different configuration", held-lock text) and raises a message naming the **source**, the **absolute path**, and the **way out** (close other connections / restart `visivo serve`), tagged `[source_locked]` so W3 can lift it into a `Diagnostic` without rewording.

Note: the biggest M11 *trigger* (the never-disposed SQLAlchemy engine in `create_file_database` on the onboarding path) is already removed by #622 — this PR handles the cases that survive every leak fix (external holders, mixed read/write config).

### W9 — the real error becomes findable
- `insight_jobs_views` "Insight file not found" / "Missing insight files" demoted info→debug (59 of 538 serve.log lines in field-test s1).
- `is_debug()` in `visivo/logger` is now the one definition of debug-on (`DEBUG == "true"`). The hot-reload server tested mere *presence* of `DEBUG`, so `DEBUG=false` turned the werkzeug/socketio access-log flood **on**; it and `socketio.run(debug=…, log_output=…)` now use the shared predicate.

Deliberately narrow per the dossier: the full Logger overhaul (levels/sinks/-v/-q) stays in the dormant Logging Improvements project.

## Test Strategy
- New: lock-conflict message test against a real held DuckDB file (names source/path/hint, never the generic advice); non-lock failures keep the generic message; `_is_lock_conflict` classification; `is_debug` for unset/false/"1"/true.
- Full suite: 2401 passed · black clean.

Findings: M11, M15/M16 (log-noise multiplier) — Error Legibility & Diagnostics


visivo/server/views/insight_jobs_views.py
```diff
@@ -26,7 +26,7 @@ def get_insights_api():
                 insight_file = os.path.join(output_dir, run_id, "insights", f"{name}.json")
 
                 if not os.path.exists(insight_file):
-                    Logger.instance().info(f"Insight file not found: {insight_file}")
```
tgsoverly, line 29: For this, I would like to know why it was so noisy, and fix this one.  Or at least understand why.

tgsoverly (changes requested):
I would like to know why the logging was getting called so much, seems like a fix in the front end is needed.

jpjsonic:

@tgsoverly — dug into the *why*, and you were right that the real fix is client-side (plus the server contract feeding it):

**The mechanism:** every chart/table surface polls `GET /api/insight-jobs/` for its insights. An insight that has *never been built* returned **404** — the same status a genuinely broken fetch would get — and the client treats any non-2xx as transient: `fetchInsightJobs` retries **3× at 1s** internally, then React Query (`retry: 1`) runs the whole thing **again**. So one render of a single not-yet-built insight = up to **6 polls and 6 log lines**, and every new object starts life not-yet-built. That's the flood (59 of 538 serve.log lines in the field test came from exactly this).

**The fix now in this PR** (last commit): the all-missing case returns `200 {insights: [], missing: [...], state: "not_built"}`; the client recognises the envelope and returns `[]` after exactly **one** fetch (unit-asserted). Built/partial responses keep the bare-array contract so cloud and older servers are untouched. The info→debug demotion stays for the residual legitimate polls (e.g. while a run is actually building).

Follow-up (Error Legibility W5, against the now-merged Diagnostic contract): distinguish *build-failed* (409 + diagnostics) from *never-built*, and fix the inverse bug in `modelJobs.js` which swallows failures as `[]`.

visivo/server/views/insight_jobs_views.py
```diff
@@ -63,11 +63,23 @@ def get_insights_api():
...
+                    # retrying 3x at 1s inside fetchInsightJobs, once more via
+                    # React Query — so ONE render of a not-yet-built insight
+                    # produced up to six polls and six log lines. The 200
+                    # envelope tells the client "nothing to fetch yet, stop".
                     return (
-                        jsonify({"error": f"No insight files found for: {missing_insights}"}),
-                        404,
+                        jsonify(
+                            {
+                                "insights": [],
```
tgsoverly, line 77: Hmmm, I still am not sure I like this.  I feel like you should get a 404 if it isn't there, but the front end should not request.  Can the front end know not to request it because it is a preview, or not even a valid insight?

We also need to do the same thing in Core for any api changes.

jpjsonic:

Rebrought up to date with `main` (was 12 behind; zero file overlap with anything that merged, so the merge is clean and CI now runs against current main).

@tgsoverly this is ready for your re-review whenever you have a minute — the root-cause frontend fix you asked for is in the diff (`viewer/src/api/insightJobs.js` + the `200 not_built` envelope in `insight_jobs_views.py`), not just the log demotion.

PR #643, opened 2026-08-28
Save-to-project: a rename commit no longer clobbers in-flight name edits (M14)

jpjsonic:
## Problem

`ExplorationPromoteModal` keeps the naming step's per-row draft text in a `nameDrafts` map, and re-seeded it with a wholesale `new Map(built…)` replacement in two places: on mount, and after every rename commit. A rename commit awaits a full checklist rebuild (a real ~1s network round-trip), so editing field B while field A's commit was in flight overwrote B's typed text with its post-rebuild store name. Because the inputs are controlled on this map, the caret stayed put when the value swapped mid-typing — later keystrokes spliced into the reverted name, producing the corrupted `${ref(telemetry_db_query_insighmy_insightt)}`-style refs seen in the field test (finding M14, Explorer Continuity dossier project 2).

## Fix

Both call sites now MERGE via a shared `mergeNameDrafts(prev, built)` helper instead of replacing:

- a still-present key keeps its existing draft — the user's in-flight text, which commits on its own blur, never here;
- only rows appearing under a new key (the just-committed row under its new name, or a genuinely new row) take the rebuilt store name;
- keys no longer in the checklist drop out.

No behavior change anywhere else: the just-committed row still re-syncs (its new key has no prior draft), selection remap and error clearing are untouched.

## Falsification evidence

With the tests kept and the implementation stashed (`git stash push -- …ExplorationPromoteModal.jsx`), both new tests fail against the pre-fix code:

```
✕ committing row A's rename never clobbers text already typed into row B (in-flight draft survives the rebuild) (57 ms)
✕ row B's preserved in-flight draft still commits on its own blur after A's rebuild (1028 ms)
    expect(element).toHaveValue(churn_by_cohort)
    Expected: "churn_by_cohort"   Received: "my_insight"
Tests:       2 failed, 99 skipped, 101 total
```

`git stash pop` restores the fix; all 101 pass.

## Test results

- `yarn test --watchAll=false --testPathPattern="ExplorationPromoteModal"`: 101 passed (2 new — the in-flight race is driven with a controlled rebuild promise, typing into row B while row A's rebuild hangs)
- Full suite `yarn test --watchAll=false`: **7014 passed, 365 suites**
- `yarn lint`: clean

### DIFF
--- a/viewer/src/components/views/workspace/ExplorationPromoteModal.jsx
+++ b/viewer/src/components/views/workspace/ExplorationPromoteModal.jsx
@@ -26,33 +26,11 @@ const RENAMABLE_TIERS = new Set(['model', 'insight', 'chart']);
 const isRenamableRow = row => row.status === 'new' && RENAMABLE_TIERS.has(row.tier);
 
 /**
- * Re-seed the name-draft map from a (re)built checklist WITHOUT clobbering
- * in-flight edits (M14 — "Stop the name-draft map clobbering in-flight
- * edits"). Committing row A's rename awaits a full checklist rebuild (a real
- * network round-trip), and the user may already be typing into row B's field
- * when it lands. A wholesale `new Map(built…)` replacement here reverted B's
- * typed text to its store name mid-keystroke — and, because these inputs are
- * controlled on this map, left the caret where it was, splicing subsequent
- * keystrokes into the reverted name (the corrupted
- * `${ref(…_insighmy_insightt)}`-style refs from the field test).
- *
- * So: MERGE. A still-present key keeps its existing draft (the user's text,
- * committed on ITS OWN blur, never here); only rows appearing under a new
- * key — the just-committed row under its new name, or a genuinely new row —
- * take the rebuilt store name. Keys no longer in the checklist drop out
- * (a name the user renamed AWAY from must not leave a stale draft behind to
- * resurface if they later rename back TO it).
- *
- * `committedKey` — the key `handleNameCommit` just committed — is the ONE
- * key that always re-seeds from `built`, even when it survives the rebuild
- * unchanged. Its draft is no longer "in-flight": the user already blurred
- * it. Inferring that re-sync from key CHURN alone (the row reappearing under
- * a new name) was wrong whenever the rename didn't move the key — the store
- * rename actions (`renameModelTab`/`renameInsight`) silently `return` for a
- * draft with `isNew === false`, while renamability here is decided by the
- * BACKEND diff's `'new'`. Those two diverge for a working copy whose project
- * object left the project mid-session (Library delete, external .visivo.yml
- * edit), and the field then displayed a name the save would never use.
+ * Merges a rebuilt checklist into the name-draft map rather than replacing it:
+ * these name inputs are controlled on this map and the rebuild is an async
+ * round-trip, so a replace reverts text the user is typing into another row.
+ * `committedKey` alone always re-seeds from `built` — its edit is already
+ * blurred, and a rename the store silently refused leaves its key unmoved.
  */
 const mergeNameDrafts = (prev, built, committedKey = null) => {
   const next = new Map();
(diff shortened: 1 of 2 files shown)


jpjsonic:

## Review fix pass — 4 findings, all 4 real, all 4 fixed (`6fa4d34e`)

I verified each finding against the PR head before acting; none turned out to be a non-issue.

### 1 + 3 (same defect, lines 49 / 268) — the just-committed row's re-sync rested on its key changing

**Confirmed real.** The re-sync of the row `handleNameCommit` just committed was inferred from key *churn* ("its NEW key has no prior draft"). That inference breaks whenever the rename doesn't move the key, and the reviewer's path to that is genuinely reachable — I checked both ends:

- `explorerStore.js` — `renameModelTab` (:648) and `renameInsight` (:942) both open with `if (!modelState || !modelState.isNew) return;`. A **silent no-op, not a throw**, so `handleNameCommit` treats it as success.
- `explorer_views.py::_diff_object` (:115) — `existing = manager.get(name); if not existing: return "new"`.
- `promoteChecklist.js` `statusFor` (:214) — `raw === 'new'` wins outright over the local `isNew` fallback.

So a working copy loaded from a project object (`isNew === false`) whose object then leaves the project mid-session — Library delete, external `.visivo.yml` edit, the cases `ExternalEditBanner` / `ExplorationDeletedRemotelyBanner` exist for — reads as status `'new'` and renders an editable name field. The user renames it, the store refuses silently, the rebuild returns the row under the **same** key, and the merge preserved the typed draft. The field then read `daily_orders` while the checkbox, the verdict pill, the row test id and the object `promoteExploration` actually saves were all still `orders_q` — exactly the "project polluted with `query_1`" outcome the D11 naming step exists to prevent.

**Fix:** `mergeNameDrafts(prev, built, committedKey)`. That one key always re-seeds from the rebuilt checklist, even when it survives unchanged; in-flight preservation is untouched for every other row, which is the whole point of M14.

**Plus** — the snap-back alone was only ever "a crude but honest signal", so a refused rename now says why instead of mutely reverting the user's typing. A clean `renameRow` call isn't proof the rename happened, so the commit checks the rebuilt checklist for the new name and, when the row is still under its old one, sets the row's existing inline `-name-error`: *"Could not rename — `orders_q` was loaded from an existing object, so it will be saved under that name."* The object still saves, and the user can see under which name.

### 2 (line 45) — the drop clause was asserted by nothing

**Confirmed real** — a testability gap, not a bug; the shipped code is correct. I reproduced the mutant (`const next = new Map(prev)`) passing all 101 tests. New guard test: rename `my_insight` → `temp_name` → back to `my_insight`, and assert the field reads `my_insight` (and that no phantom third rename fires from the resurrected draft).

### 4 (test file, line 22) — `clearAllMocks` doesn't drain `once` queues

**Confirmed real, and reproduced exactly as described.** With only the two merge call sites reverted:

```
✕ committing row A's rename never clobbers text already typed into row B   (10 ms)
✕ row B's preserved in-flight draft still commits on its own blur           (1015 ms)
✕ committing a rename on an INSIGHT row calls renameInsight                 (1014 ms)
✕ committing a rename on a CHART row calls setChartName                     (1004 ms)
✕ a collision during auto-suggestion (initial load) fails open              (1003 ms)
✕ clearing a name to blank on blur is a no-op                               (1004 ms)
Tests:       6 failed, 95 passed, 101 total
```

Four of those six point at features that are fine. The M14 tests are the first here whose queue consumption is *contingent on the behavior under test*, so one real regression sprayed timeouts across unrelated tests. `buildPromoteChecklist.mockReset()` in the `beforeEach` drains both the `once` queue and any lingering `mockResolvedValue`; no existing test depended on the leakage (104/104 green with it in place).

### Falsification

Each mutation applied alone, everything else intact:

| mutation | result |
| --- | --- |
| drop `committedKey` at the call site | **1 failed**, 103 passed — only *"a rename the store silently REFUSES snaps the field back"* |
| drop the refusal-error block | **1 failed**, 103 passed — only *"a silently-refused rename explains itself inline"* |
| `new Map()` → `new Map(prev)` | **1 failed**, 103 passed — only *"renaming a row back to a name it previously had…"* |
| revert both merge call sites to the pre-M14 wholesale replace, **with** `mockReset` | **2 failed**, 102 passed — exactly the two M14 tests, no cascade (was 6 failed / 95 passed) |

### Gates

Full viewer suite `7017 passed, 365 suites`; `yarn lint` clean. No Python touched.

tgsoverly:
Can we not allow you to edit the name of existing object in the exlorer rather than throw and error?  If I read this correctly.

tgsoverly (approved):
Good without comments... :-)

jpjsonic:

Claude trimmed this PR's comments to the new bar in `visivo/CLAUDE.md` (#668). 14 comments removed, 8 kept and tightened; net -86 lines, no behavior change.

Deleted: the M14 bug-history narration (the `…_insighmy_insightt` corruption story, "the pre-fix wholesale replacement…", "this used to be a bare fire-and-forget"), the reviewer-facing justifications above each new test, and the step-by-step narration inside them. Those test names already pin what the prose was arguing.

Kept (tightened to 1–5 lines each):
- `mergeNameDrafts` docstring — the name inputs are controlled on this map and the rebuild is an async round-trip, so a replace reverts text the user is typing into another row.
- `handleNameCommit` — `renameModelTab`/`renameInsight` bail out silently (bare `return`) for a draft with `isNew === false`, while renamability here comes from the backend diff's `status === 'new'`, so a clean `renameRow` call is no proof the rename happened.
- test `beforeEach` — `clearAllMocks` drains neither the `mockResolvedValueOnce` queue nor a lingering `mockResolvedValue`, hence the explicit `mockReset`.
- the `toBeEnabled()` barrier marking when the rebuild has landed (`renamingKey` is cleared last).

Scope was this PR's own comments plus pre-existing ones inside `handleNameCommit`, which this PR already rewrites. The rest of `ExplorationPromoteModal.jsx` carries heavy pre-existing comment debt (long ux-audit/VIS-number rationale blocks above `promotedField`, `showFallbackDashboardOffer`, `handlePromote`) — left untouched to avoid conflicts across the other open PRs.

Gates: viewer only, no Python touched. `yarn test --watchAll=false` — 365 suites / 7017 tests passed. `yarn lint` — 0 errors. Branch merges cleanly with `main`; no conflict resolution was needed.

PR #645, opened 2026-08-28
Workspace source save externalizes credentials instead of masking them into YAML (B1/M9)

jpjsonic:
## Problem (P0 blocker — Data On-Ramp P3)

Saving a source from the Workspace destroys its credential:

1. The sources API (`_serialize_object`) echoes stored secrets as Pydantic's `**********` mask.
2. The frontend round-trips that config on save; `SourceManager.save_from_config` cached `SecretStr("**********")` verbatim.
3. The commit path (`_build_child_info` → `model_dump(mode="json")` → ProjectWriter) wrote the mask **literally** into `project.visivo.yml` — total silent credential loss.

A newly typed password had the opposite failure: it landed in the YAML in **plaintext**. `externalize_source_credentials` ran only on the 2023 onboarding path (`/api/source/create/`, `project_views.py`), never on the SourceManager path — the blocking prerequisite for retiring the onboarding endpoints.

## Fix

`SourceManager.save_from_config` now runs `_protect_secrets` before caching. It covers **every** SecretStr-typed field, detected from the live model rather than a hardcoded list (`password` on all server sources, Snowflake `private_key_passphrase`, BigQuery `credentials_base64`, and any future ones — fields already holding `${env.*}` validate to `EnvVarString` and are naturally exempt):

- **Mask preservation** — an incoming value equal to the mask (derived from Pydantic itself: `SECRET_MASK = str(SecretStr(...))`, guarded by a sentinel test against the exact API echo) preserves the stored value: an existing `${env.*}` ref is kept verbatim; an existing literal keeps its real value; a mask with nothing stored behind it is dropped, never written.
- **Externalisation** — any plaintext secret that a commit would re-serialize into YAML moves to `<working_dir>/.env` (merged, never clobbered; also loaded into `os.environ` so refs resolve in the running server) and the field becomes `${env.<SOURCE>_<FIELD>}` — the same naming convention the onboarding path uses. `<working_dir>/.env` matches what `commit_views` reads and the CLI loads.
- **No YAML churn on no-op saves** — a masked secret with nothing else changed keeps the published literal byte-identical (status stays `published`, commit skips it).
- Empty secrets are dropped (mirrors onboarding).

The shared helpers (`env_var_name`, `merge_env_file`, `externalize_source_credentials`, `SECRET_MASK`) moved from `views/utils.py` to a new `visivo/server/source_credentials.py` so the manager doesn't import a views module. `views/utils.py` re-exports them (including the old `_env_var_name` name), keeping the onboarding path and its tests byte-identical — shared helper, not duplication.

Drive-by unbreak: bot commit 982d28c1 bumped the version to 2.1.1 without regenerating `visivo_project_schema.json`, so `test_committed_schema_is_fresh` failed on a clean checkout of main (committed `cli_version` default was still 2.1.0). Regenerated via `write_schema_json` — the only schema change is that default.

## Falsification

With only the guard call removed (`source = self._protect_secrets(source)`), 13 of the 17 new tests fail with the exact bug signature; the 4 that still pass are mask-sentinel/no-secret invariant tests that don't depend on the guard:

```
FAILED ...::TestMaskPreservation::test_masked_password_preserves_existing_literal_value - AssertionError: assert '**********' == '${env.WAREHOUSE_PASSWORD}'
FAILED ...::TestMaskPreservation::test_masked_password_on_noop_save_keeps_literal_untouched - AssertionError: assert '**********' == 'hunter2'
FAILED ...::TestMaskPreservation::test_workspace_roundtrip_of_api_config_destroys_nothing - AssertionError: assert '**********' == '${env.WAREHOUSE_PASSWORD}'
FAILED ...::TestNewPlaintextSecretExternalized::test_snowflake_password_and_private_key_passphrase_both_externalized - AssertionError: assert '**********' == '${env.SNOW_PASSWORD}'
FAILED ...::TestEmptySecrets::test_empty_password_is_dropped - AssertionError: assert SecretStr('') is None
13 failed, 4 passed in 0.40s
```

## Tests

- `tests/server/managers/test_source_manager_credentials.py` — 17 new tests (factory-boy sources): mask preserves env-var ref / literal / prior draft; new + changed plaintext externalized to `.env` with refs in the commit serialization; Snowflake passphrase + BigQuery credentials covered; non-secret fields update normally; `.env` merge preserves unrelated entries; full API-roundtrip regression test of the audit scenario.
- Full suite: `pytest tests/ -q` → **2460 passed, 2 skipped, 5 xfailed, 1 xpassed** (includes the previously failing `test_committed_schema_is_fresh`).
- `black --check --target-version py312 .` → clean (543 files unchanged).
- Onboarding path: `tests/server/views/test_utils.py` untouched and green through the re-exports.


jpjsonic:

## Review fix pass — 9615a025

Verified every finding against the worktree before acting. **10 fixed, 1 rejected, 1 partially accepted.**

### Fixed

**`.env` values are quoted and escaped** (`source_credentials.py:70`/`:72` — critical, both reports)
Confirmed against the installed python-dotenv 1.2.3: `Passw0rd #1` → `Passw0rd`, `hunter2 ` → `hunter2`, `"quoted"` → `quoted`, `p${a}ss` → `pss`, and a multi-line value (what `base64` produces for a BigQuery `credentials_base64` blob, wrapped at 76 columns) truncates to its first line and corrupts the rest of the file. Since the YAML now holds only a `${env.*}` reference, that .env line is the sole remaining copy.
Values are now double-quoted with `\`, `"`, `\n`, `\r` escaped and `${` encoded as `${:-$}{`. Single quotes were rejected: dotenv's single-quote escape set is only `\\`/`\'`, so an apostrophe in a password is unrepresentable, and a literal newline inside single quotes would survive parsing but break the next line-based merge. `${` needs the encoding either way — dotenv interpolates POSIX references inside quoted values and offers no escape; `${:-$}` is a reference with an empty name and a `$` default, so it resolves back to a literal `$`. 28 values now round-trip exactly.

**`env_var_name` always produces a referenceable name** (`source_manager.py:179` — critical)
Reproduced: source `2024_warehouse` yielded `EnvVarString('${env.2024_WAREHOUSE_PASSWORD}')` and `get_password()` returned that literal string. Names are now ASCII-folded (`café` → `CAF_PASSWORD`, not `CAFÉ_PASSWORD`) and `_`-prefixed when they'd start with a digit, and the reference is built through `TypeAdapter(EnvVarString)` rather than constructed past pydantic.

**Colliding source names no longer clobber each other** (`source_credentials.py:41`/`:45` — minor, both reports)
`analytics-db` and `analytics_db` both wanted `ANALYTICS_DB_PASSWORD`. A key already holding a *different* value that this source doesn't already reference is now skipped for the next free one (`..._2`), with a warning. Ownership comes from the `${env.*}` names the cached/published versions of the same source already point at, so rotating a password still reuses the source's own key rather than leaking `_2`, `_3`, ….

**A stored secret whose value IS the mask is dropped, not externalized** (`source_manager.py:178` — major)
Reproduced: `.env` got `WAREHOUSE_PASSWORD=**********`. Now dropped with a warn telling the user to re-enter, so the YAML keeps showing the symptom instead of a plausible-looking reference. A defensive guard at the write itself keeps "the mask is never written" true.

**Deletion tombstone no longer drops a live credential** (`source_manager.py:133` — minor)
Reproduced (password `None` after re-save while `_published_objects` still held `SecretStr('hunter2')`). `existing` now falls back to `_published_objects`.

**`objects_equal` ignores `path`/`file_path`** (`source_manager.py:171` + `test_…:191` — minor + major)
Confirmed both halves: `Project(**{...})` stamps `path='project.sources[0]'` on every source, `_serialize_object` strips `path`/`file_path` from the API config, so the no-op guard was dead in every real project — a byte-identical masked save went MODIFIED, wrote `.env`, and rewrote the password to a reference. Fixed in `ObjectManager.objects_equal` (matching `_serialize_object`), and the guard is now certified against a source loaded through `manager.load(project.dag())`, plus a `file_path` variant for the includes case.

**`.env` is gitignored before the first write** (`source_manager.py:181` — minor, both reports)
Mirrors the pairing `views/utils.write_project_file` already performs. Appends to an existing `.gitignore` (handling a missing trailing newline), creates one when absent, idempotent.

**`-w/--working-dir` loads that directory's `.env`** (`source_manager.py:181` — minor)
The `visivo` group loads `--env-file` relative to the process cwd before any subcommand option is parsed, so `visivo serve -w ./analytics` never read `analytics/.env` — where the Workspace writes the secret and where `commit_views._env_file_keys` reads names from. Loaded in the `working_dir` callback with `override=False`, so cwd/`--env-file` still wins.

### Partially accepted

**Rename shape** (`source_manager.py:156` — minor). The silent drop is the right behaviour — the mask must never be stored, and raising would 400 the legitimate "config copied from another source's masked response" case that `test_masked_secret_with_no_stored_value_is_dropped` covers. But the reviewer is right that it was invisible: the drop is now logged at `warn` (not `info`) and the rename shape has explicit test coverage asserting the drop *and* the warning.

### Rejected

**"Externalisation belongs on the commit path"** (`source_manager.py:181` — major). The mechanism is real: a draft save writes `.env` and mutates `os.environ`, and a discarded draft leaves an orphan key behind. But this is a design tradeoff, not data loss, and the alternative is worse. Deferring the write means the credential exists only in memory until commit — lost on restart, while `os.environ` still has to be mutated for Test Connection to work on the draft. The commit path (`commit_views.py`) is another open PR's surface and out of bounds here. What's left after a discard is an unused key in a file that is now guaranteed to be gitignored. Documented in the commit message.

### Falsification

Stashed the four implementation files, kept the tests: **26 of the new guards fail without the fix**, none for import reasons.

```
FAILED …TestEnvFileEncoding::test_awkward_password_round_trips_through_dotenv['Passw0rd #1'] - assert 'Passw0rd' == 'Passw0rd #1'
FAILED …['hunter2 '] - assert 'hunter2' == 'hunter2 '
FAILED …[' hunter2'] - assert 'hunter2' == ' hunter2'
FAILED …['"quoted"'] - assert 'quoted' == '"quoted"'
FAILED …["'quoted'"] - assert 'quoted' == "'quoted'"
FAILED …['p${a}ss'] - assert 'pss' == 'p${a}ss'
FAILED …['${HOME}'] - assert '/Users/jaredjesionek' == '${HOME}'
FAILED …::test_multi_line_credential_stays_a_single_env_entry - assert 'eyJ0eXBlIjog…fYWNjb3VudCIs' == 'eyJ0eXBlIjog…UJFR0lOIn0=\n'
FAILED …::test_awkward_secret_survives_a_later_unrelated_merge - assert 'Passw0rd' == 'Passw0rd #1'
FAILED …TestEnvVarNameIsResolvable::test_source_name_starting_with_a_digit_still_resolves - assert [] != []
FAILED …::test_non_ascii_source_name_still_resolves - assert '${env.CAFÉ_PASSWORD}' == 'hunter2'
FAILED …::test_stored_reference_survives_a_yaml_round_trip - assert False
FAILED …TestEnvVarNameCollisions::test_similar_source_names_do_not_overwrite_each_others_credentials - assert 'passB' == 'passA'
FAILED …TestStoredMaskLiteralIsNotACredential::test_published_mask_literal_is_dropped_not_externalized - assert EnvVarString('${env.WAREHOUSE_PASSWORD}') is None
FAILED …TestDeletedSourceCredential::test_masked_save_after_delete_recovers_the_published_credential - assert None is not None
FAILED …::test_masked_edit_after_delete_externalizes_the_published_credential - assert 'None' == '${env.WAREHOUSE_PASSWORD}'
FAILED …TestRenameShape::test_masked_save_under_an_unknown_name_drops_the_secret_loudly - assert False
FAILED …TestGitignore::test_env_is_added_to_an_existing_gitignore_before_the_first_write - assert '.env' in ['target', '.visivo_cache']
FAILED …::test_gitignore_is_created_when_absent - FileNotFoundError: …/.gitignore
FAILED …::test_gitignore_without_trailing_newline_is_not_corrupted - assert ['target'] == ['target', '.env']
FAILED …TestNoopSaveOfADagLoadedSource::test_masked_noop_save_of_a_dag_loaded_source_writes_nothing - assert False
FAILED …::test_masked_noop_save_with_a_file_path_writes_nothing - AttributeError: 'EnvVarString' object has no attribute 'get_secret_value'
FAILED tests/server/managers/test_object_manager.py::…::test_objects_differing_only_by_path_are_equal - assert False
FAILED …::test_objects_differing_only_by_file_path_are_equal - assert False
FAILED …::test_status_of_a_dag_loaded_object_resaved_unchanged_is_published - assert MODIFIED == PUBLISHED
FAILED tests/commands/test_options.py::test_working_dir_loads_that_directory_env_file - assert None == 'hunter2'
26 failed, 65 passed
```
(`tests/server/views/test_utils.py` fails to collect without the fix — it imports `ensure_env_gitignored`.)

### Gates

- `.venv/bin/pytest tests/ -q` → **2525 passed**, 2 skipped, 5 xfailed, 1 xpassed
- `.venv/bin/black --check --target-version py312 .` → **544 files unchanged**
- Coverage on the touched files: `source_credentials.py` 99%, `source_manager.py` 94%, `object_manager.py` 94% (the remaining misses are pre-existing connection/validation error paths)
- No Pydantic model changed, so no schema regeneration. No viewer changes.

visivo/server/source_credentials.py
```diff
@@ -0,0 +1,207 @@
...
+# on save. Derived from Pydantic itself rather than hardcoded so a masking
+# change upstream breaks tests instead of silently destroying credentials.
+SECRET_MASK = str(SecretStr("mask-probe"))
+
+
+# A legal POSIX-style environment variable name. Deliberately the same shape
+# as the name group of ENV_VAR_CONTEXT_PATTERN: a name this rejects cannot be
+# referenced as ${env.NAME} at all, so the reference would resolve to nothing
+# and the source would authenticate with the literal "${env....}" string.
+ENV_VAR_NAME_PATTERN = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
```
tgsoverly, line 48: Is this duplicated?

visivo/server/source_credentials.py
```diff
@@ -0,0 +1,207 @@
...
+    path = env_file_path(project_dir)
+    try:
+        if not path.exists():
+            return {}
+        return {key: value for key, value in dotenv_values(str(path)).items() if value is not None}
+    except OSError:  # pragma: no cover - defensive
+        return {}
+
+
+def quote_env_value(value: str) -> str:
```
tgsoverly, line 96: Wonder if there is a library for this.

tgsoverly (approved):

PR #646, opened 2026-08-28
Click-to-pick: every empty slot opens the ReferencePicker (B2)

jpjsonic:
## Problem

The ReferencePicker (VIS-792 / Track L L-2 — search, type-scoped list, click-to-select, create-new CTA) was fully built but reachable from exactly ONE place: BrokenRefCard's **Fix…** flow, a failure state. Every healthy empty slot on the canvas was an inert "Empty slot" div, and the right rail's empty item panel offered only a drop zone — filling a slot required a drag, full stop (W5, the last B2 sub-issue).

## Fix

**1. Empty canvas slots are clickable.** `Dashboard.jsx`'s canvas-mode empty-slot placeholder becomes a real `<button>` when the canvas supplies the new `onEmptySlotClick` prop (View mode + prop-less canvas render are byte-identical to before). Clicking — or focusing and pressing Enter — opens the picker for that slot.

**2. Right-rail "Choose…".** An empty item's edit panel (`ItemEditForm`, used standalone and inside Row/Dashboard forms) gains a **Choose…** button that opens the same picker. State lives in a child component mounted only for empty leaves, so `ItemEditForm` stays hook-free.

**3. The picker lists CHARTS and INSIGHTS.** `ReferencePicker` gains a multi-`types` mode: typed sections with objectTypeConfigs-coloured headers/icons, cross-section search, a shared empty state. `onSelect` now reports `(name, type)`; the single-`type` BrokenRefCard contract is unchanged.

**4. Insight picks auto-wrap (#637 pattern).** Picking a chart places `chart: ref(name)`. Picking an insight mints a unique `<insight>-chart` name (`generateUniqueName(..., {separator: '-'})`, -2 disambiguation), saves the wrapper via the same `saveChart` path as Library "Wrap in Chart…" (#632), and places the wrapper — never a bare insight. Ordering matches #637: the placement transform is validated FIRST; a rejected placement mints nothing, so no orphan draft charts. The #632/#637/picker recipe now lives in one shared module (`utils/insightWrap.js`) — `WorkspaceDndContext`'s drop path consumes it too, so the three surfaces can't drift.

**5. #617 semantics respected.** Placements land in the dashboard working copy (`commitCanvasConfig` on the canvas, `onChange` in the rail); the explicit Save footer persists them.

**6. Telemetry.** `canvas_action { kind: 'add_item', source: 'picker', type, name, wrapped_chart? }` matching #637's event shape; picker create-new emits `inline_create_used` with `source: 'picker'`.

## Falsification

Implementation stashed (tests kept): **20 tests fail**, all of them the new coverage — Dashboard's clickable slot, ItemEditForm's Choose… flow (incl. -2 collision + close-mints-nothing), ProjectCanvas's pick/wrap/rejected-placement/telemetry, ReferencePicker's multi-type mode. Sample:

```
● Dashboard › empty-slot placeholder (VIS-901 #3) › renders the empty slot as a real button when onEmptySlotClick is provided
  Expected: "BUTTON"
  Received: "DIV"
    > 308 |       expect(slot.tagName).toBe('BUTTON');
```

```
Test Suites: 4 failed, 1 passed, 5 total
Tests:       20 failed, 109 passed, 129 total
```

Stash popped; all green again.

## Test results

- Full viewer suite: `yarn test --watchAll=false` → **366 suites, 7038 tests, all passing**
- `yarn lint` → clean (exit 0)
- New Playwright story (written, not run — needs a sandbox): `viewer/e2e/stories/canvas-click-to-pick.spec.mjs` — slot click → picker sections → chart pick → working-copy assert; insight pick → wrapper minted + placed; right-rail Choose…; keyboard-only completion; console-error sweep.

New unit coverage: `utils/insightWrap.test.js` (mint recipe incl. -2/-3 and forced hyphen), `ReferencePicker.test.jsx` (+6 multi-type), `Dashboard.test.jsx` (+2 clickable/inert slot), `ProjectCanvas.test.jsx` (+7 pick wiring incl. rejected-placement-mints-nothing), `ItemEditForm.test.jsx` (+6 Choose… flow incl. -2 collision).

### DIFF
--- a/viewer/e2e/stories/canvas-click-to-pick.spec.mjs
+++ b/viewer/e2e/stories/canvas-click-to-pick.spec.mjs
@@ -17,6 +17,20 @@
  * path). The story never presses the dashboard Save, so the layout is not
  * persisted; the draft wrapper chart is uniquely named per run.
  *
+ * Because that draft SURVIVES the run, the auto-wrap assertion must never be
+ * a prefix probe: it reads the wrapper's name back out of the slot and asserts
+ * THAT EXACT chart exists exactly once, so a leftover `<insight>-chart` from an
+ * earlier run (or an earlier retry attempt — retries: 2) cannot satisfy it.
+ *
+ * ISOLATION: this file is deliberately NOT registered in the 'state-mutating'
+ * project despite drafting a chart into the backend cache. That project exists
+ * to serialise specs that share the DEFAULT :3001 sandbox; this story, like
+ * every other dedicated-port canvas story (canvas-keyboard-nav,
+ * canvas-wrap-container, canvas-nested-dnd, broken-ref-fix…), runs against its
+ * own sandbox via VIS_PICK_BASE and is environmentally unrunnable on :3001.
+ * Listing it there would be misleading AND (via that project's
+ * `dependencies: ['parallel']`) would mean it never executes at all.
+ *
  * Precondition: an isolated sandbox running the integration project.
  *   VISIVO_SANDBOX_BACKEND_PORT=8050 VISIVO_SANDBOX_FRONTEND_PORT=3050 \
  *   VISIVO_SANDBOX_NAME=pick bash scripts/sandbox.sh start
@@ -61,6 +75,14 @@ const appendEmptySlotRow = async page =>
     },
     { dashboard: DASHBOARD }
   );
+
+// `${ref(name)}` / `ref(name)` → `name`.
+const parseRefName = value => {
+  const match = String(value || '').match(/^\$?\{?\s*ref\(\s*([^)]+?)\s*\)\s*\}?$/);
+  return match ? match[1] : String(value || '');
+};
+
+const escapeRe = value => value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
 
 // Read the item config at a composite canvas path from the live working copy.
 const readItemAtPath = async (page, itemPath) =>
@@ -184,21 +206,35 @@ test.describe('Canvas click-to-pick (W5 / Track L)', () => {
     const item = await readItemAtPath(page, itemPath);
     expect(item.insight).toBeUndefined();
 
-    // The wrapper chart exists as a draft in the chart store, holding exactly
-    // the wrapped insight (same shape as Library "Wrap in Chart…", #632).
+    // Read the wrapper's name back OUT of the slot and assert THAT EXACT chart
+    // exists. A `startsWith(<insight>-chart)` probe is satisfied by a draft
+    // this story left behind on an earlier run (or on an earlier retry
+    // attempt — playwright.config.mjs sets retries: 2), so the auto-wrap half
+    // of the #637 check would silently stop biting: a collision just bumps the
+    // NEW mint to `-2` while the stale `-chart` keeps the assertion green.
+    // Naming the exact chart the slot points at is the only honest check.
+    const placedName = parseRefName(item.chart);
+    expect(placedName).toMatch(new RegExp(`^${escapeRe(insightName)}-chart(-\\d+)?$`));
+
+    // That chart exists as a draft holding exactly the wrapped insight (same
+    // shape as Library "Wrap in Chart…", #632) — and exists exactly once.
     await expect
       .poll(
         async () =>
           page.evaluate(
-            name =>
-              (window.useStore.getState().charts || []).some(c =>
-                c.name.startsWith(`${name}-chart`)
-              ),
-            insightName
+            name => {
+              const matches = (window.useStore.getState().charts || []).filter(
+                c => c.name === name
+              );
+              if (matches.length !== 1) return { count: matches.length };
+              const cfg = matches[0].config || matches[0];
+              return { count: 1, insights: cfg.insights };
+            },
+            placedName
           ),
         { timeout: WAIT }
       )
-      .toBe(true);
+      .toEqual({ count: 1, insights: [`ref(${insightName})`] });
 
     await page.screenshot({ path: `${SCREENS}/pick-03-insight-wrapped.png`, fullPage: true });
   });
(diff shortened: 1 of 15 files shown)


jpjsonic:

## Review fix pass — d632ae99

Verified each finding against the code before acting. **6 fixed, 2 rejected with reasoning.**

---

### Fixed

**1 · VIS-1025 read-only bypass in the rail (major)** — `ItemEditForm.jsx`

Confirmed: `RightRailEditPanel.persistConfig` returns early on `readOnly` (line 228), and the item-layout branch (lines 642–652) is *not* inside `right-rail-readonly-fieldset` — so `Choose…` was fully live, and `state.saveChart(...)` fired a real POST *before* `onChange` reached the hold. Net result was exactly the orphan draft chart #637 forbids, on a project the user isn't allowed to edit.

Fixed by disabling the trigger under read-only, plus a handler-level hold for a stage that flips while the picker is open. The predicate moved to **`branchingStore.isReadOnly`** (next to the `capabilities` state it reads); `RightRailEditPanel` and `useRecordSave` now import it instead of each re-deriving `capabilities.can_edit === false`.

**2 · The `#637` guard was unreachable (major)** — `ItemEditForm.jsx`

Confirmed: `setItemLeaf` writes `next[type]` for any non-blank name, so `!next.chart` could never be true. Replaced the truthiness check with `placeLeafRef`, which reads the placed ref **back** and requires it to equal the name about to be saved. That turns dead code into a live invariant: `setItemLeaf` *trims*, so an insight named `" rev"` used to mint `" rev-chart"` while the slot pointed at `rev-chart` — an orphan draft **and** a broken ref. Now it mints nothing.

**3 · Rail picker had no `onCreateNew`** — the empty-project modal was a dead end whose own copy promised an action it couldn't perform. Now passes one and emits `inline_create_used {source:'picker'}`, matching `ProjectCanvas`.

**4 · Empty-slot accessible name (WCAG 2.5.3)** — `Dashboard.jsx`. `aria-label` no longer replaces the visible label; it's the visible label plus the slot's location (`"Empty slot — click to choose (row 1, item 3)"`), built from the shared `EMPTY_SLOT_LABEL` constant and a new pure `describeSlotPath`. Four KPI slots no longer announce identically.

**5 · ReferencePicker focus management** — `aria-modal="true"` was asserting containment the DOM didn't enforce. Added a Tab/Shift+Tab focus trap and return-focus-to-opener on unmount (skipped when the opener is gone, e.g. a filled slot).

**6 · e2e auto-wrap assertion could pass on a stale chart** — the story leaves its draft behind and `retries: 2` is set, so `startsWith('<insight>-chart')` was satisfiable by an earlier run. It now reads the wrapper name back out of the slot and asserts *that exact* chart exists exactly once holding `ref(<insight>)`. Also added an Escape → focus-returns-to-slot step to the keyboard test.

---

### Rejected

**Registering the story in `state-mutating`.** Every dedicated-port canvas story (`canvas-keyboard-nav`, `canvas-wrap-container`, `canvas-nested-dnd`, `broken-ref-fix`) is unregistered; that project exists to serialise the shared **:3001** sandbox, and this spec's `BASE` is `VIS_PICK_BASE || :3050` — environmentally unrunnable on :3001. Worse, `state-mutating` carries `dependencies: ['parallel']`, which per the config's own comment means the file would never execute. Reasoning is recorded in the spec header rather than left implicit.

**Deleting the inert `{ separator: '-' }`.** The claim is correct — the base always contains a hyphen, so `suffixSeparatorFor` infers `-` for every reachable input — but the option is deliberate defence against a future change to that inference. Rather than drop it, the test that *claimed* to cover it now actually falsifies it (spy on `generateUniqueName`, asserting the option is passed), and the misleading title was fixed.

---

### Falsification

Impl stashed, tests kept — **14 failures**, all in the new guards:

```
● ItemEditForm › an insight whose wrapper name the transform rewrites mints NOTHING
    expect(jest.fn()).not.toHaveBeenCalled()
    Received number of calls: 1
    1: " rev-chart", {"insights": ["ref( rev)"]}

● ItemEditForm › read-only stage › the Choose… trigger is disabled and cannot open the picker
    Received element is not disabled:
      <button class="…" data-testid="item-row-0-item-0-choose" type="button" />

● ItemEditForm › read-only stage › a stage that flips read-only while the picker is OPEN holds the pick
    Received number of calls: 1
    1: "rev-insight-chart", {"insights": ["ref(rev-insight)"]}

● ItemEditForm › a chart whose name the transform drops places NOTHING
    Received number of calls: 1
    1: {"chart": "${ref(orphan)}", "width": 1}

● ReferencePicker › returns focus to the opener when the picker closes
    Expected element with focus: <button data-testid="opener">
    Received element with focus: <body>…</body>

● ReferencePicker › Tab from the last focusable wraps to the first (focus trap)
● ReferencePicker › Shift+Tab from the first focusable wraps to the last
● ReferencePicker › Tab from outside the dialog is pulled back in

● Dashboard › gives the slot an accessible name containing its visible label
    Expected element to have text content: undefined
● Dashboard › names each slot by its own location
● Dashboard › describeSlotPath (×2)
● ItemEditForm › an EMPTY project offers create-new from the picker
● ItemEditForm › the picker footer create link is available whenever objects exist

Tests: 14 failed, 107 passed
```

The separator spy was falsified by targeted mutation (deleting `, { separator: '-' }` from `insightWrap.js:29`):

```
● mintWrapperChartName › FORCES the hyphen separator rather than letting the base infer it
      "rev_insight-chart",
      ["rev_insight-chart"],
    - {"separator": "-"},
```

### Gates

- `yarn test --watchAll=false` → **366 suites, 7056 tests, all pass**
- `yarn lint` → **0 errors**
- `playwright test --list canvas-click-to-pick` → 6 tests discovered, spec parses

The e2e story itself was **not executed** — it needs its own sandbox on :3050 and this machine is at 99% disk. The changes to it are assertion tightening plus one added Escape/focus step; worth a run before merge.

### Noted, not changed

`ProjectCanvas.handlePickerSelect` also calls `saveChart` directly, and the canvas commit path (`commitCanvasConfig`) has **no** read-only hold at all — for any mutation, not just this one. That's a pre-existing VIS-1025 gap across the whole canvas rather than something this PR introduced, and it doesn't produce the rail's orphan (the canvas placement isn't discarded), so I left it alone rather than half-fixing it here. Same for `setItemRef`'s name-trimming: the canvas shares the whitespace-name divergence, but its rejection guard (`next === dashboardConfig`) is already reachable and tested.

tgsoverly (approved):
Ditto on the comments

PR #648, opened 2026-08-28
Post-promote always offers a destination — "New dashboard…" for the zero-dashboard first run (W6)

jpjsonic:
## Problem

`ExplorationPromoteModal`'s fallback dashboard offer (`showFallbackDashboardOffer`, and the matching `hasOffer` term in `handlePromote`) required `dashboards.length > 0`. A first-run user has ZERO dashboards, so on a clean chart promote the modal auto-closed as "nothing to act on" — the one working exit ramp from Explorer to a dashboard never fired in exactly the exit-gate scenario (Dashboard Building v1 dossier W6, finding B2; every field-test tester hits this).

## Fix

- **Gate dropped**: a promoted chart always gets a destination offer — `return_to` when armed, otherwise the fallback offer.
- **"New dashboard…" option** (sentinel `__new__`) is prepended to the offer's select and is the default when no dashboards exist. With existing dashboards nothing changes: the default stays the first dashboard.
- **New store action `placeChartInNewDashboard(chartName)`** (`dashboardStore.js`): creates via the standard inline-create path (`createDashboard` → `createWorkspaceObject('dashboard')`), then **fills the born row's existing empty slot** — #621 makes every new dashboard start with one row holding one empty slot, and the chart lands in that slot via `itemMutations.setItemLeaf` (canonical `${ref(name)}` form), never as an appended second row. Then the new dashboard's tab opens.
- **#617 semantics respected**: persistence goes through `saveDashboard` exactly like the sibling `placeChartInDashboardSlot` — the dashboard was created this instant so there is no working copy to clobber, and the post-save refetch resets `dashboardBaselines`, so the explicit-save working copy starts clean on server truth.
- **Defensive paths** (never drop the chart): a born config with no empty slot appends through the existing `slotToInsertTarget`/`insertItemAtTarget` fallback; a refetch that misses the new dashboard assumes the template's own shape.

## Falsification evidence

With the tests kept and both implementation files stashed (`ExplorationPromoteModal.jsx`, `dashboardStore.js`):

```
Tests:       14 failed, 126 passed, 140 total
  ● dashboardStore placeChartInNewDashboard … › creates a dashboard via the standard path and FILLS the born row's empty slot …
  ● … (all 7 new store tests fail — the action does not exist pre-fix)
  ● ExplorationPromoteModal › … ZERO dashboards: the fallback offer still renders … — never a dead-end auto-close (W6)
  ● … ZERO dashboards: accepting "New dashboard…" creates + places via placeChartInNewDashboard …
  ● … (6 more modal tests fail — pre-fix the modal auto-closes with no offer)
```

`git stash pop` restores the fix; all 140 pass. (The one new test green pre-fix is the deliberate behavior-preservation pin: with existing dashboards the default stays the first dashboard.)

## Tests

- 7 new `dashboardStore` tests: slot filled in place (exactly one row, one item, `${ref(chart)}`), name dedupe reports the real created name, failure propagation for both the create and the placement save, both defensive paths.
- 7 new/updated `ExplorationPromoteModal` tests, including rewrites of the three tests that pinned the old auto-close (`dashboards` undefined + the two "closes on a clean success when no dashboards exist" copies).
- E2E story: `viewer/e2e/stories/exploration-promote-fallback-dashboard-offer.spec.mjs` gains the zero-dashboard walkthrough — soft-deletes every dashboard before the app loads, runs the ordinary source-tile → query → chart → Save-to-project flow, accepts "New dashboard…", and backend-asserts the created dashboard has exactly one row with one item referencing the chart (restored by the existing `commit/discard` cleanup). Written but **not run here** — it needs the shared sandbox ports and runs in the serial `exploration-mutations` project (file already double-registered in `playwright.config.mjs`).
- Full suite `yarn test --watchAll=false`: **7024 passed, 365 suites**; `yarn lint`: clean.

Note: branched off main independently of #643 (same file, different regions — the two diffs don't overlap).

### DIFF
--- a/viewer/e2e/stories/exploration-promote-fallback-dashboard-offer.spec.mjs
+++ b/viewer/e2e/stories/exploration-promote-fallback-dashboard-offer.spec.mjs
@@ -20,8 +20,11 @@
  *
  * Also covers W6 (Dashboard Building v1 — "Post-promote always offers a
  * destination, including a brand-new dashboard"): the ZERO-dashboard
- * first-run case, manufactured in-test by soft-deleting every dashboard
- * before the app loads (afterEach's commit/discard restores them).
+ * first-run case, manufactured in-test by HIDING the project's existing
+ * dashboards from the viewer's list fetch (`page.route`) before the app
+ * loads — never by deleting them, so this file's blast radius on the shared
+ * sandbox stays "one appended row" rather than "the whole dashboard
+ * collection, restored only if afterEach gets to run".
  *
  * Precondition: sandbox running (integration project — has ≥1 real
  * dashboard), e.g.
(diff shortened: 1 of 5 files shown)


jpjsonic:

## Review fix pass — 790d8065

All six findings verified against the real component/store before anything was touched; **all six are real** (findings 1 and 4 are the same destination-flip defect reported twice, and 4 also restates 2's orphan), so nothing was rejected. Four distinct defects, all fixed.

### 1 + 4a — destination flip after create-succeeds/place-fails (major, CONFIRMED)

`placeChartInNewDashboard` creates the dashboard *before* it places the chart, and the create refetches `dashboards`. With `fallbackDashboardChoice` still `null`, the derived default re-read as the orphan the failed attempt had just created. Repro against the real component, before the fix:

```
● ZERO dashboards: a failed create-and-place does NOT flip the destination — the retry still creates (W6)
    Expected element to have text content: New dashboard…
    Received: Add churn_chart tonew-dashboard?Add
```

The retry click then routes to `placeChartInDashboardSlot(name, chart)` with no slot → `{kind:'between-rows', index: rows.length}` → a **second row below the still-empty born one**, exactly what #621 and this offer exist to prevent.

**Fix:** the default destination is now FROZEN (a ref) the first time the offer renders. Still derived synchronously — the VIS-1226 render-timing flake the original comment guards against stays fixed (no effect, no extra render).

### 2 + 4b — orphan dashboard, multiplied by every retry (major/minor, CONFIRMED)

No rollback: a failed placement left an empty dashboard in pending changes, unmentioned in the inline error, and the retry minted another. Proven pre-fix:

```
● retrying after a rolled-back failure creates the SAME name again, never a second orphan
    -   "dashboardName": "new-dashboard",
    +   "dashboardName": "new-dashboard-2",
```

**Fix:** a failed placement rolls the create back via `deleteDashboard`. That is clean for a never-published object — `ObjectManager.mark_for_deletion` **drops** the cache entry outright rather than tombstoning it, so no staged change is left behind. A rollback that itself fails is **disclosed** in the returned error (`An empty dashboard "new-dashboard" was created and could not be removed — delete it from the Library.`) rather than reported as a clean failure the project state contradicts.

### 3 — `__new__` sentinel can collide with a real dashboard (minor, CONFIRMED)

The old comment's justification was wrong: `NamedModel.name` is a plain `Optional[str]` with no validator (the viewer's `NAME_PATTERN` only gates names typed into viewer forms), so a hand-authored `.visivo.yml` can declare `dashboards: [{name: __new__}]`. `Select` matches by value and takes the first hit. Pre-fix, with such a project:

```
● a real dashboard named like the sentinel is still a distinct, selectable destination (W6)
    Expected element to have text content: __new__
    Received: Add churn_chart toNew dashboard…?Add
```

**Fix:** the sentinel escapes itself out of the way of the project's actual names (`__new__` → `___new___` → …), and the user's pick is carried as a tagged `{ isNew: true } | { name }` instead of a bare string, so a name can never be mistaken for the option.

### 5 — the "ALONGSIDE" test never asserted the option (major, CONFIRMED by mutation)

Correct: the old assertion read the **closed** react-select control, which renders only the selected label. Reproduced the review's mutation (drop the sentinel option whenever `dashboards.length > 0`) — the suite was green at 104/104 before, and now:

```
● with existing dashboards the OPEN menu really lists "New dashboard…" alongside them, and picking it creates one (W6)
    expect(received).toEqual(expected)
    -   "New dashboard…",
        "sales",
        "ops",
```

The replacement opens the menu, reads the real option list, then selects the sentinel and drives it end-to-end (`placeChartInNewDashboard` called, `placeChartInDashboardSlot` not, the created dashboard's tab opened).

### 6 — e2e blast radius (minor, accepted)

The zero-dashboard state is no longer manufactured by soft-deleting every dashboard in the shared sandbox. It is manufactured by **hiding** them from the viewer's list fetch (`page.route` on `/api/dashboards/`), which is exactly the state under test (`dashboards` empty in the store is all the modal reads) and leaves the backend untouched — the file's blast radius is back to "one appended row". Added a guard that fails loudly if a fixture dashboard is named `new-dashboard*` (the create could otherwise overwrite it), and the created dashboard is identified by *difference from the pre-test set* rather than by `toHaveLength(1)`, which holds whether or not the route filter applies to `page.request`.

**Disclosure:** the spec still has not been run live. This machine's data volume is at 100% (≈100–900 MiB free across the session), and starting a sandbox would have risked ENOSPC for the other agents sharing it. The file is `playwright test --list`-verified (parses, registered in `exploration-mutations` only, not double-registered in `parallel`), and the change removes the destructive step rather than adding one.

### Falsification

Implementation stashed, tests kept — 5 of the 6 behavioral guards fail without the fix (2 modal, 3 store, outputs above); the 6th is the anti-mutation guard for finding 5 and was proven to fail under the review's exact mutation. 3 sentinel-escape unit tests cover the escape loop.

### Gates

- `yarn test --watchAll=false`: **7033 passed, 365 suites** (3 consecutive clean full runs; one earlier full run had a single unrelated failure that did not reproduce and coincided with an ENOSPC window on the shared volume).
- `yarn lint`: clean (0 errors).
- No `yarn format`, no `yarn deploy`, no files touched from the do-not-touch list.

Files: `viewer/src/components/views/workspace/ExplorationPromoteModal.jsx`, `viewer/src/stores/dashboardStore.js`, their two test files, and `viewer/e2e/stories/exploration-promote-fallback-dashboard-offer.spec.mjs`.

PR #642, opened 2026-08-28
Regenerate the committed project schema (parity ring was red on clean main)

jpjsonic:
## Problem
`tests/docs/test_docs_parity.py::test_committed_schema_is_fresh` fails on a clean checkout of main — the committed `visivo/src/visivo_project_schema.json` was last regenerated around #585 and the models have moved since. A permanently-red freshness gate masks every future real drift, and IDE YAML validators reject legitimate new-field usage.

## Fix
One command, per the repo protocol: `python -m visivo.generate_project_schema_json`. No model changes.

## Verification
- Before: `test_committed_schema_is_fresh` **fails** on clean main (reproduced locally).
- After: `tests/docs/test_docs_parity.py` 5 passed.
- Full suite: 2443 passed, 2 skipped, 5 xfailed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Dw8Cc4b4J7oX3zQbBsjQ2U

PR #652, opened 2026-08-28
The editor's overlay keeps nesting too: a model-scoped field is not project-level (B12)

jpjsonic:
> **Update (`fdfa011f`) — review fix.** The re-nesting was being undone one line later: `_inject_draft_objects` replaces a model by name, and the wire dict the Explorer sends for *every* open model carries no `metrics`/`dimensions`, so the just-nested field was deleted outright — B12's headline symptom unchanged on the real request, and a bare `${ref(field)}` regressed to a 400. Fixed by making a draft model a partial override (`_carry_forward_nested_fields`) and re-nesting once more after the wire drafts land. The two draft-overlay tests are now parameterised over the client's actual wire shape. Full write-up in the PR comment; gates now `2492 passed`, black clean, `draft_overlay.py` 98% / `project_injection.py` 100%.

## The problem

VIS-1259 settled this for the deploy wire in #639: a metric or dimension authored inside a model is **not its own record**, and nesting is the only thing that says so — neither `Metric` nor `Dimension` has a `model` field to carry an owner, and both are `extra="forbid"`. The **local editor's overlay still flattened it**, which is finding B12 (VIS-1304).

Each manager caches its objects in one flat namespace, so `inject_cached_objects` can only append a cached field to the matching TOP-LEVEL project list. `_parent_name` rides along on the instance, which makes it look survivable, but nothing downstream can act on it: `FieldResolver._find_model_scoped_field` searches the model's **own** `metrics`/`dimensions`, so `${ref(model).field}` — the address the Explorer emits for every computed column — stops finding it.

Two ways that lands, both reproduced against `main` @ 7d31ea69:

```
FLATTENED  ${ref(orders).avg_amount} -> Exception: Column 'avg_amount' not found on model 'orders'
FLATTENED  ${ref(orders).region}     -> "mvhrzfrsckpdxwftwvpdmhaoqjbea"."region"
NESTED     ${ref(orders).region}     -> UPPER("mvhrzfrsckpdxwftwvpdmhaoqjbea"."region")
```

The second is the quiet one and it is B12's headline symptom. When a computed field's alias matches a real column, the reference does not error — the raw column answers instead, the computed expression is dropped, and **a metric reads as a plain dimension** with numbers that are merely wrong.

## Why nesting, and not the fallback the initiative first sketched

VIS-1304 asked for a decision between two strategies. It is decided by Tim's own merges plus one fact that settles it outright.

The alternative was a `_parent_name` fallback inside `_find_model_scoped_field` — teach the reader to accept a project-level-listed field whose `_parent_name` matches the model. It does not survive contact with the commit gate, which re-constructs the project from `model_dump()`:

```
A. dumped project-level metric dict: [{'name': 'avg_amount', 'expression': 'avg(amount)', ...}]
A. Project(**dump) -> ValidationError: Project-level metric 'avg_amount' must reference
                      at least one model ... (StandaloneFieldRefsValidator, #640)

B. dumped model.metrics: ['avg_amount']
B. round-trip nested _parent_name = orders
B. project-level metrics = []
```

`_parent_name` is a PrivateAttr. Flattened, the field dumps with **no owner at all** — the same loss #639 found on the wire — so a fallback keyed on it evaporates at exactly the moment the gate needs it, and #640's `StandaloneFieldRefsValidator` then rejects the field outright. Nesting is in the data and survives. A fallback would also have to be repeated in every reader (resolver, validators, writer, ERD, lineage), which is the two-shapes-for-one-meaning problem #639 rejected.

So: consistent with Tim's model, `_parent_name`-keyed, nesting is the fact.

## The fix

Three call sites share the overlay — the on-save run executor, the Explorer's draft-preview builder (`build_draft_overlay`), and #640's commit gate — so this is fixed **once, in the overlay**, rather than three times in its readers.

- #640's `_renest_model_scoped_fields` moves out of `commit_views.py` into `project_injection.py` as `renest_model_scoped_fields` and runs at the end of `inject_cached_objects`, **after** the cached models are in place, so a field re-nests onto the model the user is actually editing rather than its published version.
- The gate's separate call becomes redundant and goes; the rule now has one home.
- The filter keys off `_parent_name`, not off the type — a genuinely project-level metric stays its own object.
- A field scoped to a model that is not in the project stays top-level, so the normal validators name the mistake instead of it being silently dropped (Tim's behaviour, preserved and re-tested).

Nothing in `visivo/models/` changed, so no schema regeneration — no overlap with #642.

## Falsification

Impl removed (`renest_model_scoped_fields(project)` call deleted), tests kept — **12 fail**:

```
tests/server/jobs/test_project_injection.py
  …metric_comes_back_nested_under_its_model              AssertionError: assert [] == ['avg_amount']
  …dimension_comes_back_nested_under_its_model           AssertionError: assert [] == ['loud_region']
  …keeps_its_parent_and_is_not_addressable_project_level ValueError: not enough values to unpack
  …cached_field_replaces_the_published_one_of_the_same_name
                                                         assert [('avg_amount','avg(old)')] == …
  …re_nests_onto_the_cached_model_not_the_published_one  AssertionError: assert [] == ['avg_amount']
  …resolves_through_its_qualified_reference              Exception: Column 'avg_amount' not found
                                                                    on model 'orders'.
  …not_silently_replaced_by_a_column_it_shadows          assert 'UPPER(' in '"MVHRZ…"."REGION"'
  …nesting_survives_model_dump_where_the_private_attr_does_not
                                                         ValidationError: 1 validation error for Project
tests/query/insight/test_draft_overlay.py
  …cached_model_scoped_metric_lands_on_its_model         AssertionError: assert [] == ['avg_total']
  …cached_model_scoped_metric_compiles_as_an_aggregate   Exception: Column 'avg_total' not found
                                                                    on model 'orders_q'.
tests/server/views/test_commit_views.py
  …MODEL_SCOPED_draft_is_re_nested_and_allowed           (Tim's own test regresses)
  …gate_validates_the_nested_shape_that_would_land_on_disk
```

The three that correctly stay green with the impl removed are the negative-space guards: a project-level field is left alone, an orphaned scope stays top-level, a project with no models does not blow up.

## Tests

- `poetry run pytest tests/` → **2479 passed**, 2 skipped, 5 xfailed, 1 xpassed
- `black --check --target-version py312 .` → clean, 544 files
- Coverage on touched files: `visivo/server/jobs/project_injection.py` **100%**; `visivo/server/views/commit_views.py` 97% (the 8 uncovered lines are pre-existing and untouched here)

New: `tests/server/jobs/test_project_injection.py` (13 tests, factory-boy throughout) plus two on the Explorer's draft-preview path in `tests/query/insight/test_draft_overlay.py` covering the *cached* (saved-but-uncommitted) field, which that file previously only covered for *wire* drafts.

## Scope note

VIS-1304 lists three silent-degradation routes. This PR closes the flattening one. The other two — the client-schema fall-through that returns 200 for an unrunnable query, and the silent project-level demotion of a `draft_dimensions` entry naming an unknown model — are untouched and should build on the merged `Diagnostic` contract from #628.

### DIFF
--- a/tests/query/insight/test_draft_overlay.py
+++ b/tests/query/insight/test_draft_overlay.py
@@ -280,6 +280,66 @@ class TestBuildDraftOverlayModelScopedDraftFields:
         assert matching[0].expression == "avg(amount)"
 
 
+class TestBuildDraftOverlayModelScopedCachedFields:
+    """The other way a model-scoped field reaches this overlay: not on the wire
+    as a draft, but SAVED into the editor's cached tier and not yet committed —
+    a computed column the user promoted, then went on using.
+
+    Those arrive through ``inject_cached_objects``, which can only append them
+    to the flat top-level list. Left there, the very reference the Explorer
+    emits for them — ``${ref(model).field}`` — no longer finds them, and the
+    same "computed metric behaves like a dimension" symptom comes back on a
+    field that was, from the user's side, already promoted."""
+
+    def _bar_insight(self):
+        return {
+            "name": "draft_insight",
+            "props": {
+                "type": "bar",
+                "x": "?{${ref(orders_q).region}}",
+                "y": "?{${ref(orders_q).avg_total}}",
+            },
+        }
+
+    def _flask_app_with_cached_metric(self, project):
+        from types import SimpleNamespace
+
+        cached = Metric(name="avg_total", expression="avg(amount)")
+        cached.set_parent_name("orders_q")
+        app = FlaskAppStub(project)
+        app.metric_manager = SimpleNamespace(cached_objects={"avg_total": cached})
+        return app
+
+    def test_a_cached_model_scoped_metric_lands_on_its_model(self, project_with_model):
+        app = self._flask_app_with_cached_metric(project_with_model)
+
+        project, dag, _ = build_draft_overlay(app, self._bar_insight())
+
+        model_node = dag.get_descendant_by_name("orders_q")
+        assert [m.name for m in model_node.metrics or []] == ["avg_total"]
+        assert all(m.name != "avg_total" for m in (project.metrics or []))
+
+    def test_a_cached_model_scoped_metric_compiles_as_an_aggregate(
+        self, project_with_model, tmp_path
+    ):
+        app = self._flask_app_with_cached_metric(project_with_model)
+
+        _, dag, insight = build_draft_overlay(app, self._bar_insight())
+        model_node = dag.get_descendant_by_name("orders_q")
+        query_info = insight.get_query_info(
+            dag,
+            str(tmp_path),
+            schema_overrides={
+                "orders_q": {model_node.name_hash(): {"region": "VARCHAR", "amount": "DOUBLE"}}
+            },
+            force_dynamic=True,
+        )
+
+        post_query = (query_info.post_query or "").lower()
+        assert "avg(" in post_query
+        assert "group by" in post_query
+
+
 class TestBuildDraftOverlayNameShadowing:
     def test_a_draft_model_reusing_a_real_published_name_shadows_it_in_the_ephemeral_copy_only(
         self, flask_app, project_with_model
--- a/tests/query/insight/test_draft_overlay.py
+++ b/tests/query/insight/test_draft_overlay.py
@@ -131,6 +131,31 @@ class TestBuildDraftOverlayValidation:
                 draft_models=[{"name": "missing_sql_field"}],
             )
 
+    def test_a_draft_model_that_fails_pydantic_raises_draft_overlay_error(self, flask_app):
+        # The module's stated contract: a Pydantic ValidationError never escapes
+        # uncaught to the view. `sql` is Optional (post-#533), so the gate above
+        # catches a MISSING one — this covers a dict that fails validation
+        # outright, which is a different branch.
+        with pytest.raises(DraftOverlayError, match="Invalid draft models"):
+            build_draft_overlay(
+                flask_app,
+                insight_config(),
+                draft_models=[{"name": "orders_q", "sql": "select 1", "not_a_field": True}],
+            )
+
+    @pytest.mark.parametrize("field", ["draft_metrics", "draft_dimensions"])
+    def test_a_malformed_model_scoped_draft_field_raises_draft_overlay_error(
+        self, flask_app, field
+    ):
+        # Same contract on the model-scoped branch, which validates each config
+        # separately after popping its `model` key.
+        with pytest.raises(DraftOverlayError, match="Invalid draft"):
+            build_draft_overlay(
+                flask_app,
+                insight_config(),
+                **{field: [{"name": "avg_total", "model": "orders_q"}]},
+            )
+
     @pytest.mark.parametrize("blank_sql", ["", "   ", "\n\t"])
     def test_a_blank_or_whitespace_only_sql_is_gated_like_missing_sql(self, flask_app, blank_sql):
         # Phase 4 review fix: the gate must treat "" / "   " the same as
@@ -253,7 +278,10 @@ class TestBuildDraftOverlayModelScopedDraftFields:
     def test_a_draft_field_naming_an_unknown_model_falls_back_to_project_level(self, flask_app):
         # A metric whose `model` names no known model must not silently vanish —
         # it falls back to the project-level list so a later resolution failure
-        # is a clean, surfaced error rather than a silent drop.
+        # is a clean, surfaced error rather than a silent drop. This is also the
+        # guard on the FINAL re-nest pass `build_draft_overlay` runs after the
+        # wire drafts land: that fallback deliberately leaves `_parent_name`
+        # unset, so the pass must leave the field exactly where it is.
         project, _, _ = build_draft_overlay(
             flask_app,
             self._bar_insight(),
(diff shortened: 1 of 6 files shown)


jpjsonic:

## Review fix: the re-nesting was being undone one line later

All three findings are **real**, and two of them are the same defect. I reproduced each one against `4e1055e8` before touching anything, and against `7d31ea69` (main) for the regression claim. Nothing was rejected.

### What was actually wrong

`build_draft_overlay` ran `inject_cached_objects` (which now re-nests) and then `_inject_draft_objects`, whose `models` branch merges by name — and `merge_objects_into_list` is a **replacement** (`by_name[name] = obj`). The fresh `SqlModel` validated off the wire has `metrics=[]`, `dimensions=[]`, so it evicted the field that had just been nested onto that model one line earlier. The field ended up in **neither** `model.metrics` **nor** `project.metrics`.

That is not a hypothetical wire shape. `useDraftInsightPreview.js:380-386` builds a `draft_models` entry for *every* entry in `modelStates` with `sql` + `sourceName` — published models included — and it carries only `{name, sql, source}`. Both `/api/insight-compile-draft/` and `/api/insight-execute-draft/` go through `build_draft_overlay`, so this was the preview lane *and* the executed-numbers lane.

Reproduction on `4e1055e8`, factories + the client's own wire shape (`schema_overrides` lists `avg_total` among the raw columns, which is what `modelSchemas` does):

```
--- cached (saved-but-uncommitted) model-scoped metric ---
  no draft_models   : nested=['avg_total'] top=[] AVG=True  GROUPBY=True
  WITH draft_models : nested=[]            top=[] AVG=False GROUPBY=False
--- published (already on disk) model-scoped metric ---
  no draft_models   : nested=['avg_total'] top=[] AVG=True  GROUPBY=True
  WITH draft_models : nested=[]            top=[] AVG=False GROUPBY=False
--- bare ${ref(avg_total)} with cached metric ---
  no draft_models   : nested=['avg_total'] top=[] AVG=True  GROUPBY=True
  WITH draft_models : RAISED PydanticCustomError: The reference "ref(avg_total)"
                      on item "i" does not point to an object.
```

So on the shape the Explorer actually sends, B12's headline symptom was unchanged — `SELECT region, avg_total` with no `AVG` and no `GROUP BY`, the metric reading as a plain dimension — and the bare-ref case had **regressed** relative to main. Same input on `7d31ea69`:

```
--- bare ${ref(avg_total)} with cached metric ---
  WITH draft_models : nested=[] top=['avg_total'] AVG=True
```

Flattened-but-present resolved (the metric was still a DAG node whose `child_items()` emits `ref(orders_q)` — the edge #639 depends on); deleted does not, and `insight_compile_views.py:64-68` turns that into a 400 for an insight that used to render. Going from *misplaced* to *gone* is strictly worse, and the reviewer was right to call it high.

The published/committed-to-YAML half behaves identically and is **pre-existing** (it reproduces on main too), but it is the same root cause and the same fix, so it is closed here as well.

### The fix (both halves in `draft_overlay.py`; `project_injection.py` is unchanged)

1. **`_carry_forward_nested_fields`** — a `draft_models` entry is a *partial override* of the model the user is editing (its SQL and source), not a redeclaration of everything that model owns. Only keys the dict actually carries override, so an absent `metrics` key means "unchanged" and an explicit `"metrics": []` still means "declares none". The carried-over fields get `set_parent_name` re-asserted, because post-construction assignment bypasses `SqlModel`'s after-validator — the same contract `_inject_model_scoped_fields` already uses.
2. **A final `renest_model_scoped_fields(project)`** after `_inject_draft_objects`. The rule has to hold on the *final* overlay: `draft_models` can introduce the owner model *after* `inject_cached_objects` has run (a scratch model that lives only in the Explorer's local `modelStates` while a field scoped to it was already saved into the manager cache). The pass is idempotent, and it never claws back `_inject_model_scoped_fields`' deliberate project-level fallback, which leaves `_parent_name` unset on purpose.

`inject_cached_objects` keeps its own re-nest call — that is the one the on-save run executor and #640's commit gate rely on, and it is what lets a cached field land on the *cached* model rather than the published one.

After the fix, all six cases above come back `nested=['avg_total'] top=[] AVG=True GROUPBY=True`.

### Finding 2 (tests didn't exercise the client's wire shape) — also real, also fixed

The two cached-field tests passed `draft_models=None`, a shape the client never sends. They are now parameterised over `no_draft_models` / `client`, using a module-level `_CLIENT_DRAFT_MODEL` that documents where the shape comes from. The aggregate test's `schema_overrides` now also lists `avg_total` as a raw column, which is what makes the failure *silent* rather than loud — without it the test would have passed for the wrong reason.

New coverage:

- `TestBuildDraftOverlayModelScopedCachedFields` — the two existing tests, both wire shapes, plus a bare-`${ref(field)}` case guarding the 400 regression.
- `TestBuildDraftOverlayDraftModelPreservesNestedFields` — a published nested metric and a published nested dimension survive the client's draft model; the carried-forward field keeps its scope; an explicit `metrics` key still overrides; a scratch model inherits nothing from a model it did not shadow.
- `TestBuildDraftOverlayCachedFieldOnAWireOnlyModel` — a cached field whose owner arrives only on the wire gets nested.

### Falsification

**A. `_carry_forward_nested_fields` call deleted, tests kept — 6 fail:**

```
FAILED …ModelScopedCachedFields::test_a_cached_model_scoped_metric_lands_on_its_model[client]
       AssertionError: assert [] == ['avg_total']
FAILED …ModelScopedCachedFields::test_a_cached_model_scoped_metric_compiles_as_an_aggregate[client]
       assert 'avg(' in 'select "…"."region" as "…", "…"."avg_total" as "…" from "…"'
FAILED …ModelScopedCachedFields::test_a_cached_model_scoped_metric_still_answers_a_bare_ref[client]
       PydanticCustomError: The reference "ref(avg_total)" … does not point to an object.
FAILED …PreservesNestedFields::test_a_published_nested_metric_survives_the_client_draft_model
       AssertionError: assert [] == ['avg_total']
FAILED …PreservesNestedFields::test_a_published_nested_dimension_survives_the_client_draft_model
       AssertionError: assert [] == ['loud_region']
FAILED …PreservesNestedFields::test_a_carried_forward_field_keeps_its_scope_on_the_replacement_model
       IndexError: list index out of range
```

The second one is the whole bug in one line: the raw column answers, the aggregate is gone, and nothing raises.

**B. final `renest_model_scoped_fields(project)` deleted, tests kept — 1 fails:**

```
FAILED …CachedFieldOnAWireOnlyModel::test_the_field_is_nested_once_its_model_arrives_on_the_wire
       AssertionError: assert [] == ['avg_total']
```

The `no_draft_models` halves correctly stay green in both — they are the negative space that says the pre-existing path still works.

### Gates

- `pytest tests/` → **2492 passed**, 2 skipped, 5 xfailed, 1 xpassed (up from 2479 — 13 new/parameterised cases)
- `black --check --target-version py312 .` → clean, 544 files
- `visivo/query/insight/draft_overlay.py` coverage: **98%** (the 2 uncovered lines are the pre-existing `adapter is None` guards, unreachable from `build_draft_overlay`); `visivo/server/jobs/project_injection.py` stays **100%**
- Nothing under `visivo/models/` changed, so still no schema regeneration and no overlap with #642.

PR #653, opened 2026-08-28
Ask the same-source question that was already written (M19 / T1+T2)

jpjsonic:
## Problem

`Relation.validate_same_source` has been in the tree for a long time: complete, well-worded, fully tested — and called by nothing but its own tests. `ProjectValidator` registered eleven validators and none of them for this, and `SingleSourceValidator` only covers metrics and dimensions.

So an insight or relation whose objects span two sources compiles clean, and then:

* `Insight.get_dependent_source` picks `list(models)[0]` off a **set** and builds a query whose CTEs name tables from both databases;
* `run_insight_job.action` independently picks `all_descendants_of_type(...)[0]`;
* the query executes against one of them and the user gets that driver's "table does not exist" about the *other* source's tables.

Because both picks come off sets of objects and strings, *which* half of the join got blamed changed between runs of the same project — the failure read like a flake instead of a fact. Only the draft-preview endpoint had a guard, so `visivo run` had none.

## Fix

**New `visivo/query/source_scope.py`** — the rule as pure functions over the DAG, with one wording used by every gate, and `CrossSourceError` (a `ValueError`) carrying `message`, a `Diagnostic`, and the `error_details` dict the preview run status already speaks.

**T1 — compile time.** `CrossSourceValidator` registered last in `ProjectValidator.__init__` (following #640's `StandaloneFieldRefsValidator` pattern), covering both **relations** and **insights**. It runs after the reference and single-source validators on purpose: a broken `${ref()}`, or a metric that already spans sources, is the more local mistake and should be the sentence the author reads first. `Relation.validate_same_source` stays as a delegating shim (its `output_dir` is now optional — `get_source_for_model` has never read it).

The message names the object, both models and both sources:

```
Insight 'revenue_by_user' references models from more than one source: analytics_db, warehouse.

  Model 'orders' uses source: warehouse
  Model 'users' uses source: analytics_db

Cross-source insights are not currently supported. Every model an insight references
must use the same source.
```

**T2 — runtime guards, same sentence.** `InsightQueryBuilder.__init__` refuses before it resolves a source; `run_insight_job.action` resolves through `resolve_insight_source` instead of `[0]` and attaches `{error_type: "multi_source", error_models, error_sources, diagnostic}` to the `JobResult`; `insight_execute_views` now renders the shared message instead of its own bespoke one (it keeps `error_type: multi_source`, and gains `error_models` + `diagnostic`).

**Determinism.** Every name is sorted before it is reported. Proven, not asserted — see falsification E.

**Diagnostic.** `cross_source` appended to `DIAGNOSTIC_CODES` (append-only respected: nothing renamed or removed), emitted at `run` from the job and at `serve` from the endpoint.

**B3 (subsumed per the dossier).** `getattr(source, "db_schema"/"database", None)` — a `CSVFileSource` is a DuckDB source over a file and carries neither field, so the plain reads raised `AttributeError` on the way to a query that never needed either value (both are only consumed by the non-dynamic CTE `qualify`). This is the getattr half of VIS-1289; that issue's larger "delete csv/xls behind SourceTypeDeprecation" scope is untouched.

### What I deliberately did not do

* **The lazy half of T2** (resolve the source only when `not is_dynamic`). It is not required for the guard — the guard runs *before* the resolution and subsumes the arbitrary pick entirely. It is a T9 enabling step (a dynamic multi-source insight would no longer need a source at all), and on its own today it would only change behaviour for drafts whose source does not resolve. Independent scope; left for T9.
* **The `JoinOperatorPopover` source-mismatch band.** Viewer surface, needs its own e2e/screenshot acceptance. Still open under VIS-1294.
* `run_insight_job.job()`'s `_get_source` still uses the `[0]` pick. Raising there would abort the whole run with a traceback instead of failing one insight into a `JobResult` — the guard belongs in `action`.

## Falsification evidence

Each guard was removed with the tests left in place; every one failed, and the restore is green (`57 passed, 3 xfailed`). Script: `scratchpad/lanes2/falsify.sh`.

**A — unregister `CrossSourceValidator`:**
```
FAILED ...TestItIsRegistered::test_it_runs_last - assert 'SingleSourceValidator' == 'CrossSourceValidator'
FAILED ...TestRelations::test_a_cross_source_relation_fails_the_project - Failed: DID NOT RAISE ValidationError
FAILED ...TestInsights::test_a_cross_source_insight_fails_the_project - Failed: DID NOT RAISE ValidationError
3 failed, 6 passed
```

**B — remove the `InsightQueryBuilder` guard:**
```
FAILED ...::test_it_refuses_to_build_across_sources - Failed: DID NOT RAISE CrossSourceError
```

**C — restore the getattr-free attribute reads:**
```
E  AttributeError: 'CSVFileSource' object has no attribute 'db_schema'. Did you mean: 'get_schema'?
FAILED ...::test_a_file_source_without_db_schema_does_not_raise_attribute_error
```

**D — restore `run_insight_job`'s arbitrary `[0]` pick:**
```
>   assert result.error_details["error_type"] == "multi_source"
E   TypeError: 'NoneType' object is not subscriptable
FAILED ...TestRunInsightJob::test_the_job_fails_with_the_message_and_typed_details
```

**E — unsort the message.** This is the non-determinism, reproduced: the same broken project blames a different half of the join depending on `PYTHONHASHSEED`.
```
E  AssertionError: message varies with hash seed: {
     "...\n  Model 'apple' uses source: s_apple\n  Model 'zebra' uses source: s_zebra\n...",
     "...\n  Model 'zebra' uses source: s_zebra\n  Model 'apple' uses source: s_apple\n..."
   }
E  assert 2 == 1
FAILED tests/query/test_source_scope.py::test_the_message_is_identical_under_every_hash_seed
```

**F — restore the endpoint's bespoke wording:**
```
E  assert 'references models from more than one source: warehouse, warehouse2' in
     "Insight 'cross' references models from more than one source and cannot be previewed server-side."
```

## Tests

`2494 passed, 2 skipped, 5 xfailed, 1 xpassed` — full suite, zero failures. `black --check --target-version py312 .` clean (548 files).

New: `tests/query/test_source_scope.py` (15), `tests/models/validators/test_cross_source_validator.py` (9), `tests/jobs/test_run_insight_job_cross_source.py` (5). All use factory-boy factories.

Three existing tests are updated, all for the same reason worth knowing about: **a cross-source project can no longer be constructed at all.** `Project`'s `mode="after"` validator runs `ProjectValidator`, so `Project(...)` with models on two sources now raises `ValidationError` — which is the fix, but it means any test that wants the invalid state has to reach it the way the real bypass paths do: build the legal project, move one model's source afterwards (nothing re-validates on assignment), `invalidate_dag_cache()`. That is also exactly how the runtime-guard tests prove the guards fire on input the validator never saw.

## Schema

`python -m visivo.generate_project_schema_json` produces **no diff** — `Diagnostic` is not reachable from `Project`, so `visivo/src/visivo_project_schema.json` is unchanged and this PR does **not** overlap #642's regeneration of that file.

Linear: VIS-1294.

### DIFF
--- a/tests/jobs/test_run_insight_job_cross_source.py
+++ b/tests/jobs/test_run_insight_job_cross_source.py
@@ -1,21 +1,31 @@
 """The runtime half: what happens to a cross-source insight that got past compile.
 
-``CrossSourceValidator`` refuses to let such a project exist, so everything here
-reaches the invalid state the way the code paths that skip project validation do
-— an in-memory overlay, a project assembled in code, a mutation after
-construction. The point of the guard is that those paths fail with the SAME
-sentence rather than executing a query whose CTEs name tables the chosen source
-has never heard of.
+Two lanes, two opposite answers, and both of them are the point:
+
+* **Static** — ``CrossSourceValidator`` refuses to let such a project exist, so
+  the tests below reach the invalid state the way the code paths that skip
+  project validation do (an in-memory overlay, a project assembled in code, a
+  mutation after construction). Those paths must fail with the SAME sentence
+  rather than executing a query whose CTEs name tables the chosen source has
+  never heard of.
+* **Dynamic** — ``TestTheDynamicLane`` runs the real jobs against two real
+  DuckDB files and asserts the insight SUCCEEDS, because there the join is
+  DuckDB's over each model's own parquet. Nothing about this test is a mock:
+  if the guard ever fires on the dynamic lane again, the assertion that the
+  post_query joins both model hashes and returns the joined row fails.
 """
 
 import json
 
+import duckdb
 import pytest
 
 from tests.factories.model_factories import (
     DuckdbSourceFactory,
+    InputFactory,
     InsightFactory,
     ProjectFactory,
+    RelationFactory,
     SqlModelFactory,
 )
 from visivo.jobs.run_insight_job import action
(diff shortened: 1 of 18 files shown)


jpjsonic:

## Review fix pass — the guard was a false positive on a documented, working shape

Thanks — finding #1/#3 was right, and it was the worst class of wrong for a validation lane. I reproduced it before changing anything.

### Verified, not assumed

Two DuckDB files, `orders` on `source_a`, `users` on `source_b`, a relation, an Input, an insight over both:

```
is_dynamic: True
model orders: success=True      # materialised against source_a
model users:  success=True      # materialised against source_b
INSIGHT JOB success: True
files: ['orders.parquet', 'users.parquet']
post_query: SELECT ... FROM "<orders_hash>" INNER JOIN "<users_hash>" ON ...
EXECUTED OK, rows: [(10, 30)]
```

That is with the guards off. With them on, `Project(**cfg)` itself raised — so not just that insight, the **whole project stopped loading**, every other dashboard included:

```
PROJECT FAILED: ValidationError
  Value error, Relation 'o_to_u' connects models from different sources.
```

And `mkdocs/topics/sources.md:27` (Tim, 2026-07-23, on main) says exactly this is supported: *"you can bring data together in a single chart with insights whose models originate from different sources."* The dossier's own T2 line — "resolve only when not is_dynamic" — which the PR body dismissed as independent scope, was the correctness condition.

### The fix: the rule is about a lane, not an object

| lane | what executes | cross-source |
|---|---|---|
| **static** (`pre_query`) | every model as a CTE in ONE statement -> one `source.read_sql` | refused, unchanged |
| **dynamic** (`post_query`; an `Input` descendant, or `force_dynamic`) | `pre_query` is `None`; each model already materialised against its OWN source, DuckDB joins the parquet | allowed |

- `cross_source_insight_error` / `resolve_insight_source` take a **keyword-only** `is_dynamic`, defaulting to `False` so the strict reading is what a caller gets by accident. A test asserts the flag can't be passed positionally (position 4 is `output_dir`).
- `InsightQueryBuilder` now resolves through `resolve_insight_source(..., is_dynamic=self.is_dynamic)` instead of `insight.get_dependent_source` — one call does the refusal *and* kills the last `list(a_set)[0]` pick, so the dynamic lane's dialect source is stable too. `force_dynamic=True` (compile-draft) is covered by `self.is_dynamic`.
- `run_insight_job` and `insight_execute_views` pass their own lane. A dynamic cross-source draft now falls through to the endpoint's existing **409 `requires_client_lane`** handoff instead of 400-ing a chart that works.

### The relation rule, one level down

Same error: a relation is a join *condition*, not an executable thing, so it can't be judged alone. `CrossSourceValidator` now asks it only for relations whose two models both land inside **one static insight's** model set — the same scoping `RelationGraph(relevant_models=...)` already uses. Per-insight, not unioned: `orders` in one insight and `users` in another never meet in a single statement.

That leaves alone: a relation mid-authoring (no consumer yet), and a relation only dynamic insights use. It still fires — first, before the insight message, so the author reads about the join they wrote — when a static insight compiles it.

**Finding #7 (save-time gate on `POST /api/relations/<name>/`) — rejected, and this is why.** #640's precedent is "refuse a save the project would refuse", and it holds for metrics/dimensions because `SingleSourceValidator` makes those unconditionally wrong across sources. A cross-source relation is *not* unconditionally wrong — its legality depends on the execution mode of whatever consumes it, which the save endpoint cannot know. A hard gate there would block the documented multi-source workflow at the exact moment the user is building it. `Relation.validate_same_source` keeps its unconditional form (it is the pre-existing API) with the caveat now written on it.

### The other findings

- **#2 (low)** — confirmed, fixed at the render site: `whitespace-pre-line break-words text-left` on `InsightPreview`'s error `<p>`. `ChartPreview` already does this for its technical details; `InsightPreview` didn't, so every deliberately multi-line backend error collapsed. Test uses an identity normalizer (the default one collapses whitespace — the thing under test).
- **#6 (low)** — confirmed, fixed: `cross_source` added to `viewer/src/types/diagnostic.js`, plus a parity test that reads the JS mirror and diffs it against `DIAGNOSTIC_CODES`. Both files call each other the contract; now something enforces it.
- **#4 (medium)** — confirmed. Added a case with a **real** model whose source doesn't resolve (unset `source`, nothing tying it to a Source in the dag — the shape a draft-overlay model arrives in), plus the consequence test: an unresolved model must not manufacture a cross-source complaint naming a source that doesn't exist. Your exact mutation now fails.
- **#5 (medium)** — confirmed. The seeded script now builds the insight message and the dynamic-lane source pick, not just the relation message, over **three** sources rather than two and six seeds rather than four. With two names an unsorted build is right half the time; with three it is 1-in-6 per seed, so "all six seeds agree AND agree alphabetically" stops being something chance produces.

### Falsification (impl mutated, tests kept)

| mutation | result |
|---|---|
| drop the `is_dynamic` short-circuit | **8 failed** (source_scope x5, jobs x2, execute-views x1) |
| drop the validator's lane + relation scoping | **4 failed** |
| `resolved[model.name] = getattr(source,"name",None) or "__unresolved__"` | **2 failed** (was fully green) |
| `sorted(set(resolved.values()))` -> `list(...)` | **3-4 failed on every one of 5 runs** (was 1 green run in 6) |
| remove `whitespace-pre-line` | **1 failed** |
| remove `'cross_source'` from the viewer mirror | **1 failed** — `DIAGNOSTIC_CODES has drifted. Python only: ['cross_source']` |

### Gates

- `pytest tests/ -q` — **2509 passed**, 2 skipped, 5 xfailed, 1 xpassed
- `black --check --target-version py312 .` — clean, 548 files
- `yarn test --watchAll=false` — **7038 passed**, 366 suites
- `yarn lint` — 0 errors
- `python -m visivo.generate_project_schema_json` — no diff

Both of your original repro scripts now pass on the fixed tree with **no guards disabled**: the project constructs, and the job returns `success=True` with both parquet files and an executable join.

PR #654, opened 2026-08-28
Fail the build when a positional axis resolves to a STRUCT (WB9 / S5-14)

jpjsonic:
## Problem (S5-14)

An insight whose `x` — or any other positional axis — resolves to a **STRUCT** builds *successfully* and renders a **blank chart with a SUCCESS job**. Nothing anywhere reports a problem. That is the worst failure shape we ship: the user sees an empty panel and has no thread to pull.

Root cause, reproduced end to end on `main`:

```python
sqlglot.parse_one("?{site}", read="duckdb")
# Struct(expressions=[Column(this=Identifier(this=site))])   →  {'_0': site}
```

A doubled query string `x: ?{?{${ref(m).site}}}` leaves a residual `?{...}` in the resolved SQL; SQLGlot silently reads it as a DuckDB struct literal, the pre-query becomes `SELECT STRUCT("m"."site") AS "…"`, and the parquet column comes back `STRUCT(VARCHAR)`. Plotly draws nothing.

`prop_type_validator.py` already existed but could not catch it: it fires **only** on slice forms `?{expr}[N]`, and its `_NUMERIC`/`_STRING`/`_BOOLEAN` sets omit STRUCT, so STRUCT classed as `"unknown"` → no validation. S5-14 was fully unguarded.

The editor-side cause of the double-wrap is M24 (Stream G right-rail); this gate stops the *failure mode* in general, not just that one producer.

## Fix

`prop_type_validator.py` grows a second, independent check: a **positional-axis plottability gate** that runs on every bound prop at build time (`InsightQueryBuilder.build()`, both the static and the `force_dynamic` Explorer-preview path).

On a hit it raises `PositionalAxisTypeError` — a `ValueError` (so every existing `except Exception` build handler behaves exactly as before) carrying a structured **`Diagnostic`**, the same way `JoinPathError` carries its structured join fields. New append-only code in `DIAGNOSTIC_CODES`: `non_plottable_axis_type`.

The message names all four things the user needs:

```
Insight 'site_totals': positional axis prop 'props.x' resolves to a STRUCT built
from column 'site'. A chart axis cannot plot a STRUCT, so this insight would build
successfully and then render an empty chart. Bind 'props.x' to a single scalar
column or expression. A STRUCT here is almost always a doubled query string —
check the prop's value for a nested '?{?{ ... }}', which SQLGlot parses as a
struct literal.
```

Where the surrounding code supports it, the Diagnostic is *emitted*, not just carried: `/api/insight-compile-draft/` and `/api/insight-execute-draft/` add an **additive** `diagnostic` key to their existing 400 body (`error` unchanged; the key is absent for failures that produced no Diagnostic, so no client sees a difference until it opts in).

Incidental cleanup: the builder's `AS "<hash>"` alias strip was a regex over SQL. It is gone — the alias now rides along on the fragment and is unwrapped from the parsed AST by `_infer_expression_type`.

## False positives — the allowlist and why it is shaped this way

**This was the main design risk**, so classification is deliberately **three-way**, not a bare allowlist test:

| bucket | contents | gate behaviour |
|---|---|---|
| `plottable` | `PLOTTABLE_AXIS_TYPES` — 73 enumerated sqlglot type members | pass |
| `non_plottable` | `NON_PLOTTABLE_AXIS_TYPES` = `{STRUCT, OBJECT, NESTED, UNION, MAP}` | **fail** |
| `unknown` | everything else (sqlglot knows ~123 type names) | pass |

The costs are asymmetric. A missed silent-blank chart costs one bad render; a false positive turns a build that used to work into a hard failure and blocks everything. So the gate only ever fails on a type we are *certain* about, and a type we have not triaged — or one a future sqlglot release adds — can never start failing a working build.

**Allowlist contents** (`PLOTTABLE_AXIS_TYPES`, the positive half, enumerated so "did we just start rejecting something?" is answerable by reading one list):
- **numerics** of every width/sign/precision, including the unsigned and wide integer types ClickHouse/DuckDB emit (`UBIGINT`, `INT128`, `UDECIMAL`, `DECIMAL256`, …) and the serial/money/year scalars;
- **strings** of every flavour (categorical axes) plus the string-backed scalars — `UUID`, `INET`, `IP*`, `ENUM*`, `LOWCARDINALITY`, `XML`, and `JSON`/`JSONB`/`VARIANT`/`SUPER` (semi-structured columns arrive as JSON **text** after the parquet round-trip and plot fine as categories);
- **temporals** (`DATE*`, `TIME*`, `TIMESTAMP*`, `INTERVAL`) — date axes;
- **BOOLEAN**, which plots as a two-value categorical axis.

Names are `exp.DataType.Type` **members** (what `annotate_types` puts in `DataType.this`), not SQL spellings — so `INT` not `INTEGER`, `DOUBLE` not `REAL`. A test asserts every name in both sets is a real member, so a typo cannot become a dead entry that silently never matches. (The six alias spellings I first wrote — `INTEGER`, `REAL`, `NUMERIC`, `NUMBER`, `STRING`, `BOOL` — were caught by exactly that test and removed.)

**Denylist reasoning:** sqlglot's own `exp.DataType.STRUCT_TYPES` (`STRUCT`/`OBJECT`/`NESTED`/`UNION`) plus `MAP`. After the parquet round-trip each row's value is an object; an axis of objects draws nothing, and no slice rescues it (slicing picks one row's value — still a record). Spelled out in this file rather than imported, so the rejected set is pinned here and cannot widen under us when sqlglot reclassifies a type. A test asserts `STRUCT_TYPES ⊆` our denylist so a future sqlglot record type shows up as a failure rather than a silent gap.

**`ARRAY`/`LIST` are deliberately NOT rejected** even though they are also nested: a positional prop with a scalar slice (`x: ?{...}[0]`) binds a **single row's** value, so an array column becomes a perfectly plottable array of scalars. Rejecting arrays would be a real false positive.

**Prop scoping.** Only the **exact top-level** paths `props.<name>` for: `x`, `y`, `z` (cartesian/3-D), `lat`, `lon` (geo), `r`, `theta` (polar), `a`, `b`, `c` (ternary/carpet), `real`, `imag` (smith), `open`/`high`/`low`/`close` (OHLC), `values`/`labels`/`parents` (pie/funnelarea/sunburst/treemap/icicle). Derived by reading all 51 trace schemas in `visivo/schema/`. Deliberately excluded, each for a reason recorded in the source:
- `ids` — on every trace type, but it keys animation frames/selection, not a coordinate;
- `customdata`, `text`, `hovertext`, `meta` — Plotly explicitly allows arbitrary nested values here; rejecting a STRUCT would be a false positive;
- `props.domain.x`, `props.error_x.array`, `props.marker.*` — leaves that merely *share a name* with an axis, which is why matching is anchored to the top level;
- `parcoords`/`parcats`/`splom` `dimensions[i].values` and `sankey`'s `node`/`link` — genuinely positional but nested inside array-of-object props with their own binding rules; out of scope rather than guessed at;
- `area`'s legacy `t` — too generic a name to claim on one deprecated trace type.

## Falsification evidence

Every guard was falsified by stashing the implementation and keeping the tests.

**A — remove the gate call from `build()`** (i.e. `main`'s behaviour):
```
FAILED test_struct_on_x_fails_the_build_naming_insight_prop_column_and_type - Failed: DID NOT RAISE PositionalAxisTypeError
FAILED test_struct_on_x_carries_a_structured_diagnostic              - Failed: DID NOT RAISE PositionalAxisTypeError
FAILED test_struct_on_y_also_fails                                   - Failed: DID NOT RAISE PositionalAxisTypeError
FAILED test_struct_on_the_dynamic_preview_path_also_fails            - Failed: DID NOT RAISE PositionalAxisTypeError
FAILED TestCompileDraftPositionalAxisTypeGate::test_struct_bound_to_x_is_a_400_not_a_blank_200 - assert 200 == 400
FAILED TestCompileDraftPositionalAxisTypeGate::test_the_400_carries_the_structured_diagnostic  - KeyError: 'diagnostic'
6 failed, 115 passed
```
Note `assert 200 == 400` — that is the bug itself: on `main` the compile-draft endpoint returns a cheerful 200 the Explorer renders as an empty chart.

**B — drop the positional-prop scoping (gate every prop):**
```
FAILED test_struct_on_a_non_positional_prop_is_unaffected - PositionalAxisTypeError: ... positional axis prop 'props.customdata' resolves to a STRUCT ...
FAILED test_struct_on_a_nested_prop_named_like_an_axis_is_unaffected - AssertionError: assert True is False
FAILED test_check_returns_none_for_a_non_positional_prop  - assert Diagnostic(id='compile:non_plottable_axis_type:i:props.customdata', ...) is None
3 failed, 102 passed
```

**C1 — the naive allowlist-only design** (reject anything not on the allowlist, no third bucket):
```
FAILED test_unrecognised_types_pass_through_rather_than_failing - AssertionError: GEOMETRY
FAILED test_array_types_are_deliberately_not_rejected           - AssertionError: ARRAY
2 failed, 103 passed
```

**C2 — an over-broad denylist that swallows `DATE`/`TIMESTAMP`/`BOOLEAN`** (through the *real* builder, not just the classifier):
```
FAILED test_real_columns_of_every_common_type_still_build[DATE]      - PositionalAxisTypeError: ... resolves to a DATE built from column 'col' ...
FAILED test_real_columns_of_every_common_type_still_build[TIMESTAMP] - PositionalAxisTypeError: ... resolves to a TIMESTAMP ...
FAILED test_real_columns_of_every_common_type_still_build[BOOLEAN]   - PositionalAxisTypeError: ... resolves to a BOOLEAN ...
3 failed, 102 passed
```

All restored; 105 pass in the new file.

## Test results

- `tests/query/insight/test_positional_axis_type_gate.py` — **105 passed** (new file). Covers all three asks: STRUCT on x rejected with the message (insight + prop + column + type); STRUCT on a non-positional prop unaffected; every plottable type still accepted — parametrised over all 73 allowlist entries, plus an end-to-end pass through the real builder for `VARCHAR/DOUBLE/BIGINT/DATE/TIMESTAMP/BOOLEAN/DECIMAL(10,2)`.
- Endpoint + helper coverage in `tests/server/views/test_insight_compile_views.py` and `test_insight_draft_common.py`, `non_plottable_axis_type` registration in `tests/models/test_diagnostic.py`.
- **Full suite: `2577 passed, 2 skipped, 5 xfailed, 1 xpassed`.**
- `black --check --target-version py312 .` — clean (544 files unchanged).
- Real-project smoke: `visivo run` clean on `test-projects/integration` (every trace type, indicators, waterfalls, surfaces, splits, inputs), `demo`, `complex-project`, and `docs-examples` (95s, every documented chart). Zero over-rejection. `docs-interactivity` fails on a missing uncommitted `local.duckdb` — pre-existing and unrelated.

## Schema regeneration

`python -m visivo.generate_project_schema_json` was run after the `visivo/models/diagnostic.py` edit: **no diff** — `Diagnostic` is a runtime contract, not a project config object, so it is not part of `visivo_project_schema.json`. Nothing to commit, and therefore **no actual conflict with open PR #642**, which does regenerate that file. Flagging the nominal overlap anyway.

### DIFF
--- a/tests/models/test_diagnostic.py
+++ b/tests/models/test_diagnostic.py
@@ -129,3 +129,15 @@ def test_from_exception_never_raises_on_an_unregistered_code():
     assert diagnostic.code == "unexpected_error"
     assert diagnostic.message == "the real failure"
     assert "typo_code" in diagnostic.detail
+
+
+def test_non_plottable_axis_type_code_is_registered():
+    """WB9's code joins the append-only vocabulary; consumers branch on it."""
+    assert "non_plottable_axis_type" in DIAGNOSTIC_CODES
+    diagnostic = Diagnostic(
+        phase=DiagnosticPhase.COMPILE,
+        code="non_plottable_axis_type",
+        message="positional axis prop 'props.x' resolves to a STRUCT",
+        field="props.x",
+    )
+    assert diagnostic.code == "non_plottable_axis_type"
--- a/tests/models/test_diagnostic.py
+++ b/tests/models/test_diagnostic.py
@@ -1,3 +1,6 @@
+import re
+from pathlib import Path
+
 import pytest
 from pydantic import ValidationError
 
@@ -141,3 +144,58 @@ def test_non_plottable_axis_type_code_is_registered():
         field="props.x",
     )
     assert diagnostic.code == "non_plottable_axis_type"
+
+
+# ---------------------------------------------------------------------------
+# The Python <-> viewer mirror
+# ---------------------------------------------------------------------------
+
+VIEWER_DIAGNOSTIC_JS = (
+    Path(__file__).resolve().parents[2] / "viewer" / "src" / "types" / "diagnostic.js"
+)
+
+
+def _js_string_array(source: str, name: str):
+    """The string literals of a top-level ``export const <name> = [...]``.
+
+    Deliberately a tiny reader over a flat literal array rather than a JS
+    parser: the arrays are hand-maintained flat lists of quoted strings, and
+    the whole point of this guard is that it stays cheap enough to keep.
+    """
+    match = re.search(rf"export const {name} = \[(.*?)\];", source, re.DOTALL)
+    assert match, f"{name} not found in {VIEWER_DIAGNOSTIC_JS}"
+    return [m.group(1) for m in re.finditer(r"'([^']+)'", match.group(1))]
+
+
+@pytest.mark.skipif(
+    not VIEWER_DIAGNOSTIC_JS.exists(),
+    reason="viewer sources are not present (packaged install)",
+)
+def test_viewer_mirror_lists_every_diagnostic_code():
+    """``visivo/models/diagnostic.py`` states: "The mirror typedef for viewer
+    code lives at viewer/src/types/diagnostic.js. Keep the two in sync — the
+    shape is the contract." Nothing enforced it, and WB9's
+    ``non_plottable_axis_type`` shipped on the Python side only.
+
+    A code missing from the viewer list is not inert: viewer surfaces branch on
+    the vocabulary (join-fix cards, not-built empty states), so an unlisted code
+    falls through to the generic path with no card — the same silent failure
+    class the code was added to close.
+    """
+    source = VIEWER_DIAGNOSTIC_JS.read_text()
+    assert _js_string_array(source, "DIAGNOSTIC_CODES") == list(DIAGNOSTIC_CODES), (
+        "DIAGNOSTIC_CODES drifted between visivo/models/diagnostic.py and "
+        f"{VIEWER_DIAGNOSTIC_JS}"
+    )
+
+
+@pytest.mark.skipif(
+    not VIEWER_DIAGNOSTIC_JS.exists(),
+    reason="viewer sources are not present (packaged install)",
+)
+def test_viewer_mirror_lists_every_phase_and_severity():
+    source = VIEWER_DIAGNOSTIC_JS.read_text()
+    assert _js_string_array(source, "DIAGNOSTIC_PHASES") == [p.value for p in DiagnosticPhase]
+    assert _js_string_array(source, "DIAGNOSTIC_SEVERITIES") == [
+        s.value for s in DiagnosticSeverity
+    ]
--- a/tests/query/insight/test_positional_axis_type_gate.py
+++ b/tests/query/insight/test_positional_axis_type_gate.py
@@ -1,0 +1,360 @@
+"""WB9 / S5-14: the build-time gate on positional-axis prop types.
+
+The failure this closes: an insight whose ``x`` (or any other coordinate)
+resolves to a record-shaped SQL type builds "successfully" and renders a BLANK
+chart with a SUCCESS job. The canonical producer is a doubled query string
+``?{?{col}}`` — SQLGlot parses the residual ``?{col}`` as a DuckDB struct
+literal, which round-trips through parquet as ``STRUCT(VARCHAR)``.
+
+Three properties are pinned here, each falsifiable on its own:
+1. a STRUCT on a positional axis FAILS the build, with the insight, prop,
+   column and type in the message;
+2. a STRUCT on a NON-positional prop is untouched (the gate is scoped);
+3. every type an axis can render is still accepted (no over-rejection).
+"""
+
+import json
+import os
+
+import pytest
+from sqlglot import exp
+
+from visivo.models.diagnostic import (
+    DIAGNOSTIC_CODES,
+    Diagnostic,
+    DiagnosticPhase,
+    DiagnosticSeverity,
+)
+from visivo.models.insight import Insight
+from visivo.models.models.sql_model import SqlModel
+from visivo.models.project import Project
+from visivo.models.props.insight_props import InsightProps
+from visivo.query.insight.insight_query_builder import InsightQueryBuilder
+from visivo.query.insight.prop_type_validator import (
+    NON_PLOTTABLE_AXIS_TYPES,
+    PLOTTABLE_AXIS_TYPES,
+    POSITIONAL_AXIS_PROP_NAMES,
+    PositionalAxisTypeError,
+    axis_plottability,
+    check_positional_axis_plottability,
+    is_positional_axis_prop,
+)
+from tests.factories.model_factories import SourceFactory
+
+# ---------------------------------------------------------------------------
+# Build harness: a real InsightQueryBuilder over a one-model project.
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture
+def build_insight(tmpdir):
+    def _build(insight, schema=None, force_dynamic=False):
+        source = SourceFactory()
+        model = SqlModel(name="m", sql="SELECT * FROM t", source=f"ref({source.name})")
+        project = Project(
+            name="p", sources=[source], models=[model], insights=[insight], dashboards=[]
+        )
+        dag = project.dag()
+        schema_dir = os.path.join(str(tmpdir), "schemas")
+        os.makedirs(schema_dir, exist_ok=True)
+        with open(os.path.join(schema_dir, f"{model.name}.json"), "w") as f:
+            json.dump(
+                {model.name_hash(): schema or {"site": "VARCHAR", "amt": "DOUBLE"}},
+                f,
+            )
+        builder = InsightQueryBuilder(insight, dag, str(tmpdir), force_dynamic=force_dynamic)
+        builder.resolve()
+        return builder.build()
+
+    return _build
+
+
+# ---------------------------------------------------------------------------
+# 1. STRUCT on a positional axis is REJECTED (the S5-14 regression)
+# ---------------------------------------------------------------------------
+
+
+def test_struct_on_x_fails_the_build_naming_insight_prop_column_and_type(build_insight):
+    """The exact S5-14 shape: ``x: ?{?{col}}``. Today it builds and renders
+    blank; the gate must abort the build."""
+    insight = Insight(
+        name="site_totals",
+        props=InsightProps(
+            type="bar",
+            x="?{?{${ref(m).site}}}",
+            y="?{count(${ref(m).amt})}",
+        ),
+    )
+    with pytest.raises(PositionalAxisTypeError) as excinfo:
+        build_insight(insight)
+
+    message = str(excinfo.value)
+    assert "site_totals" in message, message  # the insight
+    assert "props.x" in message, message  # the prop
+    assert "'site'" in message, message  # the column
+    assert "STRUCT" in message, message  # the type
+    # Actionable: points at the doubled query string that produced it.
+    assert "?{?{" in message, message
+
+
+def test_struct_on_x_carries_a_structured_diagnostic(build_insight):
+    insight = Insight(
+        name="site_totals",
+        props=InsightProps(type="bar", x="?{?{${ref(m).site}}}", y="?{count(${ref(m).amt})}"),
+    )
+    with pytest.raises(PositionalAxisTypeError) as excinfo:
+        build_insight(insight)
+
+    diagnostic = excinfo.value.diagnostic
+    assert isinstance(diagnostic, Diagnostic)
+    assert diagnostic.code == "non_plottable_axis_type"
+    assert diagnostic.code in DIAGNOSTIC_CODES
+    assert diagnostic.phase == DiagnosticPhase.COMPILE
+    assert diagnostic.severity == DiagnosticSeverity.ERROR
+    assert diagnostic.object.type == "insight"
+    assert diagnostic.object.name == "site_totals"
+    assert diagnostic.field == "props.x"
+    assert "STRUCT" in (diagnostic.detail or "")
+    assert diagnostic.hint
+    # Stable across polls — derived from phase/code/object/field, not random.
+    assert diagnostic.id == "compile:non_plottable_axis_type:site_totals:props.x"
+
+
+def test_struct_on_y_also_fails(build_insight):
+    insight = Insight(
+        name="i",
+        props=InsightProps(type="bar", x="?{${ref(m).site}}", y="?{?{${ref(m).amt}}}"),
+    )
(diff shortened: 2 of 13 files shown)


jpjsonic:

## Review fix pass — `ae5ca2ad`

Eleven findings came in. I reproduced each one against the real
`InsightQueryBuilder` before touching anything. **Seven were real and are
fixed; three I'm rejecting with evidence; one is a claim-accuracy correction to
this PR body.**

---

### Fixed

#### 1. OVER-REJECTION: `OBJECT` denylisted while `VARIANT`/`SUPER`/`JSON` were allowlisted (medium)

The most serious one, because it's a false positive: a project that builds and
renders today started hard-failing. Reproduced end to end —

```
OBJECT    -> FAIL PositionalAxisTypeError: ... resolves to a OBJECT built from column 'payload' ...
VARIANT   -> BUILT OK  inferred=VARIANT
SUPER     -> BUILT OK  inferred=SUPER
JSON      -> BUILT OK  inferred=JSON
```

Snowflake's `OBJECT` **is** a `VARIANT` constrained to objects: the connector
hands both back as JSON text, they land in the same parquet column, and both
plot as categories. `snowflake_source.py` reads
`information_schema.columns.data_type` verbatim, so an `OBJECT` column records
as `"OBJECT"`, and Snowflake is the **only** supported source whose
information_schema emits a bare `OBJECT` data_type. The PR allowlisted
VARIANT/SUPER precisely because "semi-structured columns arrive as JSON *text*
after the parquet round-trip" — that reasoning cannot be true for VARIANT and
false for OBJECT.

Root cause: the denylist was derived from sqlglot's `DataType.STRUCT_TYPES`,
which groups by **SQL-level shape**, not by what reaches the browser. `OBJECT`
moved to the allowlist with that written down next to it.

The invisible-contradiction problem the reviewer identified is now pinned
directly: `test_semi_structured_types_are_classified_consistently` asserts
`OBJECT` and `VARIANT` classify the *same*, so the two lists can no longer
disagree about one member of a family without a test saying so.

#### 2. MISS: any positional prop referencing an `${input}` bypassed the gate (medium)

Reproduced:

```
props: {type: scatter, x: "?{${ref(m).site}}", y: "?{?{${ref(ycol).value}}}"}
->  BUILT OK
    resolved props.y: ?{${ycol.value}} AS "mnxwvmlcfevfjog"
    pre_query:  None
    post_query: SELECT ..., ?{${ycol.value}} AS "mnxwvmlcfevfjog" FROM "..."
```

`sqlglot.parse_one` can't read `${ycol.value}`, so inference returned `None`,
`axis_plottability(None) == "unknown"`, gate silent. And because an
input-bearing insight is dynamic, `pre_query` is `None` and `build()` skips
`validate_query` — so nothing else looked at that post_query either. The
residual `?{…}` reached DuckDB WASM and parsed as a struct exactly as it does
server-side. Blank chart, SUCCESS job, the failure WB9 exists to close.

Fix: when inference returns `None`, retry against
`replace_input_placeholders_for_parsing` — the *same* sample-value substitution
`_build_main_query`'s GROUP BY analysis already uses. Measured across dialects,
the substituted residual annotates cleanly:

```
sqlite/duckdb/snowflake/bigquery/redshift/postgres/mysql  ->  Struct / STRUCT<VARCHAR>
clickhouse                                                ->  VarMap / MAP<UNKNOWN, UNKNOWN>
```

Both already on the denylist, so no new type vocabulary was needed.

**Deliberately scoped:** the substitution lives in
`_sql_with_input_samples`, called *only* from the axis gate — never inside
`_infer_expression_type`. A sample value carries the *sample's* type, not the
runtime column's, so letting it leak would have (a) made the slice-class check
raise on an input-driven `value: ?{${ref(n).value}}[0]` and (b) flipped the
default-ordering decision when the sample happens to be numeric. Two unrelated
behaviour changes on a guess.

False-positive surface checked and pinned as tests:

```
multi .values on x         -> BUILT OK   (substitutes to `'west', 'east'` — two projections)
single .value on x         -> BUILT OK
input in a FILTER only     -> BUILT OK
expr around an input       -> BUILT OK   (upper(${ref(ycol).value}))
```

#### 3. `props.value` was neither claimed nor excluded (low)

The indicator scalar this module's own docstring opens with. Verified it is the
only top-level `value` prop across `visivo/schema/*.schema.json` (indicator,
isosurface, volume) and numeric in all three, so it joins the positional set —
with a test that fails if a future schema adds a `value` that legitimately
carries a record.

#### 4. `insight_execute_views.py` diagnostic surfacing had zero coverage (medium)

Confirmed by falsification: reverting line 140 to the pre-PR
`return jsonify({"error": message}), 400` left the whole suite green. This is
the path the Explorer's preview lane actually calls. Three endpoint tests added
(400 not blank-200, structured payload, additive-key negative).

#### 5. Wire shape: singular `diagnostic` → plural `diagnostics` list (medium)

`diagnostic.py` documents payloads as growing a **`diagnostics`** key and
`viewer/src/types/diagnostic.js::diagnosticsFrom` — the documented universal
reader — returns `[]` unless `Array.isArray(parsed.diagnostics)`. Confirmed no
Python producer emits `diagnostics` yet, so WB9 sets the precedent. Now plural
and a list, with a JS test that pins both directions:

```js
test('a singular `diagnostic` object would have been invisible', () => {
  expect(diagnosticsFrom({ error, diagnostic: body.diagnostics[0] })).toEqual([]);
});
```

#### 6. Viewer `DIAGNOSTIC_CODES` mirror (low)

Added `non_plottable_axis_type`, **and** added the guard that was missing:
`test_viewer_mirror_lists_every_diagnostic_code` reads
`viewer/src/types/diagnostic.js` and pins it against the Python dict (phases and
severities too), which is what `diagnostic.py`'s "keep the two in sync — the
shape is the contract" asked for and nothing enforced.

#### 7. `_infer_expression_type` rebuilt the MappingSchema per call (low)

Correct, and now memoized on the builder — both inputs (`self.models`,
`self.native_dialect`) are assigned once in `__init__` and never reassigned, so
once is right. Falsification shows the mechanism directly: without the cache a
two-prop insight builds the `MappingSchema` **3** times (default-sort probe on
x, then `props.x`, then `props.y`); with it, 1.

Plus two small ones from the same findings: the message said "a OBJECT" — the
article is now derived by a helper rather than hardcoded, so a future
vowel-initial addition can't reintroduce it; and the one alias-strip regex the
AST unwrap made redundant (`insight_query_builder.py:462`) is gone.

---

### Rejected

#### "ARRAY/LIST should be denylisted when unsliced" (low)

The premise is right — the exemption comment justified itself with the *sliced*
case only, and unsliced `x: ?{array_agg(...)}` really does bind an
array-of-arrays that a linear axis draws as nothing. But the proposed fix
(thread `props_slices` into the gate) would be **the same false positive as
OBJECT, in the other direction**: sqlglot spells DuckDB's genuinely-nested
`LIST` and Snowflake's `ARRAY` — a VARIANT constrained to arrays, delivered as
JSON text, categorical and perfectly plottable — with the *same*
`DataType.Type.ARRAY`, and nothing in `DataType.this` separates them. Splitting
sliced from unsliced doesn't help; the Snowflake half is a false positive in
both. ARRAY/LIST stay fail-open, the comment now records *that* reason instead
of the narrower one, and `test_an_unsliced_array_column_still_builds` pins the
case end-to-end so the decision is visible rather than assumed.

#### "An untracked `test_zz_probe.py` inflated the reported counts" (low)

Not reproducible. `git status --porcelain -uall` at `48c48f29` is **empty**, and
a pristine run of that commit gives exactly:

```
2577 passed, 2 skipped, 5 xfailed, 1 xpassed
All done! 544 files would be left unchanged.
```

— the same 2577 and the same 544 the PR body claims. Both numbers were correct
as published.

#### "73 allowlist parametrisations are deductively implied" (low)

The analysis is correct — that parametrisation feeds the classifier its own
member names and can only fail where the 1-line disjointness test already
fails. But the fix isn't deleting tests, it's adding the independent content
that was missing. The real over-rejection guards are the ones that run the
actual builder, so `test_real_columns_of_every_common_type_still_build` grew
from 7 cases to **24 SQL spellings** — deliberately spellings as a schema JSON
records them (`INTEGER`, `REAL`, `DOUBLE PRECISION`, `NUMERIC(10,2)`,
`TIMESTAMP WITH TIME ZONE`, …) rather than `DataType.Type` member names, since
that gap is exactly where a dead allowlist entry would hide — plus the
6-case semi-structured family, the unsliced array, and four input shapes. The
file's docstring now names which tests carry the weight and which only pin the
contract.

---

### Falsification evidence

Every new guard was falsified by stashing its fix and keeping the tests:

| stashed | failures |
|---|---|
| OBJECT back on the denylist | `test_semi_structured_columns_still_build[OBJECT]`, `test_semi_structured_types_are_classified_consistently`, `test_all_record_shaped_sqlglot_types_are_covered`, `test_the_excluded_record_types_are_allowlisted_not_merely_dropped` |
| input substitution removed | `test_doubled_query_string_referencing_an_input_fails_the_build` — *DID NOT RAISE*, `test_sample_substitution_does_not_leak_into_the_authored_message` |
| `value` removed from the positional set | `test_value_is_a_positional_prop`, `test_struct_on_an_indicator_value_fails_the_build` — *DID NOT RAISE* |
| article hardcoded back to `"a"` | `test_the_message_uses_the_right_indefinite_article[UNION]` |
| plural key → singular object | `test_returns_the_serialised_diagnostic_when_the_error_carries_one`, plus both view tests — `KeyError: 'diagnostics'` |
| memoization removed | `test_the_type_inference_schema_is_built_once_per_builder` — `AssertionError: 3` |
| execute view stops surfacing | `test_the_400_carries_the_structured_diagnostics` — `KeyError: 'diagnostics'` |
| `non_plottable_axis_type` removed from the JS list | `test_viewer_mirror_lists_every_diagnostic_code` (py) + 2 viewer tests |

### Gates

- `pytest tests/ -q` — **2622 passed**, 2 skipped, 5 xfailed, 1 xpassed (2577 before; +45 tests)
- `black --check --target-version py312 .` — clean, 544 files
- `yarn test --watchAll=false` — **7040 passed**, 366 suites
- `yarn lint` — 0 errors

No `visivo/models/` schema-affecting change, so no schema regeneration (the
`diagnostic.py` edit is a docstring/description string only) — no overlap with
#642.

PR #655, opened 2026-08-28
An object is only MODIFIED when its content changed (M12/M25)

jpjsonic:
## Problem

Two independent leaks made the editor's "what changed?" answer dishonest, and both are visible on a project nobody has touched.

**1. `objects_equal` compared bookkeeping, not content (M12).**
`ObjectManager.objects_equal` did `obj1.model_dump(exclude_none=True) == obj2.model_dump(exclude_none=True)` with no exclusions, so it compared `path` and `file_path`. Those are stamped on by the *parser*. Nothing a client sends can carry them — `ObjectManager._serialize_object` and `commit_views._build_child_info` both strip them on the way out, and the writer never puts them in YAML. So the two sides could never agree, **by construction**, for any client-saved object: re-POSTing the exact config the API had just returned reported `MODIFIED`, permanently, with no path back to `PUBLISHED`.

Measured on the real `test-projects/integration` fixture: **109 of its named objects reported MODIFIED after a no-op round trip**, and in every single case the only difference was `path` (plus `file_path` for objects that came from an include file).

**2. The input serializers injected non-config fields into every dump (M25).**
`SingleSelectInput` and `MultiSelectInput` unconditionally added `name_hash` (and, for multi-select, `structure`) to *every* `model_dump`. The parser sets `extra='forbid'`:

```
>>> SingleSelectInput(**SingleSelectInput(name="region", options=["East","West"]).model_dump())
pydantic_core.ValidationError: 1 validation error for SingleSelectInput
name_hash
  Extra inputs are not permitted [type=extra_forbidden, input_value='mgltrktcoqvinaelamjkalszbkkgb', ...]
```

So the thing those serializers produced was not a config the project could read back. Committing an input wrote YAML the very next parse refused — the deterministic mechanism behind the field report of a run leaving `test-projects/integration/project.visivo.yml` dirty with ~190 lines of injected `name_hash`. All 16 inputs in the integration fixture failed to re-validate from their own serialized config.

## Fix

**`objects_equal` excludes `{"path", "file_path"}`**, and the three sites that already knew this set now share one constant, `LOCATION_FIELDS`, instead of repeating it (`objects_equal`, `_serialize_object`, `commit_views._build_child_info` + the defaults block). What is not written cannot be what makes an object look changed.

**The input serializers gate their derived metadata**, the same way `ParentModel` gates `include_type`: pass `context={"include_input_metadata": True}` to `model_dump`/`model_dump_json` and `name_hash`/`structure` come back.

### Audit: does anything consume the CONFIG-level `name_hash` / `structure`?

No. This was an open question in the dossier; here is the answer, since it decides delete-vs-gate.

- **`name_hash` is a pure function of the name**, and everything that needs a hash computes its own. `run_input_job` writes the per-file hashes (`f"{input_hash}_options"`, `_defaults`) into the input's job JSON; `deploy_phase.process_inputs_async` sends `input_obj.name_hash()` explicitly on the upload record.
- **`structure` is a restatement of which of `options`/`range` is set**, and `run_input_job` derives it the same way (`run_input_job.py:520-524`) before writing it into the job JSON.
- **Viewer:** every `name_hash` read is off `files[]`, `models[]` or a job payload — `useInputsData.js:137` reads `optionsFile.name_hash`, not a config key. Every `structure` read (`useInputsData.js:70,147,149`, `inputJobs.js:15`) is off the `/api/input-jobs/` response, which `input_jobs_views` serves straight from those job files; `dist_phase` builds `data/inputs.json` from the same files.
- **Backend:** no `["name_hash"]` / `["structure"]` read anywhere in `visivo/` comes from a model dump.
- **core:** `name_hash` there is a `Job` model column populated from the deploy upload records above, not from the project config.

Deleting would therefore have been defensible. They are gated instead so the capability stays expressed and tested rather than needing archaeology to recover. **No caller passes the flag today** — that is deliberate, and stated here rather than hidden.

## Regression guard

New: `tests/integration/test_project_round_trip.py`, asserted against the **real** `test-projects/integration` fixture (copied to a temp dir first, so a regression here can never dirty the tracked tree — which is the exact failure being guarded). It states the property nothing tested:

- `test_the_fixture_actually_covers_the_interesting_types` — guards the guard; fails loudly if the fixture ever stops containing inputs.
- `test_every_published_config_is_a_valid_config` — every config the API hands out re-validates.
- `test_round_trip_reports_no_modifications` — every object stays `PUBLISHED`; `has_draft_changes()` is `False`.
- `test_round_trip_leaves_the_changes_endpoint_empty` — `/api/projects/<id>/changes/` agrees.
- `test_round_trip_does_not_dirty_the_yaml` — a Commit finds nothing to write and every YAML file is byte-identical.

Plus 5 unit tests on `objects_equal` (factory-built `SqlModel`s, including the complement: a real `sql` edit still reports `MODIFIED`) and 4 updated/new input-serializer tests.

## Falsification evidence

Each guard was falsified by reverting the implementation and keeping the tests.

**A — `exclude=LOCATION_FIELDS` removed from `objects_equal` (serializer gate intact):**

```
FAILED tests/integration/test_project_round_trip.py::test_round_trip_reports_no_modifications
  - AssertionError: objects reported changed by a no-op round trip:
FAILED tests/integration/test_project_round_trip.py::test_round_trip_leaves_the_changes_endpoint_empty
  - AssertionError: assert [{'name': 'lo...source'}, ...] == []
FAILED tests/integration/test_project_round_trip.py::test_round_trip_does_not_dirty_the_yaml
  - AssertionError: b'{"detail":"Chart name \'a-very-fibonacci-waterfall\' is not unique in the project",
    "error":"Commit would leave the project invalid."}
FAILED tests/server/managers/test_object_manager.py::...::test_path_only_difference_is_equal
FAILED tests/server/managers/test_object_manager.py::...::test_file_path_only_difference_is_equal
FAILED tests/server/managers/test_object_manager.py::...::test_get_status_is_published_after_a_no_op_round_trip
  - AssertionError: assert <ObjectStatus...D: 'modified'> == <ObjectStatus...: 'published'>
6 failed, 36 passed
```

Worth naming: with only this half broken, the commit is now *refused* (422) by #640's new validation gate rather than writing bad YAML — a whole-project rewrite re-injects inline-defined charts as top-level and collides. #640 turned a silent corruption into a loud refusal; this PR removes the reason the rewrite was attempted at all.

**B — serializer gate removed (`objects_equal` fix intact):**

```
FAILED tests/integration/test_project_round_trip.py::test_every_published_config_is_a_valid_config
  - AssertionError: configs the project cannot parse back:
FAILED tests/models/test_input.py::TestSingleSelectInput::test_with_query_options
FAILED tests/models/test_input.py::TestSingleSelectInput::test_static_options_include_name_hash_only_on_request
FAILED tests/models/test_input.py::TestMultiSelectInput::test_serialize_with_name_hash_only_on_request
FAILED tests/models/inputs/test_multi_select_coverage.py::TestStructureHelpers::test_serializer_omits_structure_by_default
ERROR tests/integration/test_project_round_trip.py::test_round_trip_reports_no_modifications
ERROR tests/integration/test_project_round_trip.py::test_round_trip_leaves_the_changes_endpoint_empty
ERROR tests/integration/test_project_round_trip.py::test_round_trip_does_not_dirty_the_yaml
  - pydantic_core._pydantic_core.ValidationError: 1 validation error for tagged-union[...SingleSelectInput...]
5 failed, 41 passed, 3 errors
```

## Test results

```
2476 passed, 2 skipped, 5 xfailed, 1 xpassed in 32.12s   (baseline on 7d31ea69: 2465 passed)
black --check --target-version py312 .  ->  544 files unchanged
```

Working tree clean after the run — the guard's own fixture copy means nothing writes to `test-projects/`.

## Notes

- **No schema regeneration needed.** `visivo/models/` was touched, but `python -m visivo.generate_project_schema_json` produces a byte-identical `visivo/src/visivo_project_schema.json` — a `@model_serializer` is not part of the validation schema. So **no overlap with open PR #642**, which regenerates that file.
- Read `git show 7d31ea69` (#640) and `git show dc4a2b88` (#639) first. This PR does not change the commit ordering or the `StandaloneFieldRefsValidator` gate #640 added; it only removes the reason a no-op commit was reaching that gate with the whole project marked dirty.
- Consistent with #640's `_renest_model_scoped_fields`: model-scoped metrics/dimensions round-trip through this guard too (`DimensionManager/new_x_rounded` and `MetricManager/composite_metric` in the fixture were both in the 109 and are both green now).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -2,6 +2,16 @@
 
 This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
 
+## Code Comments
+
+Default to no comment. Prefer a clearer name, an extracted function, or a test case that pins the behavior — those stay true as the code changes; a comment restating logic in prose does not.
+
+Add a comment only when:
+- the reasoning can't be inferred from the code or tests (a non-obvious external constraint, a workaround for a specific upstream bug), or
+- the logic is genuinely complex enough that a reader needs a map before diving in.
+
+When one is justified, keep it to a line or two — state the constraint, not the history of how it was found. This applies to Python and JS alike, and to comments already in a PR as much as new ones: when touching a file, trim existing comments down to this bar rather than leaving them.
+
 ## Development Commands
 
 ### Environment Setup
@@ -18,8 +28,6 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 - **Run CLI**: `visivo [command]`
 - **Ensure python file formatting**: ensure that all python files are formatted in accordance with our black settings
 - When editing the python in @visivo/ remember to run - `poetry run black .`
-- **Code comments**: Use comments sparingly. Avoid redundant comments that simply restate what the code does (e.g., `# Use preview-{insight_name} as run_id` followed by `run_id = f"preview-{insight.name}"`). Comments should provide context, explain "why" rather than "what", or document complex logic. Prefer clear variable names and function signatures with docstrings over inline comments.
-
 ### JavaScript Viewer Development
 - **Install dependencies**: `yarn install` (in `viewer/` directory)
 - **Run tests**: `yarn test`
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -112,6 +112,7 @@ nav:
   - Tutorials:
       - Overview: how_tos.md
       - Use Cases: use_cases.md
+      - Build a Dashboard: topics/dashboards.md
       - Sources: topics/sources.md
       - Source Types: topics/source-types.md
       - Semantic Layer: topics/semantic-layer.md
@@ -125,6 +126,7 @@ nav:
       - Annotations & Shapes: topics/annotations.md
       - CI/CD for BI: topics/ci-cd.md
       - Deployment: topics/deployments.md
+      - Static Distribution: topics/static-distribution.md
       - Telemetry: topics/telemetry.md
   - Concepts:
       - Overview: concepts/index.md
--- a/mkdocs/topics/dashboards.md
+++ b/mkdocs/topics/dashboards.md
@@ -1,0 +1,361 @@
+# Build a Dashboard
+
+A dashboard is the page your team opens — but it is the **last** object in a chain. Every
+dashboard in Visivo is assembled from the same five pieces, connected with
+[`${ref()}`](query-and-context-strings.md):
+
+```
+Source  →  Model  →  Insight  →  Chart  →  Dashboard
+(where     (a SQL     (data +     (wraps      (rows of
+data       query)     plotly      insights)   items)
+lives)                props)
+```
+
+This guide walks the whole path once, then covers the layout system — rows, items,
+widths, heights, and markdown — in depth.
+
+There are two ways to author everything on this page:
+
+- **YAML** — edit `project.visivo.yml` directly in your editor. Save, and
+  `visivo serve` hot-reloads the browser.
+- **The in-browser workspace** — run `visivo serve` and build visually: create sources
+  with the wizard, explore models, and drag objects onto the dashboard canvas. The
+  workspace writes the same YAML back to your project, so the two routes are always
+  interchangeable.
+
+## Step 1 — a Source
+
+A [Source](../concepts/source.md) is where your data lives — a database or file:
+
+```yaml
+sources:
+  - name: my_db
+    type: duckdb
+    database: target/example.duckdb
+```
+
+This guide uses DuckDB from here to the end: it is a single file, so there is nothing to
+start and every step below runs as written. See [Source Types](source-types.md) for
+PostgreSQL, Snowflake, BigQuery, SQLite, CSV, and the rest.
+
+## Step 2 — a Model
+
+A [Model](../concepts/model.md) is a SQL query saved against a Source:
+
+```yaml
+models:
+  - name: monthly_revenue
+    source: ${ref(my_db)}
+    sql: |
+      SELECT date_trunc('month', order_date::DATE) AS month,
+             SUM(amount) AS revenue
+      FROM orders
+      GROUP BY 1
+```
+
+!!! note "Write the SQL in the source's own dialect"
+
+    Visivo hands `sql` to the source unchanged — it does not translate between
+    databases. `date_trunc` and the `::DATE` cast above are DuckDB (and PostgreSQL)
+    syntax; against a `type: sqlite` source the same rollup is
+    `strftime('%Y-%m', order_date)`, and `date_trunc` fails outright with
+    `no such function: DATE_TRUNC`. Change a source's `type` and its models' SQL has to
+    move with it.
+
+    The `::DATE` cast is here because this walkthrough's `orders` table arrives as CSV
+    text (see the complete project at the end of Step 5). Drop it when `order_date` is
+    already a `DATE` column.
+
+## Step 3 — an Insight
+
+An [Insight](../concepts/insight.md) binds a Model's columns to plotly props with
+[query strings](query-and-context-strings.md):
+
+```yaml
+insights:
+  - name: revenue_by_month
+    props:
+      type: bar
+      x: ?{ ${ref(monthly_revenue).month} }
+      y: ?{ sum(${ref(monthly_revenue).revenue}) }
+```
+
+An insight knows what to draw and which data feeds it — but it does not know *where* it
+belongs on a page. That is the next step's job.
+
+## Step 4 — wrap it in a Chart
+
+A [Chart](../reference/configuration/Chart/index.md) wraps one or more Insights and
+carries the presentation-level layout (title, axes, legend):
+
+```yaml
+charts:
+  - name: revenue_chart
+    insights:
+      - ${ref(revenue_by_month)}
+    layout:
+      title:
+        text: Monthly Revenue
+```
+
+!!! warning "The Chart is the composition boundary"
+
+    A dashboard item never takes a bare insight. Items place **charts** (or tables,
+    markdown, and inputs) — so every insight must be wrapped in a chart before it can
+    appear on a dashboard. `chart: ${ref(revenue_by_month)}` will not validate;
+    `chart: ${ref(revenue_chart)}` will.
+
+    In the in-browser workspace this happens for you: dropping an insight onto the
+    canvas auto-wraps it in a new chart.
+
+Why the extra layer? Two things an insight alone cannot do:
+
+- **Combine** — a chart can render several insights on shared axes: a bar series plus a
+  line on a second y-axis, or an indicator layered over a trend. See the
+  [dual-axis example](../reference/configuration/Chart/index.md#dual-axis).
+- **Reuse** — the same insight can appear in many charts, each with its own title,
+  axis labels, and styling. The underlying query still runs once.
+
+```yaml
+charts:
+  - name: revenue_dual_axis
+    insights:
+      - ${ref(revenue_by_month)}      # bar, left axis
+      - ${ref(orders_by_month)}       # line, yaxis: 'y2'
+    layout:
+      title:
+        text: Revenue vs Orders
+      yaxis2:
+        overlaying: 'y'
+        side: right
+```
+
+## Step 5 — place it on a Dashboard
+
+A [Dashboard](../concepts/dashboard.md) arranges charts into **rows** of **items**:
+
+```yaml
+dashboards:
+  - name: my_dashboard
+    rows:
+      - height: medium
+        items:
+          - width: 1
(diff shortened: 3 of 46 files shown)


jpjsonic:

## Review fix pass — 8 of 9 findings were real, and I verified each one before touching it

Every finding was reproduced against the real `test-projects/integration` fixture first. Eight held up and are fixed in 46404229; one is rejected with reasoning below.

The review was right about the central thing, and the PR body's claim was wrong: **`exclude={"path", "file_path"}` is a top-level-only exclusion**, so the "every serializer strips them off again" / "what is not written cannot be what makes an object look changed" lines in the original body were false one level down. That is corrected here rather than papered over.

---

### Confirmed and fixed

**1. Nested location fields leaked everywhere (findings #2, #5)** — measured before the fix: `simple-dashboard`'s config carried 20 nested `path` values, `insights-dashboard` 72; a recursively-cleaned config saved back reported MODIFIED for all 6 dashboards; committing one `rows[0].height` change added 20 `path:` lines to `project.visivo.yml`.

Fixed with a shared `strip_location_fields` / `location_free_dump` in `object_manager.py`, used by `objects_equal`, `ObjectManager._serialize_object`, `InputManager._serialize_object`, `commit_views._build_child_info` and `run_views._resource_fingerprint` (the review counted 3 of 5 converted; it is 5 of 5 now).

> **The one place the suggested fix would have been a bug.** Finding #5 proposes "strip `path`/`file_path` recursively from the dump". A blind name-based walk is *not* safe: `layout.shapes[].path` and `layout.selections[].path` are real, user-authored Plotly properties — the SVG path of a drawn shape — and I confirmed them in `visivo/schema/layout.schema.json`. Stripping by key name would silently delete a valid chart's shapes from the YAML it commits: a valid config corrupted, which is strictly worse than the leak. The strip is therefore guided by the **model instance**, dropping a key only where it is a DECLARED field of the visivo model at that node; free-form prop extras are never touched. Two tests pin this, and rewriting the helper to go by key name fails both.

**2. `_parent_name` invisible to `objects_equal` (finding #1)** — confirmed, and it is a regression this PR introduced. Re-parenting `new_y_sum` from `another_local_test_table` to `daily_metrics_embedded` returned `status: "published"` with `to_publish: []` — the move silently dropped. `objects_equal` now compares it. Now returns `"modified"` and appears in `/changes/`.

**3. Defaults reported MODIFIED by presence, not by content (finding #3)** — confirmed: a no-op `GET`→`POST` of `/api/defaults/` produced a permanent phantom pending change.

> **The suggested fix would not have worked.** Finding #3 says to gate on `flask_app.project_manager.get_status()`. Those are two *different* caches: `POST /api/defaults/` writes `flask_app._cached_defaults`, and never touches `project_manager._cached_defaults`. `project_manager.get_status()` returned PUBLISHED because its own cache is empty, not because it compared correctly — delegating to it would have reported **every real defaults edit as PUBLISHED**. Fixed instead with `FlaskApp.defaults_changed()`, comparing cached against `project.defaults`, used at all four commit sites plus `has_draft_changes()`.

**4. A real defaults edit silently discarded (finding #4)** — confirmed exactly as described: 200, `published_count: 1`, cache cleared, `project.visivo.yml` byte-identical. `ProjectWriter._update` matches `value["name"] == child_name` and the `defaults:` mapping has no name. Added `SINGLETON_TYPE_KEYS` + `_update_singleton`, diffed in place so ruamel keeps comments. Also fixed the missing `mode="json"` the review flagged.

**5. The guard only tested the short-circuit (finding #6)** — correct, and the fidelity note was correct too. `round_tripped_app` now goes through **real HTTP** `GET`/`POST` on the same endpoints the viewer calls, carrying the `parentModel` sibling the GET attaches (without which the "round trip" silently un-scopes every model-local field). Added tests that commit real edits: a dashboard row height, a defaults change, an inline chart.

**6. Inline-chart commits blocked by a false duplicate-name error (finding #7)** — confirmed both ways. Pre-existing from #640, but it sits on this lane's path and blocks the strengthened guard, so it is fixed: `inject_cached_objects` now replaces an inline definition in place instead of appending a top-level duplicate.

**7. `InputManager._serialize_object` stripped only `path` (finding #8)** — correct. Note the review's own reason it was invisible is why my first falsification run *passed*: every fixture input lives in the root project file (`file_path=None`). Added `tests/server/managers/test_input_manager_serialization.py`, which sets a `file_path` and does fail against the old line.

---

### Rejected

**Finding #9 — `include_input_metadata` has no production caller.** The facts are right; I re-verified the grep. But there is no defect and the finding's own `failure_scenario` describes none. The lane's requirement was a *no-consumer proof stated openly rather than assumed*, and the PR body does exactly that ("No caller passes the flag today — that is deliberate, and stated here"). The two alternatives offered are worse: "wire the one consumer that justifies keeping it" means inventing a consumer, and deleting is a pure preference that would churn a PR already carrying four high-severity defects. Recorded as an acknowledged trade-off rather than silently kept.

---

### Falsification

Each guard was falsified by breaking one piece of the implementation with all tests kept.

| Reverted | Tests that fail |
|---|---|
| A — strip only the top level | 3 (`test_no_config_the_api_hands_out_carries_location_bookkeeping`, `test_a_real_edit_writes_the_edit_and_no_bookkeeping`, `test_serialize_object_strips_nested_bookkeeping`) |
| B — `objects_equal` ignores `_parent_name` | 3 (`test_re_parenting_a_metric_is_a_pending_change`: `assert 'published' == 'modified'`, + 2 unit) |
| C — `defaults_changed` → `is not None` | 1 (`test_saving_defaults_unchanged_is_not_a_pending_change`) |
| D — no singleton branch in `_update` | 5 (all 4 writer tests + `test_a_defaults_edit_is_actually_written`: `assert 8 == 11`) |
| E — injection appends inline drafts | 4 (incl. `Chart name 'fibonacci-plane' is not unique in the project`) |
| F — `InputManager` strips only `path` | 2 (`assert 'file_path' not in {...'file_path': '/Users/someone/repo/includes/inputs.visivo.yml'...}`) |
| G — **strip by key name** | 2 (`test_a_plotly_shape_path_is_not_mistaken_for_bookkeeping`: `KeyError: 'path'`) |
| H — serializers inject `name_hash` unconditionally | 8 + 3 errors (the original M25 half, still falsifying) |

### Gates

```
2505 passed, 2 skipped, 5 xfailed, 1 xpassed in 64.54s   (was 2476 on this branch, 2465 on main)
black --check --target-version py312 .  ->  547 files unchanged
python -m visivo.generate_project_schema_json  ->  byte-identical (still no overlap with #642)
```

Working tree clean after the run — the guard copies the fixture project, so nothing writes to `test-projects/`.

jpjsonic:

Merged `main` (through #670) and applied the new comment bar from #668.

**The #670 overlap.** #670 deletes the `name_hash`/`structure` injection from `SingleSelectInput`/`MultiSelectInput` outright. This branch had *gated* the same two fields behind `context={"include_input_metadata": True}`. Since no caller passes that flag — this PR's own description says so — and #670's audit found no consumer in either repo, the gate was a switch with nothing on the other end. Claude took main's deletion for both serializers and both of their test files, and dropped the gate. Production behavior is identical either way; #670's version is simply the smaller one.

What remains on this branch is the four things #670 does not touch: `objects_equal` excluding `path`/`file_path` at every depth, the `InputManager` serializer override, the `defaults_changed` + singleton-write fixes, and inline-definition replacement in `inject_cached_objects`.

`test_project_round_trip.py` no longer lists the serializer injection among the leaks it guards. `test_every_published_config_is_a_valid_config` stays, restated as the general `extra='forbid'` property rather than as this PR's half of a fix — it now guards #670's deletion too.

**Comment bar.** Claude trimmed every comment this PR adds or modifies: deleted the ones narrating the bug's history or restating the next line, and kept four, tightened to a line or two, each recording a constraint a reader cannot infer — `layout.shapes[].path` is a real Plotly property so the strip must be instance-guided rather than name-based; ruamel cannot represent the `Level` enums a Python-mode dump puts in `levels`; `parentModel` arrives as a sibling of `config`, not inside it; and the defaults mock must not answer the question under test. Docstrings were kept but had the how-we-found-it narration cut.

Claude deliberately left #670's own comments in `tests/models/inputs/test_multi_select_coverage.py` and `tests/models/test_input.py` untouched — they sat inside the conflict regions, but they are a day old and reviewed, and rewriting them here would be noise in a merge commit rather than a decision this PR is making.

**Gates:** `pytest tests/ -q` → 2527 passed, 2 skipped, 5 xfailed, 1 xpassed. `black --check --target-version py312 .` → 549 files unchanged. No viewer or `visivo/models/` files remain in this PR's diff, so no viewer suite and no schema regeneration.

PR #656, opened 2026-08-28
Explorer chart: fill the pane, and follow it when it moves (M28)

jpjsonic:
## Problem (M28)

The Explorer's live chart preview does not fill its pane, and does not re-lay-out when the pane changes size. Both halves were **measured in a real browser** (`getBoundingClientRect()` on the pane vs. Plotly's own resolved layout) before and after — never eyeballed.

**(a) Fill — broken flex chain.** `ExplorerChartPreview`'s root was `flex flex-1 min-h-0 flex-col`, but its parent (CenterPanel's chart pane) is a `display:block` div. `flex-1` is inert against a block parent, and there was no `h-full`, so the root's height was `auto`. Every percentage height below it — `ChartPreview`'s `h-full`, `ItemContainer`'s `h-full`, the Plot's `style.height: 100%` — then resolved against an indefinite parent and collapsed to `auto` too, leaving Plotly on its built-in **450px default**. Every earlier branch of the component already returned `h-full`, which is exactly why the loading/error/empty placeholders looked right and only the real chart was wrong.

**(b) Track — nothing observes the container.** Plotly re-measures its node only when told to, and the only thing that ever tells it is a window `resize` event: `react-plotly.js`'s `useResizeHandler` is a window listener, and plotly.js ships no `ResizeObserver` (upstream plotly.js#3984/#7059, react-plotly.js#41/#64). `CenterPanel` papers over its *own* dividers by dispatching a synthetic window resize when its split ratios change; nothing covers a pane that resizes for any other reason — a rail drag, a banner appearing above the plot.

## Fix

- `ExplorerChartPreview` root: `flex-1` → `h-full`.
- `Chart.jsx`: a `ResizeObserver` on the plot container that re-fires the window `resize` event, rAF-coalesced, skipped for a zero-size (hidden) box.
- `CenterPanel`: a `chart-preview-pane` testid so the story measures the pane instead of guessing at it.

### Staying consistent with #634 (3029258b)

#634 removed a *second* sizing authority (explicit `height` alongside `autosize`). This change adds a **trigger** for the authority that already exists, never a second one — and the observer is **gated on `layout.autosize`**: when a caller passes explicit width/height it owns the dimensions, and re-measuring the container there would reintroduce exactly what #634 removed. The dashboard passes both `height` and `width`, so it is not observed at all and its behaviour is unchanged. A unit test pins that gate directly.

## Falsification evidence

**Geometry, measured (sandbox :3001, integration project).** `pane / box / drawn`, where `drawn` is `gd._fullLayout` — what Plotly actually resolved:

| case | before | after |
|---|---|---|
| fill @1280px viewport | pane 509.14 / box 450 / drawn 450 — **59.1px short** | all within 2px |
| fill @1600px viewport | pane 260 / box 450 / drawn 450 — **190px of overflow** | all within 2px |
| right-rail drag (no window resize) | pane 454→354, drawn 452→442 — **87.8px stale** | all within 2px |

Both fill cases land on Plotly's literal 450px default in opposite directions — short at one layout, overflowing at the other.

**Each half is independently load-bearing** (e2e, impl mutated, tests kept):

- `h-full` reverted to `flex-1`, observer kept → the two fill tests fail with the *original* numbers (59.14 / 190); the track test still passes.
- Observer removed, `h-full` kept → the fill tests pass; the track test fails at **78.7px stale** (`pane 454→354, drawn 452→432.67`).
- Both present → 3/3 pass.

**Unit tests falsified** (`Chart.test.jsx`, 8 new tests):

- Effect deleted → **7 of 8 fail**: `observes the rendered plot container…`, `forwards a container resize…`, `coalesces a burst…`, `ignores a zero-size box…`, `disconnects the observer on unmount`, `cancels a pending frame on unmount`, `does not observe while the chart is still showing Loading` — e.g. `expect(received).toHaveLength(expected) / Expected length: 1 / Received length: 0 / Received array: []`.
- The 8th (`does NOT observe when the caller owns the size — #634 stays intact`) asserts an *absence*, so it survives that mutation by construction. It was falsified separately by removing the `plotlyOwnsSizing` gate: `Expected length: 0 / Received length: 1`.

## Test results

- `viewer`: **366 suites / 7045 tests pass**, `yarn lint` clean.
- New e2e story `viewer/e2e/stories/explorer-chart-fills-pane.spec.mjs`: 3/3 pass under its registered `exploration-mutations` project (it mints a real exploration and drags the workspace rail, so it is serial like its siblings).

### Pre-existing e2e failures — NOT dismissed, controlled for

A regression sweep (`dashboard-viewing`, `workspace-canvas-charts`, `exploration-preview`) shows 12 failures. I ran the **same sweep with this branch's `viewer/src` + config changes stashed** (i.e. at `7d31ea69`): **identical 12 failures, identical test names**. They are not caused by this change. Observed causes worth their own triage:

- `dashboard-viewing` Step 1 looks for `a[href="/lineage"]`, which the nav no longer renders — a stale spec.
- `workspace-canvas-charts` (4) and `exploration-preview` (7) exhaust the 30s **test** timeout at `expandSourceTable`'s source-toggle click; the button is present in the failure snapshot, so it is a budget/fragility problem in that helper, reproducible at `--workers=1`. The new story uses the same helper and passes in 6–9s, so it is not the helper alone.

## Files

- `viewer/src/components/items/Chart.jsx` (+ `Chart.test.jsx`)
- `viewer/src/components/explorer/ExplorerChartPreview.jsx`
- `viewer/src/components/explorer/CenterPanel.jsx`
- `viewer/e2e/stories/explorer-chart-fills-pane.spec.mjs`, `viewer/playwright.config.mjs`

No Pydantic models touched, so no schema regeneration — no overlap with #642.

**Cross-stream note:** the dossier assigns `Chart.jsx` to the Chart Presentation stream (I4) and suggested handing the observer over rather than editing it here. This lane was directed to ship both halves; the `Chart.jsx` change is 50 lines in one new effect plus a `ref` on `ItemContainer`, additive and gated, and should rebase cleanly — but I4 should be aware.

### DIFF
--- a/viewer/src/components/items/Chart.jsx
+++ b/viewer/src/components/items/Chart.jsx
@@ -136,7 +136,24 @@ const Chart = React.forwardRef(({ chart, projectId, itemWidth, height, width, sh
     // them anyway is not merely redundant — it hands Plotly two authorities, so
     // a re-render can paint at the fixed size before autosize strips it back.
     // Honour whichever the caller actually asked for rather than both.
-    if (!layout.autosize) {
+    //
+    // The dimensions can arrive from EITHER side — the `height`/`width` props
+    // (the dashboard's slot) or the chart's own authored `layout` — and until
+    // this branch stripped the second source, only the prop side was guarded.
+    // A config that carries `width`/`height` (e.g. integration's `surface-chart`:
+    // `{autosize: false, width: 500, height: 500}`) therefore reached Plotly as
+    // `{autosize: true, width: 500, height: 500}` once `hideToolbar` above
+    // forced autosize on, and Plotly resolves that contradiction in favour of
+    // the explicit values, silently: `initialAutoSize = (!width || !height) &&
+    // autosize` is false (plots.js), `plotAutoSize` only moves a dimension when
+    // `!layout.width` / `!layout.height`, and `Plots.resize` early-returns
+    // outright when `layout.width && layout.height`. So the chart stayed at its
+    // authored 500x500 in a preview pane of any other size AND ignored every
+    // resize sent to it. Exactly one authority reaches Plotly.
+    if (layout.autosize) {
+      delete layout.width;
+      delete layout.height;
+    } else {
       if (height !== undefined) layout.height = height;
       if (width !== undefined) layout.width = width;
     }
@@ -164,11 +181,24 @@ const Chart = React.forwardRef(({ chart, projectId, itemWidth, height, width, sh
   // Observing the container and re-firing that same window event adds a
   // TRIGGER for the sizing authority that already exists; it never introduces
   // a second one (#634 — "don't give Plotly two sizing authorities"). The
-  // corollary is the `plotlyOwnsSizing` gate: when the caller passes explicit
-  // width/height, `autosize` is off, the CALLER owns the dimensions, and
-  // re-measuring the container would be exactly the second authority #634
-  // removed — so we don't observe at all. (This is also why the dashboard is
-  // untouched: it passes both dimensions.)
+  // corollary is the `plotlyOwnsSizing` gate: when the layout carries explicit
+  // width AND height, the CALLER owns the dimensions, and re-measuring the
+  // container would be exactly the second authority #634 removed — so we don't
+  // observe at all. (This is also why the dashboard is untouched: it passes
+  // both dimensions.)
+  //
+  // The gate is plotly's OWN predicate, not a paraphrase of it. plots.js
+  // decides whether it is sizing itself in three places and all three test the
+  // dimensions, never the `autosize` key alone:
+  //   supplyLayoutGlobalDefaults: `coerce('autosize', !(layoutIn.width && layoutIn.height))`
+  //   plotAutoSize:               `!layout.width` / `!layout.height`
+  //   Plots.resize:               early-returns when `layout.width && layout.height`
+  // Gating on `!!layout.autosize` instead got BOTH sides wrong. It said yes to
+  // a chart Plotly refuses to resize (a config carrying width+height, whose
+  // forced `autosize: true` plotly discards — an observer whose every dispatch
+  // is a no-op that still wakes every window-resize listener in the app), and
+  // it said no to charts Plotly autosizes IMPLICITLY, which is any layout
+  // missing either dimension — the `autosize` key need never appear.
   //
   // Cost of using the window event rather than reaching into react-plotly.js
   // for the private per-plot handler: it wakes every other window-resize
@@ -177,7 +207,7 @@ const Chart = React.forwardRef(({ chart, projectId, itemWidth, height, width, sh
   // reflows, it is bounded to autosize charts, it is coalesced to one dispatch
   // per animation frame below, and it depends on no library internals.
   const containerRef = useRef(null);
-  const plotlyOwnsSizing = !!plotLayout.autosize;
+  const plotlyOwnsSizing = !(plotLayout.width && plotLayout.height);
   useEffect(() => {
     if (!plotlyOwnsSizing || isDataLoading) return undefined;
     const node = containerRef.current;
--- a/viewer/src/components/items/Chart.jsx
+++ b/viewer/src/components/items/Chart.jsx
@@ -1,6 +1,6 @@
 import Loading from '../common/Loading';
 import Plot from 'react-plotly.js';
-import React, { useState, useMemo, useImperativeHandle } from 'react';
+import React, { useState, useMemo, useImperativeHandle, useEffect, useRef } from 'react';
 import { ItemContainer } from './ItemContainer';
 import { itemNameToSlug } from './utils';
 import { chartDataFromInsightData } from '../../models/Insight';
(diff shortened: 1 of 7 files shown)


jpjsonic:

## Review fix pass — 3/3 findings real, all fixed (6f6da5ba)

I verified each finding in a real browser before acting rather than taking it on the reviewer's word. All three held up; one of them was *understated*, and one of them had a suggested fix that would have quietly deleted the guard.

---

### 1. `plotlyOwnsSizing` was inverted — and the layout it gated was self-contradictory (medium → confirmed, and worse than reported)

The gate tested `plotLayout.autosize`. plotly.js never asks that question. It asks **"is either dimension missing?"**, in all three places it decides:

| plots.js | predicate |
|---|---|
| `supplyLayoutGlobalDefaults` | `coerce('autosize', !(layoutIn.width && layoutIn.height))` |
| `plotAutoSize` | `!layout.width` / `!layout.height` |
| `Plots.resize` | early-returns when `layout.width && layout.height` |

**The false positive.** `surface-chart` in `test-projects/integration/project.visivo.yml` is `{autosize: false, width: 500, height: 500}` — every key schema-allowed. `Chart.jsx:125` forces `autosize: true` on top for `hideToolbar`, so plotly received `{autosize: true, width: 500, height: 500}`: two sizing authorities, which is precisely what #634 removed, and plotly resolves it in favour of the dimensions, silently.

Measured in the workspace chart view (`ChartPreview` → `<Chart hideToolbar>`) at 1600×1000, reading `gd._fullLayout` directly:

```
BEFORE  box 878×838   drawn 500×500   layout.width 500  layout.height 500  autosize true
AFTER   box 878×838   drawn 878×838   layout.width —    layout.height —    autosize true
```

So the fill half of M28 did not apply to this chart at all — a 500px figure floating in an 878×838 pane. And the observer *did* attach, firing a global `window` resize on every pane change that `Plots.resize` discards on its first line: every dispatch woke every window-resize listener in the app and moved nothing.

**The mirror miss.** A layout with no dimensions, or with only one, is autosized by plotly *implicitly* — the `autosize` key need never appear — and was never observed. Probes confirmed 0 observers for both. Latent today only because `Dashboard.jsx:574-575` always passes both.

**Fix, both halves.** The layout now carries exactly one authority: autosize on ⇒ no `width`/`height`, whichever side they came from. That extends the invariant already guarding the `height`/`width` **props** ("autosize and explicit dimensions are mutually exclusive", the #634 comment) to the **config**, which was the unguarded mirror. And the gate is now plotly's own predicate, `!(plotLayout.width && plotLayout.height)`.

Behaviour check on the other paths: the dashboard passes both dimensions, so it keeps `autosize: false` + the slot dims and attaches no observer — unchanged. A config's `width`/`height` is still honoured whenever nothing forces autosize (new test). And the config dims were *already* overridden by the slot dims on the dashboard before this PR, so stripping them in the preview is what makes `ChartPreview`'s stated contract — "a saved chart previews identically to how it renders on a dashboard" — actually true.

### 2. The story measured the chart against the pane, not against the chart's slot (low → confirmed)

`ExplorerChartPreview` renders its own chrome inside `chart-preview-pane` and above the chart: `PreviewInputControls` (a `p-3` strip, ~50px, whenever an insight references an input), the unresolved-inputs banner, the promoted-poll-failed banner. Any of them and both fill assertions fail by the chrome's height on a chart that is filling its slot exactly. The story passed only because it never creates an input.

**The suggested fix would have deleted the guard.** Anchoring on the slot alone: with the broken root the slot is auto-height *too*, so slot and plot agree at 450 and the assertion passes. Verified — reverting the root gives `slot 452 / box 450`, i.e. `|box − slot| = 2`, inside any sane tolerance.

So the pane stays in the chain, as an **edge** comparison: the slot must reach the pane's *bottom*. Falsified against the reverted root, and it still fails with the pane right there in the message:

```
✘ the plot fills its pane at 1280px
  Error: the chart slot reaches the pane's bottom edge — pane 509.140625 / slot 452 / box 450 / drawn 450
✘ the plot fills its pane at 1600px
  Error: the chart slot reaches the pane's bottom edge — pane 260 / slot 452 / box 450 / drawn 450
```

The reviewer's second point held too: the 2px tolerance was exactly consumed by `ItemContainer`'s 1px top+bottom border (border-box), so it sat at its limit with zero headroom. It is now a named `CHROME_PX + ROUNDING_PX` with the border accounted for in the comment.

New story test injects the chrome deliberately and pins both halves — the chart still fills its shrunken slot (Plotly's *resolved* height included), and the pane-anchored comparison is asserted to be the one that would have lied, so the test can't rot into a no-op. It doubles as vertical coverage for the M28 observer: the slot shrinks while the window does not.

### 3. The `flex-1` → `h-full` half had no guard CI runs (medium → confirmed)

`.rwx/test_visivo.yml` defines `viewer-test` (jest) and `viewer-lint` and **no Playwright story task**, so `explorer-chart-fills-pane.spec.mjs` only ever runs by hand. Nothing under `viewer/src/**` asserted the root's class, and coverage stays green either way because a class is a string, not a branch.

Added the assertion to `ExplorerChartPreview.test.jsx` — via a new `explorer-chart-preview` testid rather than `container.firstChild`, which trips `testing-library/no-node-access` and fails `viewer-lint`.

---

### Falsification (every guard proven to fail with its fix reverted, then restored)

| reverted | result |
|---|---|
| `plotlyOwnsSizing` → `!!plotLayout.autosize` | **2 failed / 31 passed** — "observes a chart with NO dimensions at all", "observes a chart given only ONE dimension" |
| width/height strip → `if (!layout.autosize)` | **2 failed / 31 passed** — "a config-supplied width/height does not survive the autosize hideToolbar forces", "observes that same config once hideToolbar hands sizing to the container" |
| root → `flex flex-1 min-h-0 flex-col` | jest **1 failed / 42 passed**; e2e **3 failed / 1 passed**, geometry above |

### Gates

- `yarn test --watchAll=false` — **366 suites / 7052 tests, all pass**
- `yarn run lint --max-warnings=0` — clean
- `explorer-chart-fills-pane.spec.mjs` against a sandbox on :3028/:8028 — **4/4 pass**
- Neighbouring stories (`dashboard-viewing`, `exploration-preview`, `exploration-chart-build-updates`, `chart-legend-layout`): 10 failures, **identical with these changes stashed** — a stale `a[href="/lineage"]` nav assertion, and Library drag-row pointer interception (`library-row-source-*-toggle` intercepted by its own draggable row, Tim's in-flight #647 surface). Neither is touched here.

viewer/src/components/items/Chart.jsx
```diff
@@ -148,12 +165,81 @@ const Chart = React.forwardRef(({ chart, projectId, itemWidth, height, width, sh
     [plotlyConfig]
   );
 
+  // M28 — make a CONTAINER resize reach Plotly.
```
tgsoverly, line 168: I am not sure I love all these slop comments, it really seems to be overkill.

tgsoverly: If it is in a test, that makes sense.

tgsoverly (approved):

PR #659, opened 2026-08-28
Instrument time-to-value: an ordered step-mark ladder for the first run (Guided First Run W1)

jpjsonic:
## The problem

The 2.1 exit gate is **"8 of 8 new users build a dashboard in under 20 minutes with zero hand-written YAML"**, and today nothing in the product can measure it.

- The onboarding shim emits *gesture* events — clicks, dismissals, checklist rows — but nothing that marks where a first run actually **is**.
- The CLI emits `new_installation` / `cli_command` under an anonymous `machine_id`; the viewer's events ship under a browser distinct_id. The two halves of a first run could not be joined, so no span crossed the CLI→browser boundary.

The number the release is gated on was not computable from anything we ship. This PR emits the ladder it is read off.

## The ladder

| `step_index` | event | surface | fires when |
|---|---|---|---|
| 1 | `first_run_launched` | `cli` | the local Flask server serves the viewer's `index.html` for the first time on this machine |
| 2 | `source_connected` | `viewer` | a source save succeeds — onboarding's connect step **or** `sourceStore.saveSource` anywhere else |
| 3 | `first_query_run` | `viewer` | the first query executes in the SQL editor |
| 4 | `first_model_created` | `viewer` | the first `modelStore.saveModel` succeeds |
| 5 | `first_insight_created` | `viewer` | the first `insightStore.saveInsight` succeeds |
| 6 | `first_dashboard_rendered` | `viewer` | a named dashboard with ≥1 item mounts on `/project/:dashboardName` — **terminal** |

Every mark carries `journey_id`, `step_id`, `step_index`, `ms_since_first_run`, `ms_since_previous_step`, `machine_id`, and `out_of_order`. The whole span is one subtraction, and the minutes are attributable to a specific step rather than to "onboarding".

The ladder was derived from the onboarding + workspace code as it stands, not invented: steps 2–5 sit at the same choke points the onboarding checklist already taps (`recordOnboardingAction`), but as separate marks — that helper **no-ops until onboarding is complete**, and the ladder has to measure users who skipped onboarding too.

## Three decisions worth arguing with

**1. `from_sample` on the terminal mark.** This is finding TTV-5, and without it the metric is worse than nothing: rendering the bundled example project takes ~1s, while rendering a dashboard built from the user's own data took the field testers 26–108 minutes. Optimising the wrong one would look exactly like success. **The gate metric is `ms_since_first_run` on `first_dashboard_rendered` where `from_sample = false`.** Read without that filter it measures nothing, and the taxonomy says so in bold.

**2. The journey lives in `~/.visivo/first_run.json`, not in memory.** A first run spans a CLI process, a browser, and every reload in between. "Once per session" would re-fire marks on reload, inflate the funnel, and destroy the median. The server hands the journey to the page as `window.__VISIVO_FIRST_RUN` — beside the existing telemetry-disabled flag in `data_views.py` — so the browser's marks carry the CLI's `journey_id` and `machine_id` and join to `new_installation` / `cli_command` in PostHog.

**3. `step_index` is frozen identity, not sort order.** W3's scaffold marks (`table_profiled`, `scaffold_shown`, `scaffold_applied`) will take the **next unused integers** even though they fall mid-journey chronologically. Renumbering would silently redefine every historical funnel. Funnels order by timestamp.

## Contract first

Per `visivo/CLAUDE.md`, PostHog is the single source of activation truth and the cross-repo contract is `specs/marketing-relaunch/event-taxonomy.md`. **The taxonomy edit landed first**, in the separate `specs` repo, on branch **`guided-first-run-ttv-taxonomy`** (commit `744a299`) — §4 gains a "Time-to-value step marks" section with the ladder, the required properties, the `from_sample` rule, the frozen-index rule, and the opt-out posture; §6 gains an implementation pointer. Events are `snake_case`, additive, and nothing existing is renamed — note `source_connected` is deliberately the name the VIS-879 role-flow list already reserved, not a new `first_source_connected`.

## Opt-out and privacy

With telemetry disabled (`VISIVO_TELEMETRY_DISABLED`, `~/.visivo/config.yml`, or project `defaults.telemetry_enabled: false`) the server writes **no ledger at all**, injects no journey, and emits nothing; the viewer checks `window.__VISIVO_TELEMETRY_DISABLED` again before touching the event buffer, `localStorage`, or PostHog. A disabled run emits nothing **and leaves no new file behind** — a stronger guarantee than "we write it but don't send it". Verified by test on both sides, including that the real `~/.visivo/first_run.json` is never created by the full pytest run.

Payloads are counts, booleans, and random UUIDs. No source / model / insight / dashboard / column / file / project name, no SQL, no path. `journey_id` and `machine_id` identify a first run, not a person. The CLI still never sends or stores email.

## Files

**New**
- `visivo/telemetry/first_run.py` — the journey ledger, the frozen `STEP_INDEXES`, `mark_step`, `viewer_journey_context`
- `viewer/src/components/onboarding/timeToValue.js` — `markTimeToValueStep`, emitting through the existing `fireEvent` shim (`surface: viewer`)

**Changed**
- `visivo/telemetry/events.py` — `FirstRunStepEvent` (`surface: "cli"`, `machine_id` as a property so the halves join)
- `visivo/server/views/data_views.py` — mark step 1, inject the journey
- `viewer/.../onboarding/posthogClient.js` — export the existing `isTelemetryDisabled` gate (one word)
- `viewer/.../onboarding/OnboardingFlow.jsx`, `stores/sourceStore.js`, `stores/modelStore.js`, `stores/insightStore.js`, `explorer/SQLEditor.jsx`, `project/Project.jsx` — the five call sites

No pydantic model changed, so no schema regeneration and no conflict with #642. No file on the hot-conflict list is touched.

## Tests

- `viewer/.../onboarding/timeToValue.test.js` — 23
- `viewer/src/stores/timeToValueMarks.test.js` — 11 (call sites; success-only, once-per-journey, no names)
- `viewer/.../project/Project.timeToValue.test.jsx` — 11 (terminal mark, `from_sample`, item counting)
- `tests/telemetry/test_first_run.py` — 21
- `tests/server/views/test_data_views_telemetry.py` — +6

**Gates:** `yarn test --watchAll=false` → **369/369 suites, 7098/7098 tests**. `yarn lint` → clean. `pytest tests/` → **2497 passed, 2 skipped, 5 xfailed, 1 xpassed**. `black --check --target-version py312 .` → clean.

### Falsification

Each guard was removed, the tests re-run, and the tree restored. Every one failed as it should:

| Guard removed | Result |
|---|---|
| once-per-journey ledger check in `mark_step` | `test_marks_the_step_once_and_only_once`, `test_idempotence_survives_a_new_process` fail; server-side `test_the_mark_fires_once_per_machine_not_once_per_page_load` fails with `assert 4 == 1` |
| opt-out check in `mark_step` / `get_or_create_journey` | 4 `TestOptOut` failures, incl. a journey being handed to the viewer under an opt-out |
| journey injection in `data_views.py` | 3 failures — `journey script tag missing` |
| once-per-journey check in `markTimeToValueStep` | 3 failures, incl. the viewer re-firing a step the CLI owns |
| opt-out early return in `markTimeToValueStep` | opt-out marks emitted **and** a ledger written to `localStorage`; store-level "telemetry disabled" test fails too |
| `from_sample` on the terminal mark | 2 failures — `Expected: true, Received: undefined` (the TTV-5 trap goes unguarded) |
| `Item.rows` recursion in `countDashboardItems` | `Expected: 3, Received: 2` |

Full output: `scratchpad/lanes3/falsify_ttv_out.txt`.

## Known limit, stated plainly

`first_dashboard_rendered` fires when the dashboard **mounts** with ≥1 item, not when a chart has painted data. The paint-level signal lives in `items/Chart.jsx`, which is owned by an open PR (#656); tightening this to a real first-paint is a one-line follow-up there. Mount-with-items is the honest available proxy and the taxonomy describes it as exactly that.

### DIFF
--- a/tests/server/views/test_data_views_telemetry.py
+++ b/tests/server/views/test_data_views_telemetry.py
@@ -20,6 +20,8 @@ from unittest.mock import patch
 
 import pytest
 
+from visivo.telemetry import first_run
+
 
 @pytest.fixture(autouse=True)
 def isolated_first_run(tmp_path, monkeypatch):
@@ -27,14 +29,22 @@ def isolated_first_run(tmp_path, monkeypatch):
 
     Without this, serving index.html in a test would write to the real
     ~/.visivo/first_run.json and send a real event.
+
+    A journey is only minted on an interactive run (the ``_is_interactive``
+    backstop that keeps containers/CI with a fresh $HOME from reporting a
+    brand-new first run on every cold start). pytest captures stdout, so these
+    tests have to say they are standing in for a human at a terminal.
     """
     monkeypatch.setenv("HOME", str(tmp_path))
+    monkeypatch.setattr("visivo.telemetry.first_run._is_interactive_run", lambda: True)
+    first_run.reset_journey_cache()
     monkeypatch.setattr(
         "visivo.telemetry.client.get_telemetry_client",
         lambda enabled=True: _RecordingClient.instance,
     )
     _RecordingClient.instance = _RecordingClient()
-    return _RecordingClient.instance
+    yield _RecordingClient.instance
+    first_run.reset_journey_cache()
 
 
 class _RecordingClient:
@@ -147,14 +157,55 @@ class TestFirstRunJourneyInjection:
         # And no ledger file is left behind on an opted-out machine at all.
         assert not (tmp_path / ".visivo" / "first_run.json").exists()
 
+    def test_the_injected_journey_carries_what_the_gate_metric_needs(
+        self, integration_client, monkeypatch
+    ):
+        """`install_age_ms` and `sample_dashboards` have to reach the browser.
+
+        Steps 2-6 are emitted by the viewer, so a property that only rides on
+        the CLI's step 1 cannot filter the terminal mark — which is the one the
+        exit gate is read off.
+        """
+        monkeypatch.delenv("VISIVO_TELEMETRY_DISABLED", raising=False)
+        journey = _injected_journey(_get_index_html(integration_client))
+
+        assert "install_age_ms" in journey
+        assert "College Football" in journey["sample_dashboards"]
+
     def test_injected_journey_cannot_break_out_of_the_script_tag(
         self, integration_client, monkeypatch
     ):
+        """Falsifiable version: a journey that actually contains `</script>`.
+
+        The real journey is UUIDs and integers, so asserting on it exercises
+        nothing — the escaping could be deleted outright and the test would
+        still pass. `sample_dashboards` made the journey carry strings for the
+        first time, so this now guards a live edge rather than a hypothetical.
+        """
         monkeypatch.delenv("VISIVO_TELEMETRY_DISABLED", raising=False)
+        monkeypatch.setattr(
+            "visivo.telemetry.first_run.viewer_journey_context",
+            lambda project_defaults=None: {
+                "journey_id": "j-1",
+                "started_at_ms": 1,
+                "install_age_ms": None,
+                "machine_id": "m-1",
+                "steps": {},
+                "sample_dashboards": ["</script><script>window.__pwned=1</script>"],
+            },
+        )
         html = _get_index_html(integration_client)
 
-        # The journey is JSON built from UUIDs and integers, but the escaping
-        # is what keeps it that way if a value ever changes shape.
-        script = re.search(r"<script>window\.__VISIVO_FIRST_RUN=.*?</script>", html)
-        assert script, "journey script tag missing"
-        assert script.group(0).count("</script>") == 1
+        # Read the assignment exactly the way a browser's tokenizer does: the
+        # <script> element ends at the FIRST literal `</script>`. Everything up
+        # to it must still be the complete journey. Without the escaping the
+        # element is cut mid-JSON and this raises.
+        marker = "window.__VISIVO_FIRST_RUN="
+        start = html.index(marker) + len(marker)
+        end = html.index("</script>", start)
+        raw = html[start:end]
+
+        assert "</script>" not in raw
+        assert json.loads(raw)["sample_dashboards"] == [
+            "</script><script>window.__pwned=1</script>"
+        ]
--- a/tests/server/views/test_data_views_telemetry.py
+++ b/tests/server/views/test_data_views_telemetry.py
@@ -1,18 +1,62 @@
-"""Tests for the viewer telemetry opt-out injection in data_views (VIS-843).
+"""Tests for the viewer telemetry injection in data_views.
 
-`visivo serve` must honor the CLI/local telemetry opt-out: when telemetry is
-disabled, the served index.html sets `window.__VISIVO_TELEMETRY_DISABLED=true`
-so the viewer's PostHog client never initializes. When telemetry is enabled,
-that flag must be absent so the viewer's default-on telemetry runs.
+Two things ride on serving index.html:
+
+  VIS-843 — `visivo serve` must honor the CLI/local telemetry opt-out: when
+  telemetry is disabled, the served page sets
+  `window.__VISIVO_TELEMETRY_DISABLED=true` so the viewer's PostHog client
+  never initializes, and when enabled that flag must be absent.
+
+  Guided First Run W1 — serving the page is the first moment the product is
+  in front of the user, so it is where the time-to-value ladder's step 1
+  (`first_run_launched`) fires, once per machine. The journey is then injected
+  as `window.__VISIVO_FIRST_RUN` so the viewer's marks (steps 2-6) carry the
+  same journey_id and machine_id and the whole span is one subtraction.
 """
 
+import json
+import re
 from unittest.mock import patch
+
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def isolated_first_run(tmp_path, monkeypatch):
+    """Keep the journey ledger and PostHog out of the developer's machine.
+
+    Without this, serving index.html in a test would write to the real
+    ~/.visivo/first_run.json and send a real event.
+    """
+    monkeypatch.setenv("HOME", str(tmp_path))
+    monkeypatch.setattr(
+        "visivo.telemetry.client.get_telemetry_client",
+        lambda enabled=True: _RecordingClient.instance,
+    )
+    _RecordingClient.instance = _RecordingClient()
+    return _RecordingClient.instance
+
+
+class _RecordingClient:
+    instance = None
+
+    def __init__(self):
+        self.events = []
+
+    def track(self, event):
+        self.events.append(event)
 
 
 def _get_index_html(integration_client):
     response = integration_client.get("/")
     assert response.status_code == 200
     return response.get_data(as_text=True)
+
+
+def _injected_journey(html):
+    match = re.search(r"window\.__VISIVO_FIRST_RUN=(\{.*?\})</script>", html)
+    assert match, f"no __VISIVO_FIRST_RUN in served html: {html[:400]}"
+    return json.loads(match.group(1))
 
 
 class TestViewerTelemetryInjection:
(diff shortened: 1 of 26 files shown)


jpjsonic:

## Fix pass — review findings, verified one by one (`a2594eab`)

I reproduced every finding against the real modules before touching anything. **Seven are real and fixed. Two are rejected**, with reasoning below. Every fix ships a test that fails when the fix is removed; the falsification table is at the bottom.

### Fixed

**1. `ms_since_first_run` was `0`, not `null`, in the cloud viewer.** With no injected journey, `resolveJourney()` minted `started_at_ms: Date.now()`, so the terminal mark reported a ~0ms time-to-value. Probe confirmed: `PROBE A ms_since_first_run = 0 machine_id = null`. The taxonomy specifies `null` "when the journey start is unknown (cloud viewer …)", and `core/app/src/App.jsx:50` does lazy-load this exact `Project` component with `posthogEnv.js` falling back to the hardcoded key — so cloud is always-on and its population is far larger than the CLI's. Locally-minted journeys now carry `started_at_ms: null`.

**2. Every established install looked like a first run.** The ledger's *absence* cannot mean "new" — no install has one until this ships. Probe with a machine_id aged a year: `fired: True`, `ms_since_first_run: 0`. The reviewer's suggested fix (gate on the machine_id file being absent) would have been **wrong**: `visivo init` and every other CLI command creates `machine_id` before `serve` ever runs, so it would suppress nearly every genuine first run. Instead every mark now carries **`install_age_ms`** — the age of `~/.visivo/machine_id` when the journey started — resolved *before* the stat so a brand-new install created by this very request reports ~0 rather than "unknown". And a journey is only **minted** on an interactive run (the same `machine_id._is_interactive` backstop that already gates `new_installation`), which closes the container/serverless-with-fresh-`$HOME` case in finding 16. Only minting is gated: a user who started at a terminal keeps being measured from a detached follow-up run.

**3. `from_sample` was about the onboarding branch, not the dashboard.** Confirmed wrong in both directions. The local server now names the bundled sample dashboards (`sample_dashboards`, read off `visivo/templates/samples`) on the injected journey, and the mark asks whether *this* dashboard is one of them. The onboarding-path fallback survives only in the cloud viewer, which has no server to ask.

**4. The terminal mark never fired for the cohort being measured.** `LocalRouter.jsx:151` routes `/workspace/dashboard/:dashboardName` to `Workspace` → `ProjectCanvas`, never `Project`, and `OnboardingFlow.completeAndNavigate` sends the *data* path — the non-sample users the gate is about — into the workspace. Extracted to a shared `useFirstDashboardRenderedMark` hook, mounted on both surfaces. The canvas wraps the same render-only `<Dashboard>`, so a user watching their first chart paint there has reached value; the mark is idempotent, so it counts once.

**5. An unwritable `$HOME` split the journey in two.** Probe with `_write_ledger` forced to `False`: three loads → `fired=True` three times, and on every load the emitted `journey_id` differed from the injected one, making the CLI and viewer halves unjoinable. Now one journey per process, cached by ledger path.

**6. The injected `machine_id` joined to nothing in CI/containers.** Probe with `CI=true`: event `ci-cf9ef66e…` vs injected `ci-418aa937…` vs a second injection `ci-5c6f8d2f…`. `viewer_journey_context` now reads the same process-wide memo the events use. The two tests that named this guarantee asserted only truthiness — they now assert **equality with the event's**.

**7. `first_query_run` fired on submit, not on return.** Moved to the `result` effect, behind a ref set in `handleRun` so a result this editor did not ask for cannot claim it. A failing/cancelled query no longer consumes step 3.

**Plus the two lows, and the vacuous test:**
- Marks are now **written back to the server ledger** (`POST /api/telemetry/first-run/step/`, which emits nothing — the browser already sent the event), so "once per journey" survives leaving the origin. Probe confirmed the bug: two `source_connected` under journey `J-1` after a ledger clear. `localStorage` is scoped to `http://localhost:<port>`; the ledger is not, and `viewer_journey_context` seeds the next page load from it.
- The ledger write is **atomic** (temp + `os.replace`) and the read-modify-write is **locked**, so two simultaneous `index.html` requests cannot both claim a step and a crash mid-write cannot silently mint a second journey.
- The `</script>` escaping test **was** vacuous — I confirmed the guard could be deleted with all ten passing. It now injects a journey actually containing `</script>` and reads the assignment the way a browser's tokenizer does (up to the first literal `</script>`, then `json.loads`). `sample_dashboards` also means the journey carries strings for the first time, so this now guards a live edge.

### Rejected

**`out_of_order` "reports false for a genuinely out-of-ladder-order journey."** Accurate as a description, not a defect. The contract defines it as "True when this mark's timestamp precedes the previous mark's — a clock-skew / double-count guard". It is a timestamp check, and the code matches the contract exactly. Funnels order by timestamp; `step_index` is identity, not sort order — that is stated in the taxonomy and in both implementations. I've made the doc say *explicitly* that it is not a ladder-order check so no one reads it as one. The related complaint that the three "in order" tests are tautologies is half right (`timestamps == sorted(timestamps)` is trivially true), but `indexes == [1,2,3]` does pin the frozen `step_index` assignment, which is worth keeping. I did not weaken or delete them.

**`first_run_launched` bypassing `is_ci_environment()`.** Partly addressed rather than taken as stated: I added the `_is_interactive()` backstop (which `machine_id.py`'s own comment calls the universal one, precisely because it needs no platform-specific signal), but did **not** suppress on `is_ci_environment()`. `is_ci` already rides on every mark as a §2 super property, and suppressing it would have broken the ability to see CI as data at all. The taxonomy now names the filters.

### Contract

The taxonomy was amended first, as before — specs branch `guided-first-run-ttv-taxonomy`, commit `fcee5e8`: `install_age_ms` added, the gate metric restated as `from_sample = false` **and** a young `install_age_ms`, `from_sample` documented as a fact about the dashboard, step 6 documented as firing on the Workspace canvas too, `out_of_order` documented as a timestamp guard, and the write-back documented under the once-per-journey rule.

### Gates

`pytest tests/` → **2525 passed**, 2 skipped, 5 xfailed, 1 xpassed. `black --check --target-version py312 .` → clean. `yarn test --watchAll=false` → **371 suites / 7127 tests**. `yarn lint` → clean. No file on the hot-conflict list touched; no pydantic model changed, so no schema regeneration and no conflict with #642.

### Falsification

Each guard removed, its tests re-run, the tree restored. Every one failed as it should:

| Guard removed | Result |
|---|---|
| `started_at_ms: null` on a viewer-minted journey | 2 failed — `a viewer-minted journey reports ms_since_first_run as null, not zero` |
| `install_age_ms` on the viewer mark | 3 failed — `the CLI-reported install age rides on every viewer mark` |
| the write-back call | 2 failed — `a cleared site-data mid-journey does not re-fire what already fired` |
| `from_sample` from the server-named samples | 3 failed — incl. `a sample-path user who then built their own dashboard reports false` |
| the terminal mark on the workspace canvas | 4 failed — `fires when a dashboard with items mounts on the canvas` |
| `first_query_run` waiting for the return | 2 failed — `does not mark first_query_run on submit — the contract says the query must RETURN` |
| the per-process journey cache | 3 failed — `assert 3 == 1`, and two journey-id mismatches |
| the interactivity backstop | 1 failed — `test_a_non_interactive_run_mints_no_journey_at_all` |
| the memoized machine_id | 1 failed — `assert 'ci-3e18d0af…' == 'ci-71466974…'` |
| `install_age_ms` on the CLI mark | 4 failed — `KeyError: 'install_age_ms'` |
| the atomic ledger write | 1 failed — `assert '' == '{"journey_id…'` |
| `record_viewer_step` | 1 failed — `KeyError: 'source_connected'` |
| the `</script>` escaping | 1 failed — `JSONDecodeError: Unterminated string` |

jpjsonic:

Claude merged `main` and applied the new comment bar from `visivo/CLAUDE.md` (#668) to this PR's diff.

**Conflict.** One file: `viewer/src/config/urls.js`. #669 added the `socketIo` key and this branch added `firstRunStep` at the same point in the `server` map. Both are kept — the derivation below nulls each of them for `dist`, so the dist-polling fix and the write-back endpoint each behave as their own tests expect. #666's artifact sweep touches no file on this branch and merged clean; it only walks `target/{insights,models,inputs}` and never goes near `~/.visivo`.

**Comments.** Deleted: restatements of the next line, narration of how the ladder was arrived at or which review finding it answers, and prose duplicating what a test name already pins — the `from_sample` trap in both directions, the `Item.rows` recursion, the jest no-op, the null `started_at_ms`. Kept and tightened to a line or two: the frozen `step_index` rule, `localStorage` being origin-scoped while the ledger is not, `get_machine_id()` not being idempotent in CI, the threaded-serve lock, claim-before-emit, `fetch` keepalive across a navigation, and the HTML tokenizer ending a `<script>` at the first literal `</script>`. Docstrings stayed; their bodies lost the narration. Nothing outside this PR's regions was swept.

**Gates on the merged tree.** `pytest tests/` → 2542 passed, 2 skipped, 5 xfailed, 1 xpassed. `black --check --target-version py312 .` → clean, 547 files. `yarn test --watchAll=false` → 371 suites, 7143 tests. `yarn lint` → clean.

**Opt-out re-verified end to end** after the merge: with telemetry disabled, `mark_step` / `get_or_create_journey` / `viewer_journey_context` / `record_viewer_step` all decline, nothing is tracked, the served page carries no `__VISIVO_FIRST_RUN`, and `~/.visivo` is not created at all. The enabled path still fires step 1 once across two page loads, hands the viewer the same `machine_id` the event shipped under, and keeps the ladder in order through the write-back.

tgsoverly (changes requested):
This seems like it is building the logic in the telemetry, rather than us deriving it from events.  Feels like we should be able to do this with just a few events.

PR #660, opened 2026-08-28
fix(viewer/canvas): the resize ghost tells the truth — derive the preview from the row total, not COLS=12 (M26 / VIS-1260)

jpjsonic:
## The bug

`Item.width` is a **relative weight**, not a column index in a fixed grid. `Dashboard.jsx` lays a row out as

```js
const totalWidth = visibleItems.reduce((s, i) => s + (i.width || 1), 0);
gridTemplateColumns: `repeat(${totalWidth}, minmax(0, 1fr))`,  gap: '0.7rem'
// each item: gridColumn: `span ${item.width || 1}`
```

`CanvasResizeLayer.jsx:58` hardcoded `const COLS = 12` and painted the ghost as `startBox * (liveCols / startCols)`. Two independent errors:

1. **The denominator moves.** Growing item 0 of a `6/6` row to 8 makes the row 14 wide, so the slot renders at 8/14 — not the 8/12 the old ghost drew.
2. **The gap is real.** `gap: 0.7rem` comes out of the track budget, so a span is not `rowWidth · span / total`.

The readout said `N / 12` on rows that were never 12 wide, and the 12 only ever looked right because every `ROW_TEMPLATES` entry happens to sum to 12.

## Measured, not eyeballed

Headless Chromium, real CSS grid built with Dashboard.jsx's exact row styles: run the gesture math, apply the committed widths, measure the rendered slot with `getBoundingClientRect`.

| row shape | drag | start→live | ghost AFTER | rendered | err AFTER | ghost BEFORE | err BEFORE |
|---|---:|---|---:|---:|---:|---:|---:|
| `[6,6]` | +70 | 6→8 (/14) | 566.64 | 566.63 | **0.01** | 659.21 | 92.58 |
| `[6,6]` | −120 | 6→4 (/10) | 393.28 | 393.28 | **0.00** | 329.60 | 63.68 |
| `[1,2]` | +150 | 1→2 (/4) | 444.40 | 444.41 | **0.01** | 585.06 | **140.66** |
| `[3,5,1]` | +140 | 3→5 (/11) | 539.34 | 539.34 | **0.00** | 654.22 | 114.88 |
| `[3,5,1]` | −200 | 5→3 (/7) | 507.90 | 507.89 | **0.01** | 397.02 | 110.87 |
| `[9,2]` | −33 | 9→7 (/9) | 619.72 | 619.73 | **0.01** | 507.50 | 112.23 |
| `[2,3]` | +90 | 3→5 (/7) | 682.53 | 682.53 | **0.00** | 952.55 | **270.02** |
| `[3,3,3,3]` | +55 | 3→5 (/14) | 221.39 | 221.38 | **0.01** | 252.68 | 31.31 |

```
worst |ghost − rendered| BEFORE = 270.02px
worst |ghost − rendered| AFTER  =   0.01px    (dossier acceptance: within 2px)
```

The dossier's own case reproduces exactly: 1000px `6/6` row dragged to width 8 → old ghost **666.7px**, drop **571.4px** (gap-free arithmetic); with the real 11.2px gap, 659.2 vs 566.6.

The formula itself was checked against the browser across 10 row shapes before anything was built on it: **worst |formula − rendered| = 0.0161px** (Chromium resolves layout on a 1/64px grid).

## The fix

**New `viewer/src/components/views/project/canvas/canvasGridGeometry.js`** — one owner for the sum-normalized formula, the identity CSS itself resolves:

```
track  = (rowWidth − (total − 1) · gap) / total
spanPx = span · track + (span − 1) · gap
total' = startTotal − startCols + spanCols     // a right-edge drag rebalances
total' = startTotal                            // a left-edge drag transfers
```

The layer measures the row's live `getBoundingClientRect().width` and its computed `column-gap` at pointer-down and derives every frame from those, so a zoomed/transformed canvas stays consistent.

Three changes in `CanvasResizeLayer.jsx`, as the dossier scoped them:

- **Truthful ghost** — `previewItemWidthPx(liveCols)`, expressed as a ratio against the item's own start slot so zero travel is pixel-identical to the card underneath.
- **Honest readout** — `N / <row total>`: the total a right-edge drag rebalances *to* (`8 / 14`), or the unchanged total for a left-edge transfer. Never `N / 12`.
- **One emphasized outline + a faded card** — the competing origin-geometry frame is deleted. During a gesture there is exactly one ring (the ghost, tagged `data-canvas-outline="emphasized"`), the card being resized fades to 35%, and only the handle being dragged is painted, riding the ghost's matching edge. Esc / pointercancel / unmount abort and un-fade — and pointercancel no longer *commits* an interrupted drag.

### One change beyond the dossier's three — please sanity-check this

The dossier names coarse quantization as a defect but does not list it among the three fixes. I made it a fourth, because a truthful ghost with the old mapping is a *new* legibility bug: `deltaCols = Math.round(dx / (rowWidth / total))` assumes each column is worth a constant number of pixels, which sum-normalization makes false. On a `[1,2]` row, one extra column costs 300px of travel but is worth only 152px of growth — so the honest ghost would visibly detach from the hand.

`nearestSpanForWidth` is the exact inverse of the preview: the span whose **rendered** edge lands nearest the pointer. Consequences worth a look:

- It kills the cited 225px dead zone (a 900px 2-item row now flips at ~75px), and near the top of a row's range it is *more* sensitive than before — dragging item 0 of a `[9,2]` row 33px left now commits `width: 7` where the old mapping needed 134px of travel to move at all.
- The reachable range is asymptotic (an item can never exceed `12 / (siblings + 12)` of the row), so drags past that pin at 12 — same saturation point as before, reached sooner.
- **Degenerate case handled:** on a solo-item row every span renders full-bleed, so the geometry cannot respond at all. Ties resolve toward the start span, making that gesture an honest no-op rather than a silent `width: 12` → `width: 1` rewrite. (Guarded both at the math and component level; mutation G below proves it.)

If you'd rather keep the old `dx → cols` mapping, it's a one-function swap — everything else in this PR stands.

## Tests

- **`canvasGridGeometry.test.js` (26 new)** — the primary guard. The dossier's headline numbers, browser-measured pixel fixtures, and a round-trip: for 8 row shapes × every item × every span 1–12, the preview must equal the geometry recomputed from the config `setItemWidth` **actually commits**. Two independent derivations; a drift in either fails.
- **`CanvasResizeLayer.test.jsx` (10 new)** — truthful ghost pixels and honest readout on a deliberately non-12 `[1,2]` row, ghost-vs-committed-layout agreement within 2px, exactly one emphasized outline, faded card + restore on release / Esc / pointercancel / unmount, solo-row no-op.
- **`canvas-resize-truthful-ghost.spec.mjs`** — new Playwright story. **Written, not run**: it rewrites a row's widths, so it is env-gated to its own isolated `:8060/:3060` sandbox and must never point at the shared one. Verified it parses and lists (5 tests).
- `WorkspaceDndContext.canvasCommit.test.jsx` — the existing full-stack drag updated. Its row is `9 + 2 = 11` wide, and its comment said "12 grid cols over an 800px row"; that mental model was the bug.

Full viewer suite **367 suites / 7087 tests green**, `yarn lint` clean.

### Falsification

Every guard was proven to fail with the implementation broken (tests untouched, tree restored after each):

| mutation | result |
|---|---|
| **A** `rowTotalForSpan` stops moving the denominator (the original bug) | **11 failed** — `THE finding: … renders 571px, not 667px`, `GHOST == DROP`, `TRUTHFUL PREVIEW`, `HONEST READOUT`, `a SOLO-item row…`, `a real width drag…commits` |
| **B** `gridTrackPx` ignores the 0.7rem gap | 4 failed — the browser-measured fixtures |
| **C** readout reverts to `N / 12` | 3 failed — `HONEST READOUT`, the left-edge transfer, the existing width-drag test |
| **D** a second emphasized outline at the pre-drag geometry | 1 failed — `exactly one emphasized outline` |
| **E** the card is never faded | 4 failed — both fade tests, Esc, unmount |
| **F** Esc / pointercancel stop aborting | 2 failed — `Escape ABORTS`, `pointercancel ABORTS` |
| **G** tie-break toward the start span dropped | 2 failed — `a SINGLE-item row is a no-op`, `a SOLO-item row never rewrites its width` |

Restored: 72/72 green.

## Notes

- Does **not** touch `Dashboard.jsx` (owned by #646). Its inline `totalWidth` reduce and `getWidth` are the second copy of this formula and should migrate onto `canvasGridGeometry` once #646 lands — a copy of the formula is exactly how the two drifted apart.
- No Pydantic models touched, so no schema regen.
- Second-wave verdict honoured: the relative-weight schema stays; only the surface written against the wrong mental model changed.

Linear: [VIS-1260](https://linear.app/visivo/issue/VIS-1260/w7m26-truthful-resize-ghost-derive-preview-from-row-totals-not-cols12)

### DIFF
--- a/viewer/e2e/stories/canvas-resize-truthful-ghost.spec.mjs
+++ b/viewer/e2e/stories/canvas-resize-truthful-ghost.spec.mjs
@@ -13,9 +13,16 @@
  *
  *   1. TRUTHFUL PREVIEW — the ghost's box at release and the slot the drop
  *      actually renders agree within 2px, on a row whose widths do NOT sum to 12.
+ *      Checked on a NON-FIRST item as well: sum-normalization shrinks the tracks
+ *      in front of the dragged slot too, so the slot MOVES as it grows, and a
+ *      ghost pinned at the item's measured left is right about size and wrong
+ *      about place for every item except the first.
  *   2. HONEST READOUT — the readout's denominator is the live row total.
  *   3. DIRECT MANIPULATION — exactly ONE emphasized outline is on screen during
- *      the gesture, and the card being resized is faded underneath it.
+ *      the gesture, and the card being resized is faded underneath it. The count
+ *      spans the whole overlay stack: the selection overlay's persistent ring
+ *      carries the same marker (src/.../canvasEmphasis.js) and stands down for
+ *      the gesture, so "1" here is a real property, not a private label.
  *   4. ABORT — Escape drops the gesture: nothing commits and the card un-fades.
  *
  * Precondition: an isolated sandbox running the integration project.
@@ -99,8 +106,15 @@ const pressAndDrag = async (page, handle, dx) => {
   return { sx, sy };
 };
 
+// Mirrors EMPHASIZED_OUTLINE_SELECTOR in
+// src/components/views/project/canvas/canvasEmphasis.js. Every STRONG (2px
+// mulberry) outline on the canvas carries it — the resize ghost AND the
+// selection overlay's persistent ring — so this counts what the user sees, not
+// one component's private marker.
+const EMPHASIZED = '[data-canvas-outline="emphasized"]';
+
 const countEmphasizedOutlines = page =>
-  page.evaluate(() => document.querySelectorAll('[data-canvas-outline="emphasized"]').length);
+  page.evaluate(sel => document.querySelectorAll(sel).length, EMPHASIZED);
 
 const opacityOf = (page, canvasPath) =>
   page.evaluate(
@@ -193,6 +207,56 @@ test.describe('canvas resize ghost == drop (M26 / W7)', () => {
     expect(Number(total)).toBe(liveTotal);
   });
 
+  test('the ghost LANDS where it is painted on a NON-FIRST item', async () => {
+    // Item 0 is the one index whose left edge cannot move (its offset is 0 by
+    // construction), so the truthful-preview check above cannot see a frozen
+    // ghost `left`. Item 1 can — and on a `[1, 2]` row that grows to `[1, 6]`
+    // the slot slides left by a third of the row while the ghost, pre-fix,
+    // stayed put and hung off the right edge.
+    await openCanvas(page);
+    await setRowWidths(page, [1, 2]);
+    await expect
+      .poll(async () => (await readRows(page))[0].items.map(i => i.width).join(','), {
+        timeout: WAIT,
+      })
+      .toBe('1,2');
+
+    const itemKey = 'row.0.item.1';
+    await selectKey(page, itemKey);
+    const handle = page.getByTestId(`canvas-resize-width-${itemKey}`);
+    await expect(handle).toBeVisible({ timeout: WAIT });
+
+    const rowBox = await page.locator('[data-canvas-path="row.0"]').first().boundingBox();
+    const before = (await readRows(page))[0].items[1].width;
+    await pressAndDrag(page, handle, Math.round(rowBox.width * 0.2));
+
+    const ghost = page.getByTestId('canvas-resize-ghost');
+    await expect(ghost).toBeVisible({ timeout: 4000 });
+    const ghostBox = await ghost.boundingBox();
+    // The ghost must stay INSIDE the row it previews — the frozen-left ghost
+    // was painted past the row's right edge on every last-item drag.
+    expect(
+      ghostBox.x + ghostBox.width,
+      `ghost right ${(ghostBox.x + ghostBox.width).toFixed(1)} vs row right ` +
+        `${(rowBox.x + rowBox.width).toFixed(1)}`
+    ).toBeLessThanOrEqual(rowBox.x + rowBox.width + 2);
+    await page.screenshot({ path: `${SCREENS}/m26-03-non-first-item.png`, fullPage: true });
+    await page.mouse.up();
+
+    await expect
+      .poll(async () => (await readRows(page))[0].items[1].width, { timeout: WAIT })
+      .not.toBe(before);
+    const slot = await page.locator(`[data-canvas-path="${itemKey}"]`).first().boundingBox();
+    expect(
+      Math.abs(slot.width - ghostBox.width),
+      `ghost ${ghostBox.width.toFixed(1)}px vs rendered ${slot.width.toFixed(1)}px`
+    ).toBeLessThan(2);
+    expect(
+      Math.abs(slot.x - ghostBox.x),
+      `ghost left ${ghostBox.x.toFixed(1)}px vs rendered ${slot.x.toFixed(1)}px`
+    ).toBeLessThan(2);
+  });
+
   test('exactly one emphasized outline, and the card under it is faded', async () => {
     await openCanvas(page);
     await setRowWidths(page, [1, 2]);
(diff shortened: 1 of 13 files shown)


jpjsonic:

## Review fix pass — the ghost now LANDS where it is painted

Pushed as `ce0495a1`. Five findings fixed, three rejected with reasoning. I verified every finding against the code before touching it; the two "high" ones were both real and both under-covered by the guards I had written.

### The headline: the width was truthful, the POSITION was not

The reviewer is right, and I measured it. Sum-normalization shrinks **every** track when the row total grows — including the tracks in *front* of the dragged item — so the slot slides LEFT as it widens. My ghost froze `left` at the item's measured box, which is correct only for the first item in a row. That is the single index every truthful-preview guard I wrote happened to use (`row.0.item.0`), and the e2e story's `Math.abs(slot.x - ghostBox.x) < 2` was also only ever run on item 0.

A measurement harness now ships with the PR — `viewer/e2e/tools/measure-canvas-grid.mjs`, plain `node`, no sandbox, no Playwright config involvement (not a `.spec.mjs`). It builds rows in headless Chromium from `Dashboard.renderRow`'s exact styles, measures every slot, and diffs against the module:

```
worst |formula − rendered| (static)       = 0.021591px
worst |ghost − rendered| (width)          = 0.022500px
worst |ghost − rendered| (left)           = 0.017857px
worst |frozen box.left − rendered| (left) = 592.140625px  <- pre-fix
```

That last line is the finding, in pixels, measured. It exits non-zero above 0.05px (two Chromium LayoutUnits).

### What changed

| # | Finding | Verdict | Fix |
|---|---|---|---|
| 1, 3 | ghost `left` frozen at `box.left` | **CONFIRMED** | `previewItemLeftShiftPx` + `spanLeftOffsetPx` + `precedingColsInRow` in `canvasGridGeometry`; the layer offsets the ghost by that shift |
| 2, 4 | selection ring competes with the ghost; the count guard is a tautology | **CONFIRMED** | `workspaceCanvasResizeKey` parks the ring for the gesture; `canvasEmphasis.js` makes the marker a shared contract both outlines stamp |
| 6 | solo-row width handle is inert | **CONFIRMED** | handle withheld (see the caveat below) |
| 5 | geometry never coupled to the renderer | **CONFIRMED** | `dashboardGridContract.test.jsx` renders the real `<Dashboard>` |
| 7 | "browser-measured" fixtures were the formula's own output | **CONFIRMED** | harness committed; fixtures replaced with the browser's verbatim readings |
| 8 | merge-ordering note on `WorkspaceDndContext.canvasCommit.test.jsx` | not a defect | acknowledged below |

**1/3 — the left edge.** Expressed as a SHIFT off the measured box, not an absolute left, on purpose: zero travel stays pixel-identical to the card underneath and any row padding / canvas transform cancels out. The one scale factor (`box.width / startSlotPx`) now applies to both axes. A left-edge TRANSFER still anchors the RIGHT edge — the total, and therefore every track, holds still there — and `previewItemLeftShiftPx` returns 0 for it by construction.

**2/4 — the two rings.** `CanvasSelectionOverlay` painted a full-strength 2px `ring-primary` at the pre-drag geometry for the entire gesture, in a sibling layer the card fade could never reach. It now stands down (with its hover hint) while `workspaceCanvasResizeKey` is set — a new ephemeral `workspaceStore` key the resize layer publishes on pointer-down and clears on release / Esc / pointercancel / **unmount**. `ProjectCanvas.jsx` is owned by #646 and is untouched, which is exactly why the signal goes through the store rather than a prop.

The old guard was a tautology, as reported: `data-canvas-outline="emphasized"` was a marker only the ghost could carry, so `toHaveLength(1)` could not fail. `canvasEmphasis.js` now owns that contract and **both** strong outlines stamp it; `canvasOutlineEmphasis.test.jsx` mounts BOTH overlays the way `ProjectCanvas` does and counts what is actually on screen. Subtle rings (hover hints, insertion bars) are deliberately not marked — the invariant is about the strong "this is the thing" outline.

**6 — the solo row. ⚠️ This is the one behavioural choice in the pass, flag it if you disagree.** On a one-item row `Σ widths` IS the item's width, so every span renders full-bleed and `nearestSpanForWidth`'s tie-break correctly refuses to move — but the handle, the ghost and a frozen readout were all still painted. I withheld the handle rather than keeping it live: a handle that provably cannot change a pixel is the same class of lie this lane exists to remove. The row keeps its (working) height handle, a container keeps its corner, and the item's stored weight is still editable in the right-rail Edit form, where it is a number rather than a promise about geometry. The `nearestSpanForWidth` tie-break stays as the backstop for any other caller. One line to revert if you'd rather keep the affordance.

**5 — geometry vs renderer.** You're right that the round-trip compared the module to itself. `dashboardGridContract.test.jsx` renders the REAL `<Dashboard>` (leaves mocked) and pins the grid *input* it emits — `repeat(Σ widths, minmax(0,1fr))`, `gap: 0.7rem`, `grid-column: span <width>` — to `rowGridTotal` / `precedingColsInRow` / `readColumnGapPx`. jsdom does not resolve grid layout, so the browser half of that claim is the committed harness; what jsdom *can* check is the thing that actually drifts. `Dashboard.jsx` itself is untouched.

**7 — the fixtures.** They were the formula's output rounded, as reported: `596.27` where the browser reads `596.2813`. They are now verbatim `getBoundingClientRect()` readings, LayoutUnit rounding included, compared under an explicit `BROWSER_EPS = 0.05` rather than a decimal `toBeCloseTo` precision that happened to pass. The `test.skip(!VIS_GHOST_BASE)` gate stays — it matches the repo's isolated-sandbox convention — but the numbers no longer depend on it.

**8 — merge ordering.** Agreed, not a defect and no owned path is touched. `WorkspaceDndContext.canvasCommit.test.jsx` re-runs green (22 tests) alongside the canvas suites; whichever of this and #646 lands second should re-run `npx jest --testPathPattern='WorkspaceDndContext.canvasCommit|views/project/canvas'`.

### Falsification — every new guard, impl stashed

```
=========== F1: ghost left frozen at box.left (the pre-fix behaviour) ===========
    ✕ a NON-FIRST item: the ghost moves left as it grows, and the drop matches it
    ✕ the LAST item of a 4-up row: a 440px lie becomes a 0px one
    ✕ SHRINKING a non-first item pushes the ghost RIGHT
Tests:       3 failed, 28 passed, 31 total

=========== F2: selection ring NOT suppressed during a gesture ===========
    ✕ DURING a resize it is the GHOST — the selection ring stands down
    ✕ a height gesture also owns the emphasis (the ghost spans the row)
    Received length: 2      <- the two competing rings, counted
    Received length: 2
Tests:       2 failed, 5 passed, 7 total

=========== F3: solo-row width handle painted again (dead affordance) ===========
    ✕ a SOLO-item row gets NO width handle: the gesture provably cannot move a pixel
Tests:       1 failed, 30 passed, 31 total

=========== F4: rowGridTotal pretends the row is a fixed 12 (renderer drift) ===========
    ✕ the row grid has Σ item.width tracks — exactly rowGridTotal, not a constant
    ✕ each item is `grid-column: span <width>`, so precedingColsInRow finds its start
    ✕ an unset width counts as one track on BOTH sides of the contract
Tests:       3 failed, 1 passed, 4 total

=========== F5: spanLeftOffsetPx drops the gap it straddles ===========
Tests:       2 failed, 39 passed, 41 total
--- the BROWSER harness on the same mutation ---
worst |formula − rendered| (static) = 145.588393px
FAIL: worst deviation 145.588393px > 0.05px

=========== F6: hover ring not suppressed mid-gesture ===========
    ✕ the HOVER hint also stands down — no stale ring chasing a dragging pointer
Tests:       1 failed, 6 passed, 7 total

=========== F7: unmount does not clear the resize key ===========
    ✕ unmounting mid-gesture releases the suppression rather than stranding it
Tests:       1 failed, 6 passed, 7 total
```

F5 is worth a note: the harness did **not** catch that mutation on the first attempt, because it re-implemented the left formula inline. I fixed the harness to call `spanLeftOffsetPx` — nothing in it re-implements the geometry now — and re-ran (`F5b`). A measurement tool that quietly duplicates the thing it measures is the same defect as finding 5, one level up.

### Gates

- `yarn test --watchAll=false` — **369 suites / 7118 tests green**
- `yarn lint` — **clean** (0 errors)
- `node e2e/tools/measure-canvas-grid.mjs` — exit 0
- E2E story extended (non-first-item ghost-left assertion, at-rest outline count, solo-row handle absence). Still gated on `VIS_GHOST_BASE`, so **not executed in this pass** — the isolated `:8060/:3060` sandbox was not stood up.

### Files

Touched: `canvasGridGeometry.js`(+tests), `CanvasResizeLayer.jsx`(+tests), `CanvasSelectionOverlay.jsx`, `workspaceStore.js`(+tests), `canvas-resize-truthful-ghost.spec.mjs`. New: `canvasEmphasis.js`, `canvasOutlineEmphasis.test.jsx`, `dashboardGridContract.test.jsx`, `e2e/tools/measure-canvas-grid.mjs`. No file owned by another open PR is modified.

PR #661, opened 2026-08-28
One codec owns `?{ }`, and the sort field stops teaching a form the parser rejects (M6 / M24 / C15)

jpjsonic:
## The bug

Add a Sort in the insight edit form's right rail, pick a field with `@`, type ` DESC`, Save. Visivo writes this into your YAML:

```yaml
interactions:
  - sort: ${ref(orders).month} DESC
```

and then refuses to load it, because `InsightInteraction.sort` is a `QueryString` — the parser accepts only `?{ ... }`. The UI produced YAML the tool rejects, and you get to debug it.

The same field advertised its example as `"date DESC"`. That fails twice over: unwrapped the parser rejects it, and wrapped, `?{date DESC}` reaches the binder as a loose identifier the semantic layer cannot resolve. People copied it.

## Why not just wrap on that one save path

Six sites each decided for themselves whether a value needed wrapping. One didn't wrap (M6). Another wrapped what was already wrapped (M24 — `?{?{ ... }}`, which renders as a blank chart). Fixing the save path alone would have moved the seam, not closed it.

## `expressionCodec.js` owns the grammar

```
decodeQueryString(v)      → { body, slice, form, repaired }
encodeQueryString(parts)  → the canonical `?{body}[slice]`, IDEMPOTENT
canonicalizeQueryString   → the composition; what save paths call
```

`decode` unwraps nested wrappers to a **fixpoint**, so a value an earlier double-wrapping write already corrupted decodes to the body the author meant and reports `repaired: true`. `encode` strips any wrapper the body already carries before wrapping, so it never nests.

**Idempotence is tested as a property, not as examples.** `SHAPES` in `expressionCodec.test.js` is a 42-entry table enumerated from the models — `?{ }` query strings, bare `${ref(...)}` context strings, `>{ }` eval strings (`EvalString`, a different grammar — passed through rather than wrapped into `?{>{ ... }}`), legacy `query()`/`column()`, every slice form (`[0] [-1] [1:5] [::2] [0,2]`), already-wrapped, doubly and triply wrapped, and the empties. Four properties run over all of it: canonicalize is idempotent (checked to three applications, so a two-cycle can't pass by accident), the output never nests a wrapper, body and slice survive re-decoding, and the result is a value a JS mirror of the Pydantic `QueryString` validator accepts.

`queryString.js` became a thin layer over the codec, so `serializeQueryString` is idempotent for all six of its existing callers — which fixes M24 in `PropertyRow.jsx` **without editing PropertyRow** (Tim's file).

Routed through the codec: `InsightBuildSection`'s two prop-drop sites and its interaction row, `explorerStore`'s "Add to exploration", `pillFieldSwap`, `saveAsMetricFlow`. The interaction row's old greedy `/^\?\{([\s\S]*)\}$/` never matched a *sliced* value, so `?{x}[0]` leaked its braces into the field and the next keystroke stored `?{?{x}[0]}`.

`expressionCodecOwnership.test.js` walks `viewer/src` and fails on any new ad-hoc `` `?{${...}}` `` literal. `WorkspaceDndContext.jsx` is allowlisted **with the reason in the file** — it is owned by open PR #646 and wraps correctly today; the entry says to remove it once #646 lands.

## The help text can't drift again

`viewer/src/schemas/interactionHelp.js` holds each field's `label`, `description`, `yamlExample` and `example`, copied verbatim from `InsightInteraction`. Two parity tests keep them there:

- `interactionHelp.test.js` re-derives all six strings from the vendored `visivo_project_schema.json` (`$defs/InsightInteraction`) — descriptions from the field descriptions, examples parsed out of the docstring's YAML block.
- `tests/models/test_interaction_help_parity.py` reads them straight off `InsightInteraction.model_fields`, so a **stale vendored schema cannot hide a drift either**. It also asserts each advertised example actually constructs an `InsightInteraction`, carries a real `${ref(...)}`, and that the bare form the old hint taught is still rejected.

Sort now reads: *"Column or expression to sort rows by; append ASC or DESC to control direction. Type @ to insert references, or drag them in from the library. For example: `${ref(orders).month} ASC`"* — a value that saves and runs. (The description is markdown on the docs site; the backticks are stripped for display only, so `INTERACTION_HELP` keeps the model's exact bytes.) `NULLS FIRST/LAST` is deliberately not advertised — the ORDER BY strippers still leak it (W8, VIS-1316).

Both editors now show the expression **body** and re-wrap on save, so the same user action produces byte-identical stored values in the right rail and the Explorer Build rail. The slice suffix is held aside in form state rather than dropped when you edit the body. The live preview is wrapped too — it runs the real `InsightInteraction` model, so a bare body there made the preview disagree with the saved insight for every interaction you typed rather than opened.

## Backend: the compiler's own words leaked into the UI

`/api/expressions/validate/` wraps your expression in `SELECT <expr> FROM __placeholder__` and swaps every `${ref(...)}` for an identifier sqlglot can read. When the parse failed it returned sqlglot's message verbatim:

```
Required keyword: 'expression' missing for <class 'sqlglot.expressions.GT'>. Line 1, Col: 28.
  SELECT __visivo_ctx__ > FROM __placeholder__
```

Three things there appear nowhere in your project. Now:

```
Required keyword: 'expression' missing for <class 'sqlglot.expressions.GT'>.
  ${ref(orders).amount} >
```

The context substitution numbers its tokens (`__visivo_ctx_0__`, `__visivo_ctx_1__`) so each ref maps back to the exact text you typed — a single shared sentinel made that impossible. The position marker is dropped because it counts characters in a statement you never wrote. `/translate/` gets the same treatment. This is W8(a) only; the bare-identifier diagnostic and the NULLS strippers live in `visivo/query/**`, which open PRs #653/#654 own.

## Falsification

Each guard was proven to fail with its implementation removed (impl stashed, tests kept):

| Reverted | Failures |
|---|---|
| `encodeQueryString` → unconditional wrap | **47** jest tests, incl. every `encodeQueryString itself is idempotent for …` case |
| `InsightEditForm` save-path wrap removed | **6** jest (`…saves only non-empty interactions`, M24, whitespace-drop, C15, remove, round-trip) |
| `"date DESC"` restored in `interactionHelp.js` | **6** jest + **4** pytest (`test_help_description_matches_the_model_field[sort]`, `…example_is_accepted_by_the_model[sort]`, `…names_a_real_ref[sort]`, `test_no_help_string_advertises_date_desc`) |
| `sanitize_expression_error` → ANSI-strip only | **12** pytest, incl. all 7 `test_validate_error_leaks_no_internal_sentinels[…]` |

## Gates

- `viewer`: `yarn test --watchAll=false` → **369 suites, 7314 tests, all passing**; `yarn lint` → **0 problems** (baseline was also 0, so the new files carry their own `no-template-curly-in-string` banners rather than adding noise)
- `pytest tests/ -q` → **2503 passed**, 2 skipped, 5 xfailed, 1 xpassed
- `black --check --target-version py312 .` → clean
- Verified in a live sandbox (isolated :8011/:3011) against `test-projects/integration`: the three interactions on `combined-interactions-test-insight` render as pills with no leaked braces, each hint ends with its working example, and Save stays disabled — the decode→encode round trip does not dirty an untouched insight.

## Ownership

No file owned by an open PR was edited. `PropertyRow.jsx`, `RefTextArea.jsx`, `PillMenu.jsx`, `inlineCreateStore.js`, `Library.jsx` (Tim) and `WorkspaceDndContext.jsx` (#646) are untouched — PropertyRow's double-wrap is fixed through `serializeQueryString`, and WorkspaceDndContext's one wrapper site is allowlisted with a note.

No Pydantic model changed, so no schema regeneration is needed (#642's file is untouched).

### DIFF
--- a/visivo/server/views/expression_views.py
+++ b/visivo/server/views/expression_views.py
@@ -7,6 +7,80 @@ from sqlglot import exp
 from visivo.query.patterns import CONTEXT_STRING_VALUE_PATTERN
 from visivo.query.sqlglot_utils import get_sqlglot_dialect, has_aggregate_function
 
+# ---------------------------------------------------------------------------
+# Error sanitisation
+#
+# Both endpoints below parse a bare EXPRESSION by wrapping it into a throwaway
+# statement — `SELECT <expr> FROM __placeholder__` — and `/validate/`
+# additionally swaps every `${ref(...)}` context token for an identifier
+# sqlglot can read. Those are compiler scaffolding. When the parse fails,
+# sqlglot's message quotes the wrapped SQL verbatim, so the user was shown
+# `__visivo_ctx__` and `__placeholder__` — tokens that appear nowhere in their
+# project and that they cannot act on — plus a "Line 1, Col: N" position
+# measured against a statement they never wrote.
+#
+# `sanitize_expression_error` puts the message back into the user's own words.
+# ---------------------------------------------------------------------------
+
+# Sentinel names. `_CTX_TOKEN_TEMPLATE` is numbered so each ref maps back to
+# the exact text the author typed rather than to a generic stand-in.
+_CTX_TOKEN_TEMPLATE = "__visivo_ctx_{index}__"
+_CTX_TOKEN_PATTERN = re.compile(r"__visivo_ctx(?:_\d+)?__")
+_PLACEHOLDER_TABLE = "__placeholder__"
+
+_ANSI_PATTERN = re.compile(r"\x1b\[[0-9;]*m")
+# sqlglot appends "Line 1, Col: 28." — a position in the WRAPPED statement.
+_POSITION_PATTERN = re.compile(r"\s*Line \d+, Col: \d+\.")
+# The `FROM __placeholder__` tail. sqlglot's TOKENIZER errors truncate the SQL
+# they echo, so the sentinel can arrive clipped ("__placeholder_"); match any
+# clipped form, but insist on the leading underscore so a real table called
+# `placeholder` is left alone.
+_HARNESS_TABLE_PATTERN = re.compile(r"(?:\s+FROM)?\s+_{1,2}placeholder_{0,2}")
+# The `SELECT ` head, at the start of an echoed line or just inside the quotes
+# of an "Error tokenizing '...'" message. An expression that legitimately began
+# with SELECT would lose that word from the echo — acceptable, because these
+# endpoints parse EXPRESSIONS, where a leading SELECT is already an error.
+_HARNESS_SELECT_PATTERN = re.compile(r"(?m)(^[ \t]*|['\"])SELECT\s+")
+
+
+def substitute_context_tokens(expression: str):
+    """Swap each `${...}` context token for a unique parseable identifier.
+
+    Returns ``(substituted_sql, {token: original_text})``. The tokens are
+    numbered rather than sharing one name so
+    :func:`sanitize_expression_error` can restore the author's exact text.
+    """
+    mapping = {}
+
+    def _replace(match):
+        token = _CTX_TOKEN_TEMPLATE.format(index=len(mapping))
+        mapping[token] = match.group(0)
+        return token
+
+    return re.sub(CONTEXT_STRING_VALUE_PATTERN, _replace, expression), mapping
+
+
+def sanitize_expression_error(error, token_map=None) -> str:
+    """Rewrite a sqlglot parse failure into a message about the user's text.
+
+    Strips ANSI decoration, substitutes every context token back to the ref the
+    author wrote, removes the `SELECT ... FROM __placeholder__` harness, and
+    drops the position marker (it counts characters in the wrapped statement,
+    not in the expression the user typed).
+    """
+    message = _ANSI_PATTERN.sub("", str(error))
+
+    for token, original in (token_map or {}).items():
+        message = message.replace(token, original)
+    # Anything the map missed still must not escape.
+    message = _CTX_TOKEN_PATTERN.sub("…", message)
+
+    message = _HARNESS_TABLE_PATTERN.sub("", message)
+    message = _HARNESS_SELECT_PATTERN.sub(r"\1", message)
+    message = _POSITION_PATTERN.sub("", message)
+
+    return "\n".join(line.rstrip() for line in message.split("\n")).strip()
+
 
 def register_expression_views(app, flask_app, output_dir):
     """Register expression translation API endpoints."""
@@ -64,7 +138,7 @@ def register_expression_views(app, flask_app, output_dir):
                     continue
 
                 try:
-                    wrapped_sql = f"SELECT {expression} FROM __placeholder__"
+                    wrapped_sql = f"SELECT {expression} FROM {_PLACEHOLDER_TABLE}"
                     parsed = sqlglot.parse_one(wrapped_sql, read=read_dialect or "duckdb")
 
                     select_expr = parsed.expressions[0] if parsed.expressions else None
@@ -99,7 +173,10 @@ def register_expression_views(app, flask_app, output_dir):
                     errors.append(
                         {
                             "name": name,
-                            "error": str(e),
+                            # Same rule as /validate/: the message reaches the
+                            # UI, so it may not carry the `__placeholder__`
+                            # harness this endpoint wrapped the expression in.
+                            "error": sanitize_expression_error(e),
                         }
                     )
 
@@ -157,17 +234,18 @@ def register_expression_views(app, flask_app, output_dir):
                     results.append({"name": name, "valid": False, "error": "Empty expression"})
                     continue
 
-                substituted = re.sub(CONTEXT_STRING_VALUE_PATTERN, "__visivo_ctx__", expression)
+                substituted, token_map = substitute_context_tokens(expression)
                 try:
                     sqlglot.parse_one(
-                        f"SELECT {substituted} FROM __placeholder__",
+                        f"SELECT {substituted} FROM {_PLACEHOLDER_TABLE}",
                         read=read_dialect or "duckdb",
                     )
                     results.append({"name": name, "valid": True})
                 except Exception as e:
-                    # sqlglot decorates messages with ANSI underlines — strip
-                    # them; this error renders in the viewer UI.
-                    message = re.sub(r"\x1b\[[0-9;]*m", "", str(e))
+                    # This error renders in the viewer UI, so it must talk about
+                    # the expression the author wrote — not about the harness
+                    # that wrapped it (M13).
+                    message = sanitize_expression_error(e, token_map)
                     results.append({"name": name, "valid": False, "error": message})
 
             return jsonify({"results": results}), 200
--- a/visivo/server/views/expression_views.py
+++ b/visivo/server/views/expression_views.py
@@ -20,6 +20,24 @@ from visivo.query.sqlglot_utils import get_sqlglot_dialect, has_aggregate_functi
 # measured against a statement they never wrote.
 #
 # `sanitize_expression_error` puts the message back into the user's own words.
+#
+# The mechanism matters. sqlglot does not echo the whole statement — it echoes a
+# WINDOW of it, clipped at both ends:
+#
+#   tokens.py:1085   start = max(self._current - 50, 0) ... context = self.sql[start:end]
+#   parser.py:1722   start_context = self.sql[max(start - error_message_context, 0):start]
+#
+# So string surgery on the echoed text cannot work: a sentinel that the window
+# cuts through arrives as a FRAGMENT (`__visivo_ctx_0__` -> `isivo_ctx_0__`, or
+# `x_1__`, or `_0__`), which no amount of exact replacement will catch, and any
+# pattern loose enough to catch every fragment is loose enough to eat the
+# author's own text. The earlier version of this function did both: it leaked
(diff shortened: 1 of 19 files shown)


jpjsonic:

## Review fix pass — 8c03876

Eleven findings. **Seven reproduced and are fixed**; **four are answered without a code change** (reasoning below). One finding turned out to be two, and its root cause was in the backend, not the codec.

---

### 1 · The codec wrapped what it could not parse — CONFIRMED (2 findings, high)

`decodeQueryString` classed anything it failed to unwrap as a `bare` body, and `encodeQueryString` wraps bare bodies. Verified against the shipped module under node:

```
canonicalize('?{a = 1\nAND b = 2}')
  -> '?{?{a = 1\nAND b = 2}}'
  -> '?{?{?{a = 1\nAND b = 2}}}'
  -> '?{?{?{?{...}}}}'
```

One layer per application, unbounded. Same for `?{}` → `?{?{}}`, `?{x}[a]` → `?{?{x}[a]}`, `?{a}[0][1]`, `?{unbalanced`. Each of those is a value the `QueryString` validator **refuses today** and each became one it **accepts**, with the braces reaching SQL as literal text — a loud rejection traded for silent nonsense. And it was reachable without touching the field: open an insight whose filter is multi-line, change only the **description**, save.

Two complementary changes:

- the codec's wrapper detector is now `[\s\S]` where the parser is `.`, so a stored multi-line value is recognised as the wrapper it is and unwrapped for editing — restoring the newline-awareness of the `/^\?\{([\s\S]*)\}$/` this PR replaced in `InsightBuildSection`;
- a value that is wrapper-**shaped** but unparseable decodes as `malformed`, and `encodeQueryString` hands it back **byte-identical**. `?{}` / `?{  }` / `?{?{}}` decode as empty and are dropped.

The codec is now **total and non-destructive**: it either produces a canonical value whose body the parser reads back unchanged, or it changes nothing.

The two fixes catch different things, which is the point:

| mutation | jest failures |
|---|---|
| revert the newline tolerance | 3 |
| revert the malformed passthrough | 26 |
| revert the empty-wrapper handling | 2 |

### 2 · The editor now says so, rather than writing it out — CONFIRMED (medium)

Totality means `encodeQueryString` can hand back something the parser will refuse. The save path may not just write that. `interactionValueProblem` names the reason and `InsightEditForm` blocks the save with it under the field, retracting the complaint as soon as the author edits:

- `>{ … }` → "Eval strings (>{ … }) belong to tests and alerts. Use a ?{ } expression here."
- an interior newline → "An interaction expression must be a single line."
- `?{x}[a]` → "This is not an expression the parser can read. Check the ?{ } wrapper and any [slice] suffix."

Removing the guard fails 4 tests.

### 3 · The gate and the reader disagreed — CONFIRMED, and this was the root cause

Chasing the multi-line finding downstream: `QueryString.validate_and_create` accepted anything starting `?{` and ending `}`, but `get_value()` reads the body with `QUERY_STRING_VALUE_PATTERN`, which is narrower. Values in the gap — `?{}`, `?{  }`, any interior newline — validated, were written to YAML, and returned `None` much later, where `InsightInteraction.field_values_with_js_template_literals` hands them to `re.sub` and the run dies with `TypeError: expected string or bytes-like object, got 'NoneType'`.

The gate is the pattern now (plus a non-blank body), and the message names which of the four ways the value missed:

```
?{}        -> QueryString '?{ }' needs an expression between the braces. Got: '?{}'
?{a\nb}    -> QueryString '?{ }' expressions must be on a single line — a newline
              inside the braces cannot be parsed. Got: '?{a\nb}'
?{x}[a]    -> QueryString must be '?{ expression }', optionally followed by an
              [index|slice] suffix such as [0], [1:5], [::2] or [0,2]. Got: '?{x}[a]'
date DESC  -> QueryString must be wrapped in '?{ }' (optionally followed by an
              [index|slice] suffix). Got: 'date DESC'
```

Full pytest suite green afterwards — nothing in the repo was relying on the gap. Reverting the gate fails 14 tests. `visivo_project_schema.json` **regenerates unchanged**, so no conflict with #642.

### 4 · The sanitiser leaked clipped sentinels and rewrote the user's text — CONFIRMED (2 findings)

Both halves reproduce through the Flask test client:

```
"1 + 2 + 3 + ${ref(m).c} + 'unterminated"
  -> "Error tokenizing 'isivo_ctx_0__ + 'unterminated'"          LEAKED
"case when ${ref(orders).status} = 'SELECT foo' then 1 end +"
  -> "... case when ${ref(orders).status} = 'foo' ..."           REWROTE THE USER
```

The mechanism is that sqlglot echoes a **window** of the SQL, clipped at both ends — `tokens.py:1085` takes ±50 chars, `parser.py:1722` takes ±`error_message_context`. A sentinel the window cuts through arrives as a fragment (`isivo_ctx_0__`, `x_1__`, `_0__`, `LECT `), so exact replacement cannot catch it and no pattern loose enough to try is safe from eating the author's own text.

So the fix is not a better pattern: **the echoed SQL is replaced with the expression the author typed**, which we already have. The harness never gets the chance to appear. The `SELECT` strip is now anchored to a line start or to an `Error tokenizing` opening quote, never to an arbitrary quote.

New guard sweeps expression lengths so the ±50 window falls at many offsets, asserts the author's text is echoed **in full**, and asserts that once it is removed the residue contains **no tail of any harness token** (every suffix of `__visivo_ctx_N__`, `__placeholder__`, `SELECT`, length ≥ 2). Reverting the sanitiser fails 23 tests.

### 5 · The shape table was calibrated to pass — CONFIRMED (high)

Reproduced: the omitted shapes fail the **unmutated** previous implementation. Added 15 — multi-line bodies (including CRLF and a padded-with-newlines one), the empty wrapper, `?{?{}}`, and five non-slice bracket suffixes.

Two properties were also weak and are replaced:

- the acceptance mirror was `startsWith('?{') && endsWith('}')`, which accepts `?{?{x}}` — it could not have failed on the bug the file is named for. It is now `isParserReadableQueryString`, the exact mirror of the Python gate.
- a new property asks the **parser** what body it reads out of the codec's output and compares it to the body the codec claims to have preserved. Double-wrap and the parser reads `?{x}` where the codec decoded `x`.

### 6 · The ownership guard covered one spelling of four — CONFIRMED (low, filed twice)

Verified: `` `?{ ${body} }` `` (the spaced form the project YAML and `InsightInteraction`'s own docstring publish), `'?{' + body + '}'`, `"?{" + …`, and `` `${prefix}?{${body}}` `` all passed. Four rules now, each with explicit catches / does-not-fire cases, including the strip side the PR body claimed but did not guard.

The new strip rule found a **real offender**: `pillFieldSwap.js`'s own `/^\?\{([\s\S]*)\}$/`, now routed through `parseQueryString` (a sliced slot stays out of scope exactly as the old regex left it — the swap rewrites to a bare `?{${ref(name)}}` and would drop the slice). Restoring that regex fails the guard by file and line.

### 7 · Cross-editor parity was prose — CONFIRMED (low)

`viewer/src/components/views/common/interactionSaveParity.test.jsx` renders **both** editors, drives the same edit through each real `RefTextArea`, and compares the stored strings over six shapes (sliced, range-sliced, padded, double-wrapped, self-wrapped). The one place they legitimately differ — the right rail drops an emptied interaction from the config, the Build rail holds `''` while the author is mid-edit — is now stated as a decision with its own test rather than left to drift.

---

## Rejected

**"Three of the four idempotence properties are structurally incapable of failing."** The mutation chosen (unconditional wrap inside `encodeQueryString`) is indeed invisible through `encode ∘ decode`, but the conclusion does not follow. Those properties fail on the shapes that were *missing from the table*: the multi-line entry alone fails `canonicalize is idempotent`, `a canonical value re-decodes to the same body`, and `encode(decode(x)) never nests a wrapper` against unmutated code. The table was too narrow, not the properties tautological — and widening it is the same remedy as finding 5, which is where it is handled.

**The eval-string passthrough is "still M6."** Typing `>{ … }` into a Sort field is a user error, not a normal action silently producing invalid YAML. Manufacturing `?{>{ … }}` so it validates would be strictly worse. Refusing it **at the field** is the fix, which is what the editor now does — so the invariant is closed without the codec lying about its contract.

## Left alone, deliberately

- `?{?{x}}` still passes the Python gate — its body parses, as the literal text `?{x}`. Rejecting nested wrappers at the model would fail projects that today merely render a blank chart, and the viewer now repairs them on decode. Worth a separate decision.
- Legacy `query(…)` / `column(…)` bodies are still wrapped by `encodeQueryString` rather than passed through, which contradicts the module's own "neither rewrites nor rejects them". Unreachable for interactions (the validator refuses them outright) and out of scope for these findings.

## Gates

`yarn test` 7514 / 370 suites green · `yarn lint` clean · `pytest tests/` 2558 passed, 2 skipped, 5 xfailed · `black --check --target-version py312 .` clean · schema regenerates unchanged.

visivo/server/views/expression_views.py
```diff
@@ -7,6 +7,137 @@
...
+    return re.sub(CONTEXT_STRING_VALUE_PATTERN, _replace, expression), mapping
+
+
+def _scrub_harness(text: str, token_map=None) -> str:
+    """Best-effort removal of harness text from a message part.
+
+    Used on the DESCRIPTION half of a message — never on the author's own
+    expression, which is copied through verbatim.
+    """
+    for token, original in (token_map or {}).items():
```
tgsoverly: I am not following whey this area is needed, seem odd to me that we would need to remove stuff like this.  Can we avoid it al together?

tgsoverly (approved):

PR #658, opened 2026-08-28
Quick Start: lead with the command that actually works (C46/C79)

jpjsonic:
Linear: [VIS-1271](https://linear.app/visivo/issue/VIS-1271/quick-start-rewrite-url-stability-remove-false-serve-prompts-for) (items 1 & 2 — the truth pass; redirects + `--strict` CI left open, see below)

## The problem

`mkdocs/index.md` is the single most-read page Visivo publishes, and it told new users to do something the CLI has never done. Three claims, every one of them checkable from the code:

| Where | Claim | Reality |
| --- | --- | --- |
| `index.md:26` | "Visivo will prompt you to select from several example dashboards" | `visivo/commands/serve.py` contains no `click.prompt`. In an empty directory it sets `new = not discover.project_file_exists` and builds an in-memory `Project(name="Quickstart Visivo")` with empty lists. |
| `index.md:48` | The tip block repeats it: "Prompts you to choose from available examples / Initializes the selected example into your directory" | Same code path. It writes no project file at all. |
| `index.md:67` | ```yaml\ncharts:\n  - name: revenue_chart\n    layout:\n      title: Monthly Revenue\n``` | The tool's own parser rejects this: `Validation error for layout at location: ['title']: "Monthly Revenue" is not of type "object"`. |

A new user following that page ran `visivo serve`, got a blank workspace and a `No jobs run.` log line, and had nothing to compare it against.

## What I verified by hand

Fresh arm64 venv on `origin/main`, `VISIVO_TELEMETRY_DISABLED=true`, each in its own clean tmp dir:

```
$ visivo init --example ev-sales --bare
Loading example project 'ev-sales' for 'ev'
Successfully loaded example project 'ev-sales'
Done                                                  (1.27s)
-> .env  data/  ev_sales.duckdb  project.visivo.yml

$ visivo run
Built schema for source ev_sales_db (1 tables, 5 columns)  [SUCCESS]
Updated data for insight total_revenue_kpi ...             [SUCCESS]
  (5 insights)                                        Run finished in 0.23s

$ visivo init --example college-football --headless --port 8007
-> server up, initial build green, GET / -> 200,
   /api/project/ -> {"name":"full","dashboard_count":1,...}

$ visivo serve --port 8007 < /dev/null          # empty directory
Project loaded in 0.05s
Running initial project build...
No jobs run. Ensure your filter contains nodes that are runnable.
View your project at: http://localhost:8007
-> no prompt, no project.visivo.yml written, GET / -> 200
```

[...]

PR #644, opened 2026-08-28
Docs: the dashboard-building path and the visivo dist capability matrix

jpjsonic:
## The problem

Two P1/P2 docs gaps from the audit:

1. **No dashboard-building guide.** `mkdocs/topics/` had 14 topics and not one taught the thing the product exists for — assembling a dashboard. The blocker (is an insight a dashboard item?) was settled by the 08-27 ruling: insights are **not** first-class items; the Chart wrapper stays.
2. **`visivo dist` was one sentence** at the bottom of `deployments.md`, with no statement of what a static bundle can and cannot do — and no warning anywhere that a bundle ships the query **results** as static files.

## The fix

**`mkdocs/topics/dashboards.md` — "Build a Dashboard"**
- The full path `source → model → insight → chart → dashboard`, with the Chart taught as the **composition boundary**: an item never takes a bare insight; a chart wraps one or more insights (dual-axis combination, reuse across charts).
- The layout system in depth: rows/items, **sum-normalized relative widths** (`1/1/2` ≡ `25/25/50`; ratios only, per row), row heights (token table + integer pixels), markdown items (inline, named + `${ref()}`, align/justify), nested row-containers, empty items.
- Both authoring routes: YAML and the in-browser workspace (including the #632/#637 auto-wrap behavior when dropping an insight on the canvas).

**`mkdocs/topics/static-distribution.md` — "Static Distribution"**
- What **works** (full rendering; inputs + filter/sort/split run client-side in DuckDB-WASM over the shipped parquet; any static host; `--deployment-root` sub-paths; `_redirects` deep links), what **degrades** (data frozen at `visivo run` time; interactivity bounded by shipped data; SPA fallback is host-dependent), what is **impossible** (live queries, editing/workspace/Explorer, alerts, access control, cloud stages).
- The **data-exposure warning**: `dist/data/files/*.parquet` holds every row of every materialized result — including columns no chart displays — plus dashboard configs and model SQL text in the JSON manifests. Anyone who can fetch the files sees the data.

Both pages registered in the `mkdocs.yml` nav (Tutorials); `deployments.md`'s Static section now links the new page. No Python source changes. No new `known_words.txt` entries were needed (spellcheck is clean).

## Ground-truthing (every claim tested against the current code)

Built probe projects and ran `visivo run && visivo dist` on the current `dist_phase.py` (post-#633):

- `data/files/*.parquet` contains the full result rows, including the undisplayed `region` column → the "every row, including columns no chart displays" wording.
[...]

PR #662, opened 2026-08-29
Docs example harness: execute every YAML fence against the real schema

jpjsonic:
## Why

Nothing in this repo has ever executed a documented example. **M8** — four field testers copied the `visivo init` scaffold's dashboard block and were rejected by Visivo's own parser — was one instance of a class, not a one-off: any ` ```yaml ` block we publish can be config the tool refuses, and we would never know. PR #636 closed that hole for the scaffold. This closes it for everything else.

**Discovery result: 51 of the 152 YAML examples we publish were rejected by the real schema.**

| Outcome | Count |
| --- | --- |
| Fixed by correcting the example | **33** |
| Exempted with an in-page `visivo-example: skip` marker | 12 |
| Re-tagged ` ```text ` — lists of expression syntax, not YAML documents | 3 |
| Marked `visivo-example: invalid` — the page's own counter-example | 1 |
| Resolved by the harness itself (validating an `annotations:` fragment via `Layout`) | 1 |
| Handed to PR #658, which owns `mkdocs/index.md` | 1 |

## The harness — `tests/docs/test_docs_examples.py`

Extracts every YAML fence from

- hand-authored `mkdocs/**/*.md` (reusing `test_docs_parity.EXCLUDED_DOC_SUBPATHS`), and
- every docstring under `visivo/`, walked with `ast` so only real docstrings count — **model docstrings *are* the generated reference pages**, so a broken fence there ships to docs.visivo.io verbatim,

and validates each with the #636 technique: `yaml.safe_load` the fence, then construct the **real** model — via a `TypeAdapter` over the matching `Project` field annotation, so `sources:` is checked by the actual discriminated `SourceField` union, `dashboards:` by `DashboardField`, and so on. Fragment shapes that legitimately appear in docs but are not `Project` fields (`rows:`, `items:`, `props:`, `interactions:`, `columns:`, `format_cells:`, `shapes:`, `annotations:`) are validated by their **real owner** (`Dashboard`, `Row`, `Insight`, `Table`, `Layout`) — the check applied at the right altitude, not a weakening of it. A fence carrying a top-level `name:` is validated with `Project` itself, which additionally resolves every `${ref()}` inside it.

The fence regex handles what the docs actually contain: ` ```yaml `, ` ``` yaml ` (with a space), `title="…"` attributes, and **indented** fences inside `!!! admonition` and `=== "tab"` blocks. A naive column-0 `​```yaml` regex finds 32 of 149; `test_indented_and_spaced_fences_are_discovered` pins that down so the harness cannot silently shrink back to a no-op.

Failures name **the file, the fence number, the line, the fence's first line, and the validation error** — a harness whose output is unreadable gets turned off the first time it fires:

```
mkdocs/topics/annotations.md fence #5 (line 253)
  first line: models:
  `charts:` rejected by the schema:
  1 validation error for list[Chart]
  0.layout
    Value error, Validation error for layout at location: ['title']:
    "Highlighted Region in Scatter Plot" is not of type "object"
```

## Docs bugs found and fixed (33 fences)

Every one of these was a copy-pasteable lie until this PR.

| Bug | Count | Where |
| --- | --- | --- |
| `layout: title: "string"` — Plotly's layout schema requires `title` to be an **object**, so Visivo rejects the string form outright | 12 | `topics/annotations.md`, `topics/including.md` |
| `>{ … }` eval strings written unquoted — `>` opens a YAML **folded block scalar**, so these blocks were YAML *syntax errors*, not just schema errors | 6 | `topics/testing.md` ×4, `visivo/models/alert.py` |
| `models: - args: [...]` — the `CsvScriptModel` shape, **deleted in cf70f943**, still documented as the primary example on the Chart reference page | 6 | `visivo/models/chart.py` ×3, `topics/annotations.md` ×3 |
| `Test.if` given a bare context string; `if` is an `EvalString` and must be `>{ … }` | 1 | `topics/testing.md` |
| `alerts: … destinations: - ${ref(…)}` — `Alert.destinations` is `List[DestinationField]`, a discriminated union of concrete Destinations with **no `Ref` member**, so refs are rejected. Examples now use the inline form that works (see *Decision needed* below) | 2 | `visivo/models/alert.py`, `visivo/models/defaults.py` |
| The **top-level `Project` example** — a model with no `source:` and no `defaults.source_name` is rejected: *"'orders' does not specify a source"* | 1 | `visivo/models/project.py` |
| `- width 1` (missing colon) — a YAML syntax error on the published **Dashboard** reference page | 1 | `visivo/models/dashboard.py` |
| `chart: ...` — literal YAML `...` ellipses that the `Ref` pattern rejects | 1 | `visivo/models/dashboard.py` |
| A **SlackDestination** example filed under `alerts:` instead of `destinations:` | 1 | `visivo/models/destinations/slack_destination.py` |
| `attach:` inlining the attached source's fields instead of nesting them under `source:` | 1 | `visivo/models/sources/sqlite_source.py` |
| `format_cells.scope: columns` — the enum is `row` \| `column` \| `table` | 1 | `visivo/models/table.py` |
| `type: rectangle` — the Plotly shape enum is `circle` \| `rect` \| `path` \| `line`; plus a broken list indent that made the whole block unparseable | 1 | `topics/annotations.md` |
| Snowflake source missing the **required** `database` field | 1 | `topics/environment-variables.md` |
| An insight `props:` fragment with no `type:` — the props union is discriminated on it | 4 | `topics/query-and-context-strings.md` |

(Bug count exceeds fence count: two fences carried two bugs each — `testing.md`'s conditional block was both an unquoted eval string *and* a bare `Test.if`, and `annotations.md`'s shape example had both a broken indent and `type: rectangle`.)

The `visivo/models/**` docstring edits change schema descriptions, so `visivo/src/visivo_project_schema.json` is regenerated and committed (enforced by `test_committed_schema_is_fresh`).

## Exemptions — explicit, in the page, individually justified

There is deliberately **no skip list in the test file**. A fence that is not Visivo config carries an HTML comment on the line above it, so a doc reader sees the reason in the page:

```
<!-- visivo-example: skip - a GitHub Actions workflow file, not Visivo project config -->
```

Three kinds: `skip` (never validated), `invalid` (a deliberate counter-example — **asserted to fail**), `broken` (tracked debt — **asserted to fail**). `test_markers_carry_a_reason` rejects a bare marker, and `test_must_fail_markers_still_describe_failing_fences` fails the build the moment an `invalid`/`broken` fence starts validating — **so the exemption list can only shrink**. There are **zero `broken` markers**: nothing was parked.

**All 13 exemptions, each justified:**

| Fence | Kind | Why |
| --- | --- | --- |
| `topics/deployments.md` ×7 | skip | RWX Mint / GitHub Actions workflow files (`.mint/manage_ci.yml`, `.github/workflows/*.yml`) — real YAML, not Visivo project config |
| `topics/ci-cd.md` ×1 | skip | GitHub Actions workflow file |
| `topics/annotations.md` ×2 | skip | Plotly **property listings**: every value is a type description (`'color'`, `'number greater than or equal to 0'`), not a value. They are reference tables that happen to be fenced as YAML |
| `topics/telemetry.md` ×1 | skip | The global `~/.visivo/config.yml`, not a project file |
| `how_it_works.md` ×1 | skip | A **dbt™ `schema.yml`**: the `models:` block there is dbt's model catalog (`description`/`columns`), validated by dbt, not Visivo's `models:`. See *Decision needed* |
| `topics/query-and-context-strings.md` ×1 | invalid | The page's own counter-example (`# Invalid: the slice must be last`). Asserted to still fail |

Three other fences on that page were **lists of expression syntax**, not YAML documents (`${ref(x).col}  # a comment` ×3 lines). Their fences are re-tagged ` ```text `, which is what they are — no marker needed, and they leave the harness's scope honestly rather than by exemption.

### One handoff, not an exemption

`mkdocs/index.md:67` has the same `layout: title: "Monthly Revenue"` bug fixed 12× elsewhere here, but **PR #658 is rewriting that file**, so this branch does not touch it. It is recorded in a `HANDOFFS` dict in the test module, named with the owning PR, and held to the same no-rot rule: `test_must_fail_markers_still_describe_failing_fences` fails if the fence starts passing, `test_handoffs_are_still_needed` fails if the fence disappears. **@#658: please apply `title: {text: Monthly Revenue}` and delete the `HANDOFFS` entry.**

Also checked and clean: `topics/dashboards.md` (12 fences) and `topics/static-distribution.md` from #644 — all pass untouched.

## Decision needed (not fixed here — reported)

1. **`Alert.destinations` does not accept refs, but its own field description says it does.** The description reads *"Destination objects defined inline **or `${ ref() }`s to destinations**"*, yet the annotation is `List[DestinationField]` — a `type`-discriminated union of `SlackDestination | EmailDestination | ConsoleDestination` with no `Ref` member. Both `alerts: … destinations: [${ref(x)}]` and the bare `ref(x)` form are rejected, at the `Alert` level *and* at the `Project` level. This PR rewrites the two examples to the **inline** form that actually works, and deliberately does **not** touch the field description or the annotation: fixing this is either a model change (add `Ref` to the union) or a description correction, and that is a product call, not a docs call.
2. **`how_it_works.md`'s dbt tab** mixes dbt's `models:` (with `description`/`columns`) and Visivo's `insights:`/`charts:` in one `models/schema.yml`. Skipped with that reason, but somebody who knows the dbt integration should confirm whether Visivo ever parses that file — if it does, the block is wrong; if it does not, the `insights:` in it are unreachable.

## CI

One-line change to `.rwx/test_docs.yml`'s existing `docs-parity` task — it now runs `tests/docs/test_docs_parity.py tests/docs/test_docs_examples.py`. Naming the file explicitly rather than widening to `tests/docs/` keeps the docs pipeline's scope unchanged apart from this addition. The new tests also run in the full suite via `test_visivo.yml`.

## Verification

- `pytest tests/ -q` → **2488 passed**, 2 skipped, 5 xfailed, 1 xpassed
- `black --check --target-version py312 .` → clean (544 files)
- `PYTHONPATH=$PWD mkdocs build` → builds; `validate_mkdocs_build.sh` → spellcheck clean. Rendered HTML inspected: every marker sits **inside** its `admonition`/`tabbed-block` container and renders as an invisible HTML comment, so no page structure changed.

### Falsification

Guards were proved to fail with the fix removed, not merely to pass:

1. **Harness vs. unfixed docs** — `git stash` the doc/docstring fixes, keep the harness: `2 failed, 15 passed` — **38 markdown fences** plus **11 docstring fences** named with file, fence number, line, first line and validation error. (38 rather than 39 because `mkdocs/index.md` is held by the `HANDOFFS` entry.)
2. **Marker without a reason** — strip the reason from telemetry.md's marker:
   `AssertionError: These 'visivo-example' markers carry no (or too short a) justification … mkdocs/topics/telemetry.md fence #2 (line 102): <!-- visivo-example: skip -->`
3. **`invalid` marker rot** — fix the deliberate counter-example so it validates:
   `AssertionError: These fences are recorded as failing but now validate cleanly. Delete the marker … recorded as 'visivo-example: invalid'`
4. **Handoff removed** — empty the `HANDOFFS` dict:
   `mkdocs/index.md fence #1 (line 67) … "Monthly Revenue" is not of type "object"`

Plus 9 in-module self-checks that pin the harness's own teeth, including `test_harness_rejects_the_m8_shape` (`insight:` as a row item — the original bug) and `test_harness_rejects_unquoted_eval_strings`.


jpjsonic:

## Review pass: 8 findings fixed, 4 rejected

Every finding was re-derived against the branch before acting on it, and every fix was falsified by re-introducing the defect and watching a named test go red. Full script + transcript in the commit; the interesting outputs are below.

### Fixed

**`skip` had no anti-rot rule — and the one skip covering real config was hiding two-thirds of a fence** (findings #2, #7)

`invalid`/`broken` are asserted to still fail; `skip` was `continue`d before validation and never re-examined. That is the silent skip list this harness replaced, relocated into the page — and it is the one marker anybody can reach for to make a failing example go away forever. `skip` now carries the same rule (`test_skip_markers_still_describe_non_visivo_config`); all 12 existing skips already satisfy it, so it costs zero doc churn.

The reviewer's specific case is `mkdocs/how_it_works.md`'s dbt tab: an `insights:` block with `props.type`/`interactions` and a `charts:` block with `layout.title`, sitting beside dbt's `models:` catalog, all of it exempted by one whole-fence `skip`. New marker kind:

```
<!-- visivo-example: skip-keys(models) - ... only the models: block is dbt's catalog ... -->
```

Only the named keys are exempt; everything beside them is validated, and each named key is asserted to still be rejected on its own (so a `skip-keys` cannot rot either). Reproducing the reviewer's mutation — `type: scatter` → `type: scattr` and `layout.title` back to the string form — in that tab:

```
mkdocs/how_it_works.md fence #2 (line 55)
  first line: insights:
  `charts:` rejected by the schema:
    "Widget Sales by Week" is not of type "object"
1 failed, 32 passed
```

Under the old marker that mutation was `17 passed`. The failure message now offers `skip-keys` **before** `skip`, so the escape hatch it steers you toward is the narrow one.

**Refs were resolved in zero fragment fences** (finding #1)

Confirmed: 72 unmarked ref-bearing fences, all validated per key, and a `TypeAdapter` over one field cannot resolve a `${ref()}`. 24 of them carry everything they reference — whole projects missing only a `name:`. `validate_fence` now builds a `Project` for those, **after** per-key validation, so this can only add strictness. `_REF_RESOLVING_FLOORS` pins the counts per file. The reviewer's exact mutation:

```
- ${ref(scatter_data-source)}  ->  ${ref(scatter_data-sorce)}   (annotations.md fence #5)

FAILED test_ref_bearing_fences_keep_resolving_their_refs
  mkdocs/topics/annotations.md: 10 of 11 ref-bearing fences still resolve
```

**Not** attempted: page-scoped ref resolution (checking every `ref()` against every name defined anywhere on the page). Measured it first — 22 of the docs' 82 refs legitimately point outside their page ("assume you have a chart named `revenue_chart`"), spread over 8 files, 6 of them in `mkdocs/topics/dashboards.md`, which this branch is under orders not to rewrite. It would have been noisier than the bug.

**HANDOFFS keyed on a line number in a file another PR owns** (finding #3)

Confirmed exactly as described. Now keyed on a content fingerprint. Prepending one unrelated sentence to `mkdocs/index.md`: `33 passed` (was 2 failed). Actually *fixing* the fence still reds, which is the signal worth raising:

```
FAILED test_handoffs_are_still_needed
  mkdocs/index.md (fingerprint fcb7e485a014) — PR #658 owns mkdocs/index.md
```

**`Alert.destinations` description promised refs the field rejects** (finding #4)

Confirmed: the description sat one line below the docstring example this PR rewrote for that exact reason, and it is committed into `visivo_project_schema.json` (the IDE tooltip) and rendered on the reference page — structurally out of the fence harness's reach. Fixed, and generalised into `test_fields_that_promise_refs_actually_accept_them`, which walks every model in `visivo.models`: **14 fields promise `${ ref() }`; 13 keep the promise; `Alert.destinations` was the only liar.** Restoring the old string reds that test.

*Open product question, deliberately not decided here:* the alternative fix is to make `Alert.destinations` actually accept a ref to a top-level `destinations:` entry, which is what a user would expect. That is a model change, not a docs change — say the word and it gets its own PR.

**Two silent-drop paths in the extractor** (findings #5, #11)

`_FENCE_RE` accepts ` ```yml ` but the `.py` fast path filtered on the literal substring `"yaml"`, and the regex had no `re.IGNORECASE`. Zero instances today; both are one keystroke away and both fail silently, with no discovery-guard signal. Fixed and pinned. Falsified by restoring the substring prefilter and by dropping `IGNORECASE`.

**Discovery floors 40% below reality** (finding #10)

`> 60` against 96 and `> 40` against 53. Now 90 and 50. A glob regression shedding 25% of the markdown fences:

```
FAILED test_markdown_fences_are_discovered
  Only 72 YAML fences found ... below the floor of 90
```

**Three unused fragment specs** (finding #12)

Confirmed `interactions:\n  - bogus_key: 1` returned PASS. Each unused spec only widened the unknown-top-level-key allowlist, and `interactions` in particular would have admitted a fence and then not checked it. `interactions`, `columns`, `width` dropped; `test_every_fragment_spec_is_used_by_a_real_fence` keeps the dict honest, and that shape is now refused outright rather than waved through. (The underlying laxness — the interaction model accepting unknown keys, so a user's `splitt:` is silently ignored — is a real config-validation gap, but tightening it would reject existing user projects. Worth its own issue, not this PR.)

### Rejected

**`visivo/src/visivo_project_schema.json` vs open #645** (finding #6) — real overlap, not a defect in this branch, and the finding says so itself: the resolution is sequencing plus `python -m visivo.generate_project_schema_json` on whichever PR lands second, never a hand-merge of a 3.3MB single line. This branch's change stays machine-generated; I re-ran the generator after the `alert.py` edit and re-verified structurally: **0 keys added, 0 removed, 1 value changed** (`$defs.Alert.properties.destinations.description`).

**`.rwx/test_docs.yml` vs open #658** (finding #8) — inference about #658's diff, trivial resolution (keep all three test paths on the one pytest line). File untouched in this pass.

**`mkdocs/topics/deployments.md` attributed to #644** (finding #9) — self-rejecting, and correct to de-rate: #644 is already merged into this branch at `85afaae3`, so the ownership list is stale, not the file. Three of the sixteen listed PRs are merged; the list wants regenerating before the next lane uses it.

### Gates

```
2504 passed, 2 skipped, 5 xfailed, 1 xpassed        (docs harness: 17 -> 33 tests)
black --check --target-version py312 .              544 files unchanged
```

No viewer files touched. No file on the do-not-edit list touched — `mkdocs/index.md`, `mkdocs/topics/dashboards.md` and `.rwx/test_docs.yml` are all untouched by this commit.

tgsoverly (approved):
Great!

PR #650, opened 2026-08-28
Every failed job carries a Diagnostic; skipped jobs are recorded (M16 W3)

jpjsonic:
## The problem

A failed job knew what went wrong; nothing structured survived the trip. `JobResult` carried `message` — a dot-padded, ANSI-coloured string built for a terminal — and that was the whole story. The viewer had no typed failure to render.

Worse, a job **skipped** because an upstream failed produced *nothing at all*. The DAG runner logged one line and dropped the node:

```
Skipping job for 'revenue_insight' because it has a failed dependency
```

No `JobResult`, no `failed_job_results` entry. An insight sitting above a broken source silently produced no data, and only the source's failure surfaced anywhere.

## What this does

**`JobResult` grows an optional `diagnostic`** (the W0 contract from `visivo/models/diagnostic.py`). `message` is untouched and stays the terminal renderer; `diagnostic` is what will reach the viewer in W4.

Every failure construction site in `visivo/jobs/` now populates it — enumerated from the code, not assumed:

| Site | Code |
|---|---|
| `run_insight_job` query failure | `missing_relation` / `ambiguous_relation` / `query_execution_failed` |
| `run_insight_job` model-less insight (**new**, B9) | `missing_model` |
| `run_sql_model_job` query | `query_execution_failed` |
| `run_sql_model_job` schema-only | `schema_build_failed` |
| `run_source_schema_job` aggregation error | `schema_build_failed` |
| `run_source_schema_job` connected-but-no-tables | `schema_build_failed` |
| `run_source_schema_job` connection error | `source_locked` / `source_connection_failed` |
| `run_input_job` | `invalid_value` / `query_execution_failed` |
| `dag_runner` skipped job (**new**) | `dependency_failed` |

**Skipped jobs are now recorded.** A runnable node dropped for a failed dependency synthesizes a `dependency_failed` diagnostic naming the skipped object, with `related[]` pointing at the failed upstream(s), and a `SKIPPED` (yellow) terminal line that is visibly not a failure of its own. Presentation nodes — charts, dashboards, rows — stay unrecorded on purpose: they sit above failed dependencies too, and recording them would bury the real failure under derived noise.

The run summary now separates the two:

```
Run failed in 0.4s with 1 query error(s) (2 dependent job(s) skipped).
```

`FilteredRunner` / `DagRunner` expose `.diagnostics` for W4's `error_json` emitter to consume.

## Folded in: B9

[...]

PR #663, opened 2026-08-29
The agent contract: `visivo schema` with a core subset, and `--json` on compile/run/test

jpjsonic:
## Why

An LLM cannot author a Visivo project from the schema we ship. `visivo/src/visivo_project_schema.json` is **3,363,103 bytes across 103 `$defs`**, and over 98% of that is vendored Plotly: `Layout` alone is 423,983 bytes, plus one definition per trace type. Nobody puts that in a prompt, so agents guess at the grammar instead of reading it.

And there is no structured way to read what a command did. An agent has to scrape human log lines and an exit code, so `visivo compile` failing on a dangling ref tells it "1" and nothing else.

This PR ships both halves of the agent contract from the Guided First Run dossier (W7).

## 1. `visivo schema`

```bash
visivo schema                  # core authoring subset (default) -- 93,173 bytes
visivo schema --core           # the same, said explicitly
visivo schema --full           # everything, Plotly included     -- 3,363,104 bytes
visivo schema --props bar      # the Plotly vocabulary --core prunes, one type
visivo schema --layout         # the Plotly layout vocabulary
visivo schema -o core.json     # write to a file
visivo schema --indent 2       # pretty-print
```

Follows the existing command pattern: `visivo/commands/schema.py` for the Click command, `visivo/commands/schema_phase.py` for the logic.

`--core` is the **default**, because emitting 3.3 MB into a terminal on a bare `visivo schema` would be hostile, and because the core subset is the one an agent should be reading.

### The selection rule (re-derivable, not hand-maintained)

The subset is derived from the live models on **every invocation**. Six mechanical steps, documented in full in the `schema_phase.py` module docstring:

1. **Build from `generate_project_schema()`, not `generate_schema()`.** `generate_schema()` *is* the function that merges the vendored Plotly JSON on top of the Pydantic dump — it is what takes the file from 115 KB to 3.3 MB. Choosing the un-merged function prunes every trace-prop `$def` and the big `Layout` in one step, **by construction**, so a Plotly trace type added next year is excluded automatically with no list to update. Nothing dangles: in the Pydantic-native schema `Insight.props` is already `InsightProps` (`type` required, any other key allowed) and `Layout` is already an open object.
2. **Keep only the authoring surface at the root.** `CORE_PROJECT_PROPERTIES` names what an agent writes by hand; `OMITTED_PROJECT_PROPERTIES` names the rest *with a reason each*, as data rather than a comment.
3. **Transitive `$ref` closure**, dropping every unreachable `$def`.
4. **Delete the machine-set fields** (`path`, `file_path`, `project_file_path`, `project_dir`, `cli_version`) from every object — the CLI fills them in, and an agent that writes them is making a mistake. ~9 KB across 45 defs.
5. **Delete Pydantic's auto-generated property titles** — `"title": "Source Name"` above a property named `source_name` is ~4 KB of restatement.
6. **Restore the Python-keyword aliases.** `Test.if_` carries an underscore only because `if` is reserved in Python; every doc example writes `if:`. Any property named `<keyword>_` is emitted under the bare keyword too, and the underscore spelling is marked `deprecated`. The test is `keyword.iskeyword(name[:-1])`, which is *why* this does not drag in the 1.0-era `targets`/`target` aliases that a blanket `by_alias=True` would.

The rule cannot silently go stale: `test_core_schema_classifies_every_project_field` fails, and `core_schema()` raises `SchemaSelectionError` naming the offending field, if `Project` ever grows a field neither list classifies.

The emitted document carries an `x-visivo-schema` block naming the mode, CLI version, what was omitted and why, and the commands that return the omitted parts — so an agent reading only the JSON knows what it is missing and how to get it.

### Two deliberate deviations from the dossier, and why

- **Size: 93 KB, not the dossier's ~55 KB.** The 55,338-byte figure came from 17 defs including only "common sources". Getting there means dropping `SnowflakeSource` (7,389 bytes as `BigQuerySource`), `BigQuerySource`, `RedshiftSource` and `ClickhouseSource` — i.e. the contract would lie about what Visivo connects to, and an agent asked to wire up Snowflake would have no schema for it. I kept every source type. 93 KB is ~23k tokens, still cacheable, and comfortably under the dossier's own stated acceptance bar ("writes a file under 100KB").
- **No committed `visivo_core_schema.json`.** The dossier proposed a second committed artifact beside `write-schema-json`. A second committed schema doubles the parity-ring surface that just went red on clean main (#642). Generating on demand takes 30 ms, can never be stale, and needs no CI check. Say the word if you'd rather have the file.

`Layout` survives as the ~90-byte open-object stub the `$ref`s need, rather than dangling the reference — which satisfies the dossier's "exclude Layout" in the sense that matters (the 423,983-byte Plotly one is gone).

## 2. `--json` on compile, run and test

One JSON object on **stdout**, every human-readable line on **stderr**, exit code unchanged — so `visivo schema > core.json` and `visivo compile --json | jq` work, and `visivo run --json && visivo test --json` still short-circuits.

Stdout purity is enforced structurally: the command body runs inside `redirect_stdout(sys.stderr)`, which catches the `Logger`, the test runner's `.`/`F` dots, and any stray `print` deep in a job. `command_line.py` also refuses to start the Halo spinner (which writes escape codes straight to stdout) for `--json` or `schema` invocations, and suppresses the "Starting Visivo..." banner and the trailing execution-time line.

Three of the tests shell out to `python -m visivo.command_line` rather than using `CliRunner`, because `CliRunner` invokes the subcommand directly and never runs `command_line.py` — which is exactly where the banner, the timing line, the spinner and the error routing live. That gap was real: `visivo schema --core --full` was routing its error through `Logger` to **stdout**, so `visivo schema > core.json` on a bad invocation wrote the error message into the file the agent was about to parse. Fixed in `b80ea1f9`.

### The envelope (stable; every key always present)

```json
{
  "visivo_json_version": 1,
  "command": "compile",
  "success": true,
  "cli_version": "2.1.1",
  "duration_ms": 812,
  "result": { },
  "errors": [ ]
}
```

`errors` is always a list, empty on success, of objects with seven stable keys:

```json
{
  "code": "bad_reference",
  "name": "total_units_kpi",
  "message": "The reference \"ref(does_not_exist)\" on item \"total_units_kpi\" does not point to an object.",
  "file": "/work/project.visivo.yml",
  "line": 16,
  "object_path": "project.insights[0]",
  "details": null
}
```

`name`/`file`/`line` are filled in even when the Pydantic error does not carry them: `_resolve_in_files` re-parses the project YAML with the line-annotating loader and looks the object up **structurally** (no pattern matching), either by `name` or by the `(collection, index)` location — and reports nothing rather than the wrong object when `includes` make the index ambiguous.

`result` is command-specific and documented in `visivo/commands/json_output.py` and `mkdocs/reference/agent-contract.md`: `compile` returns the object names actually parsed per collection; `run` returns one entry per job with the runner's alignment dots stripped and the exception text and artifact path pulled out of the formatted line; `test` returns every assertion with its own name resolved off the project.

### Two behaviour notes

- `run_phase` and `test_phase` gain `defer_exit`. `DagRunner.run()` calls `sys.exit(1)` on a failed job, which would end the process before the envelope is printed — and, worse, unwinds `FilteredRunner.run()` before it aggregates the per-job results the envelope reports. `defer_exit` runs soft and hands the exit code back to the caller. One visible consequence, documented: **`run --json` attempts every filtered DAG and reports every failing job in one pass**, instead of stopping at the first failing dashboard.
- If anything deep in the stack calls `sys.exit()` anyway, `json_command` still emits an envelope (`"code": "exited"`) before honouring the code. An agent that got no JSON cannot tell a crash from a clean run.

### PR #650 (diagnostics)

`visivo/models/diagnostic.py` is **not touched**. `json_output.py` deliberately defines its own flat error shape instead of emitting `Diagnostic`s. Once #650 lands, `_error` should be replaced by the Diagnostic contract and `visivo_json_version` bumped to 2 — noted in the module docstring so it does not get lost.

## Schema JSON

`visivo/src/visivo_project_schema.json` was **not regenerated** — no Pydantic model changed in this PR. `git diff --name-only origin/main` touches nothing under `visivo/models/`.

## Tests

`tests/commands/test_schema_phase.py` (22), `test_schema.py` (15), `test_json_mode.py` (12), `test_json_output.py` (16) — 65 new tests.

Coverage of the acceptance criteria:
- the command emits valid JSON, and stdout carries nothing but the document (both with and without `-o`);
- `--core` compiles as a JSON Schema (`jsonschema_rs.validator_for` raises on a dangling `$ref`), is under 100 KB, is <1/20th of `--full`, keeps `SqlModel`/`Metric`/`Dimension`/`Relation`/`Insight`/`InsightInteraction`/`Chart`/`Dashboard`/`Row`/`Item` **and all ten source types**, and excludes every Plotly trace-prop def, `xaxis`/`yaxis`/`colorscale`, and the fat `Layout`;
- **`--core` validates all three real sample projects** in `visivo/templates/samples/` (parametrised over the glob, so a new sample is covered automatically) — and still rejects a misspelled key at the root and inside an object, so the "it validates everything" failure mode is ruled out;
- `--json` on each of compile/run/test is parseable and **stable across a success and a failure case** — the envelope key set and the per-entry key sets are asserted identical in both.

Full suite: **2535 passed, 2 skipped, 5 xfailed, 1 xpassed**. `black --check --target-version py312 .` clean.

### Falsification

Every guard was proved to fail with its implementation removed (stash the impl, keep the test, run, restore). 16 mutations, 16 caught:

| Mutation | Test | Result |
| --- | --- | --- |
| `core_schema` built from `generate_schema()` (Plotly not pruned) | `test_core_schema_prunes_the_plotly_vocabulary` | FAILED — "Plotly trace-prop defs leaked into the core subset" |
| machine-set fields left in every def | `test_core_schema_drops_machine_set_fields_everywhere` | FAILED — "Attachment still exposes machine-set field(s) ['path']" |
| keyword aliases not restored | `test_core_schema_restores_python_keyword_aliases` | FAILED — `'if' in {...}` |
| unclassified `Project` field silently ignored | `test_core_schema_raises_when_a_project_field_is_unclassified` | FAILED — "DID NOT RAISE SchemaSelectionError" |
| root `additionalProperties: true` | `test_core_schema_rejects_a_misspelled_key` | FAILED — "a misspelled key at the project root must be rejected" |
| `insights` dropped from the authoring surface | `test_core_schema_validates_the_shipped_sample_projects` | FAILED on all 3 samples |
| `_resolve_in_files` disabled | `test_compile_json_failure_carries_name_message_file_and_line` | FAILED — `file` is `None` |
| `redirect_stdout` removed | `test_compile_json_keeps_human_output_off_stdout` | FAILED — log line landed in the JSON stream |
| `defer_exit` no longer implies `soft_failure` | `test_run_json_failure_reports_the_failing_job` | FAILED — exit 0, no job errors |
| `SystemExit` escapes `json_command` without emitting | `test_json_command_still_emits_when_the_body_calls_sys_exit` | FAILED — `JSONDecodeError` on empty stdout |
| `test_errors` returns nothing | `test_test_json_failure_reports_each_failing_assertion` | FAILED — exit 0 |
| `_PROJECT_COLLECTIONS` drifts from the schema's authoring surface | `test_compile_result_covers_every_authorable_collection` | FAILED — and `test_compile_json_success` PASSED, so this guard is the only thing catching a collection silently vanishing from the report |
| `_split_location` call removed from the ClickException branch | `test_errors_from_a_yaml_syntax_click_exception_carry_file_and_line` | FAILED — `assert None == '/work/my project/project.visivo.yml'` |
| `x-visivo-schema` advertises a flag that does not exist | `test_every_command_the_schema_advertises_actually_runs` | FAILED — "No such option: --core-subset" |
| `AGENT_INVOCATION` dropped from `QUIET_INVOCATION` (banner not suppressed) | `test_end_to_end_schema_stdout_is_only_the_document` + `..._compile_json_stdout_is_only_the_envelope` | BOTH FAILED — `JSONDecodeError` at char 0 |
| `SCHEMA_INVOCATION` branch removed from `_report_error_for_agent` | `test_end_to_end_schema_error_leaves_stdout_empty` | FAILED — the error message landed in stdout |

The first pass of mutation 5 did **not** fail, because the guard only checked a misspelled key *inside* an object; the test was strengthened to cover the project root too, and then it failed as shown.

## Docs

New `mkdocs/reference/agent-contract.md` (+ one nav line): both surfaces, the selection rule, the envelope, all three `result` shapes, and a minimal agent loop. `visivo schema` also appears automatically in `reference/cli.md`, which generates from Click. codespell clean.

## Out of scope / found along the way

Building the `--json` failure fixtures turned up a pre-existing bug that `run --json` makes obvious, and it is worse than it first looked. **`visivo run` on all three shipped sample projects runs zero jobs and exits 0.**

Each sample names its single dashboard the same as the project (`EV Sales` / `College Football` / `GitHub Releases`). `run_phase` builds its default `dag_filter` as `+<dashboard name>+`, the DAG then holds two nodes with that name — the `Project` and the `Dashboard` — and `filter_dag("+EV Sales+")` returns **zero** sub-DAGs. No jobs, no error, exit 0, and an empty dashboard for anyone following the guided first run.

Isolated (my first guess, the space in the name, was wrong):

| project name | dashboard name | jobs run |
| --- | --- | --- |
| `EV Sales` | `EV Sales` | **0** |
| `EV Sales` | `EV Sales Overview` | 7 |
| `EV Sales` | `EVSales` | 7 |
| `ev-project` | `EV Sales` | 7 |

Not fixed here — it is a `filter_dag` name-resolution question, and `visivo/query/**` / the DAG are hot in other open PRs. Filed separately as **VIS-1324**.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -146,6 +146,7 @@ nav:
       - Authentication: cloud/authentication.md
   - Reference:
       - CLI: reference/cli.md
+      - Agent Contract: reference/agent-contract.md
       - Configuration:
           - Alert:
               - reference/configuration/Alert/index.md
--- a/mkdocs/reference/agent-contract.md
+++ b/mkdocs/reference/agent-contract.md
@@ -17,6 +17,7 @@ Both write to **stdout** and send every human-readable line to **stderr**, so
 
 ```bash
 visivo schema                  # the core authoring subset (default)
+visivo schema --core           # the same, said explicitly
 visivo schema -o core.json     # write it to a file
 visivo schema --indent 2       # pretty-print
 visivo schema --full           # the complete schema, Plotly vocabularies included
--- a/mkdocs/reference/agent-contract.md
+++ b/mkdocs/reference/agent-contract.md
@@ -32,7 +32,7 @@ about 3.3 MB across 103 definitions. Most of that is the Plotly property
 vocabulary: the `Layout` definition alone is over 400 KB, and there is one
 definition per chart type. No agent can hold that in a prompt.
 
-The core subset is about 90 KB — roughly 23k tokens, small enough to cache in a
+The core subset is about 95 KB — roughly 24k tokens, small enough to cache in a
 system prompt — and describes the objects you actually write in
 `project.visivo.yml`: sources, models, metrics, dimensions, relations, insights,
 charts, tables, markdowns, inputs, dashboards and tests.
@@ -46,15 +46,19 @@ drift from the models. The rule, in full, lives in the module docstring of
 1. Build from the Pydantic dump of `Project`, before the vendored Plotly JSON is
    merged in. That single choice prunes every trace-prop definition and the
    large `Layout` — and prunes any Plotly type added in future automatically.
-2. Keep only the project fields you author by hand. `dbt`, `alerts`,
+2. Describe only the project fields you author by hand. `dbt`, `alerts`,
    `destinations` and the machine-set bookkeeping fields are out, each with a
    stated reason recorded in the emitted schema itself.
 3. Take the transitive `$ref` closure of what is left and drop everything
    unreachable.
 4. Delete the fields the CLI fills in — `path`, `file_path`, `project_file_path`,
-   `project_dir`, `cli_version` — from every object.
+   `project_dir`, `cli_version` — from every object that *inherits* them.
 5. Delete Pydantic's auto-generated property titles, which restate the key.
-6. Restore the aliases that only carry a trailing underscore because the name is
+6. Rewrite the `oneOf`s that no value can satisfy. A union chosen by a Python
+   callback — a password, which may be a literal or an `${env.*}` reference —
+   dumps as two branches that both accept any string, and `oneOf` demands
+   exactly one match.
+7. Restore the aliases that only carry a trailing underscore because the name is
    a Python keyword, so a test's gate is spelled `if`, as the docs show, and the
    `if_` spelling is marked deprecated.
 
@@ -65,6 +69,17 @@ vocabulary for one chart type it asks for it with `visivo schema --props bar`.
 The emitted document carries an `x-visivo-schema` block naming the mode, the CLI
 version, what was omitted and why, and the commands that return the omitted
 parts.
+
+!!! note "Omitted is not prohibited"
+    `alerts`, `destinations` and `dbt` are undescribed because they are
+    orthogonal to building a dashboard — not because writing them is a mistake.
+    Since the project root forbids unknown keys, each is still *accepted*, as a
+    permissive stub pointing at `--full`. A project that already uses them
+    validates unchanged.
+
+    The same goes for `path` inside `includes:`. It shares a name with the
+    machine-set `path` step 4 removes, but it is declared by `Include` itself
+    and is the only key an include has, so it stays.
 
 !!! note "Authoring vs. compiler output"
     Because step 4 removes the machine-set fields and every Visivo object
@@ -144,10 +159,20 @@ was picked up:
   "project_name": "EV Sales",
   "working_dir": "/work",
   "output_dir": "/work/target",
-  "objects": { "models": ["ev_sales"], "insights": ["units_by_quarter"], "…": [] },
-  "object_counts": { "models": 1, "insights": 1 }
+  "objects": {
+    "models": ["ev_sales"],
+    "metrics": ["revenue", "ev_sales.total_units"],
+    "insights": ["units_by_quarter"],
+    "…": []
+  },
+  "object_counts": { "models": 1, "metrics": 2, "insights": 1 }
 }
 ```
+
+Objects written *inside* another object are reported too, not just the ones at
+the project root: a metric or dimension defined on a model appears qualified as
+`model.name` — which is also how you reference it, `${ref(model).name}` — and an
+insight written inline in a chart appears under `insights` by its own name.
 
 ### `run` result
 
@@ -172,11 +197,22 @@ removed:
 }
 ```
 
+`error` carries the whole diagnostic, which is often several lines — the failing
+expression and the position within it, not just the first sentence. `artifact`
+is a path: the file the job wrote, or the SQL dumped for a query that failed.
+
 Every failed job also appears in `errors` with `"code": "job_failed"`.
 
 A `--json` run reports **every** failing job in one pass rather than stopping at
 the first failing dashboard, which is the one behavioral difference from a
 plain `visivo run`.
+
+!!! note "`total: 0` means nothing ran"
+    A run whose filter matches no runnable node succeeds and exits 0 — the same
+    as `visivo run`, which prints "No jobs run" and also exits 0. `--json` does
+    not turn that into an error, because the exit code is contractually the
+    same with and without the flag. Check `job_counts.total` rather than
+    `success` when you need to know that work actually happened.
 
 ### `test` result
 
--- a/mkdocs/reference/agent-contract.md
+++ b/mkdocs/reference/agent-contract.md
@@ -1,0 +1,205 @@
+# Agent Contract
+
+Two CLI affordances exist so that an LLM agent can author a Visivo project by
+reading the grammar instead of guessing at it, and can read command results
+structurally instead of scraping log lines.
+
+- `visivo schema` — the project JSON Schema, small enough to put in a prompt.
+- `--json` on `visivo compile`, `visivo run` and `visivo test` — one JSON object
+  on stdout describing what happened.
+
+Both write to **stdout** and send every human-readable line to **stderr**, so
+`visivo schema > core.json` and `visivo compile --json | jq` work unmodified.
+
+---
+
+## `visivo schema`
+
+```bash
+visivo schema                  # the core authoring subset (default)
+visivo schema -o core.json     # write it to a file
+visivo schema --indent 2       # pretty-print
+visivo schema --full           # the complete schema, Plotly vocabularies included
+visivo schema --props bar      # every Plotly property valid for `props: {type: bar}`
+visivo schema --layout         # every Plotly layout property
+```
+
+### Why there is a core subset
+
+The schema that ships with Visivo, and that the viewer validates against, is
+about 3.3 MB across 103 definitions. Most of that is the Plotly property
+vocabulary: the `Layout` definition alone is over 400 KB, and there is one
+definition per chart type. No agent can hold that in a prompt.
+
+The core subset is about 90 KB — roughly 23k tokens, small enough to cache in a
+system prompt — and describes the objects you actually write in
+`project.visivo.yml`: sources, models, metrics, dimensions, relations, insights,
+charts, tables, markdowns, inputs, dashboards and tests.
+
+### How the subset is chosen
+
+The subset is derived mechanically every time the command runs, so it cannot
+drift from the models. The rule, in full, lives in the module docstring of
+`visivo/commands/schema_phase.py`; in short:
+
+1. Build from the Pydantic dump of `Project`, before the vendored Plotly JSON is
+   merged in. That single choice prunes every trace-prop definition and the
+   large `Layout` — and prunes any Plotly type added in future automatically.
+2. Keep only the project fields you author by hand. `dbt`, `alerts`,
+   `destinations` and the machine-set bookkeeping fields are out, each with a
+   stated reason recorded in the emitted schema itself.
+3. Take the transitive `$ref` closure of what is left and drop everything
+   unreachable.
+4. Delete the fields the CLI fills in — `path`, `file_path`, `project_file_path`,
+   `project_dir`, `cli_version` — from every object.
+5. Delete Pydantic's auto-generated property titles, which restate the key.
+6. Restore the aliases that only carry a trailing underscore because the name is
+   a Python keyword, so a test's gate is spelled `if`, as the docs show, and the
+   `if_` spelling is marked deprecated.
+
+Nothing dangles as a result. `props` stays an open object that requires `type`,
+and `layout` stays an open object; when an agent needs the real Plotly
+vocabulary for one chart type it asks for it with `visivo schema --props bar`.
+
+The emitted document carries an `x-visivo-schema` block naming the mode, the CLI
+version, what was omitted and why, and the commands that return the omitted
+parts.
+
+!!! note "Authoring vs. compiler output"
+    Because step 4 removes the machine-set fields and every Visivo object
+    forbids unknown keys, the core schema rejects a document that carries them —
+    notably the `project.json` written at compile time. Validate hand-written
+    YAML with the core schema; validate compiler output with `--full`.
+
+---
+
+## `--json` on compile, run and test
+
+```bash
+visivo compile --json
+visivo run --json
+visivo test --json
+```
+
+Each prints exactly one JSON object on stdout and nothing else. The exit code is
+unchanged, so `visivo run --json && visivo test --json` still short-circuits.
+
+### The envelope
+
+Every key is always present.
+
+```json
+{
+  "visivo_json_version": 1,
+  "command": "compile",
+  "success": true,
+  "cli_version": "2.1.1",
+  "duration_ms": 812,
+  "result": { },
+  "errors": [ ]
+}
+```
+
+| Key | Meaning |
+| --- | --- |
+| `visivo_json_version` | Bumped when the shape changes in a way a consumer would notice. |
+| `command` | `compile`, `run` or `test`. |
+| `success` | Whether the command achieved what it was asked to do. |
+| `cli_version` | The CLI that produced the document. |
+| `duration_ms` | Wall clock for the command body. |
+| `result` | Command-specific; see below. |
+| `errors` | Always a list, empty on success. |
+
+### Errors
+
+Each entry has the same seven keys. Any of `name`, `file`, `line`,
+`object_path` and `details` may be `null` when the CLI genuinely does not know
+the value, so a consumer can index without guarding.
+
+```json
+{
+  "code": "bad_reference",
+  "name": "revenue_by_month",
+  "message": "The reference \"ref(orders)\" does not point to an object.",
+  "file": "project.visivo.yml",
+  "line": 31,
+  "object_path": "project.insights[0]",
+  "details": null
+}
+```
+
+`code` is the machine-readable kind: a Pydantic error type such as
+`bad_reference` or `extra_forbidden` for configuration problems, `job_failed`
+for a query that did not run, `test_failed` for a failing assertion, and
+`cli_error` for a usage or YAML-syntax problem.
+
+### `compile` result
+
+What the CLI actually parsed, so an agent can confirm the object it just wrote
+was picked up:
+
+```json
+{
+  "project_name": "EV Sales",
+  "working_dir": "/work",
+  "output_dir": "/work/target",
+  "objects": { "models": ["ev_sales"], "insights": ["units_by_quarter"], "…": [] },
+  "object_counts": { "models": 1, "insights": 1 }
+}
+```
+
+### `run` result
+
+One entry per job the DAG runner executed, with the runner's alignment padding
+removed:
+
+```json
+{
+  "project_name": "EV Sales",
+  "output_dir": "/work/target",
+  "jobs": [
+    {
+      "name": "units_by_quarter",
+      "type": "Insight",
+      "success": false,
+      "summary": "Failed job for insight units_by_quarter",
+      "error": "Column 'quarter' not found on model 'ev_sales'.",
+      "artifact": "target/insights/units_by_quarter/query.sql"
+    }
+  ],
+  "job_counts": { "succeeded": 2, "failed": 1, "total": 3 }
+}
+```
+
+Every failed job also appears in `errors` with `"code": "job_failed"`.
+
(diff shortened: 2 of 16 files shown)


jpjsonic:

Two follow-up commits since the description was written:

- **`fa824089`** — `_PROJECT_COLLECTIONS` in `json_output.py` restated the authoring surface that `schema_phase.CORE_PROJECT_PROPERTIES` already owns, so a new list-valued `Project` field would trip the schema guard but vanish silently from `compile --json`'s object report. A guard test now derives the expected set from `CORE_PROJECT_PROPERTIES` (less `includes`, whose members carry a path rather than a name). Also stopped `_names()` falling back to `str(item)`, which would have dumped a whole Pydantic repr into the envelope for a nameless object.

- **`5ccccbae`** — YAML syntax errors now keep their file and line. `load_yaml_file` already raises a `ClickException` carrying `Location: <file>:<line>[<col>]`, and the envelope was throwing that away into a prose message. A bad indent is one of the commonest first-run failures; an agent should get the line, not a paragraph:

  ```json
  {"code": "cli_error", "file": ".../project.visivo.yml", "line": 4, ...}
  ```

  `_split_location` reads the first line only (a project path may contain a space) and drops the column suffix.

Both falsified: mutation 12 (`_PROJECT_COLLECTIONS` derivation removed → `test_compile_result_covers_every_authorable_collection` FAILED) and mutation 13 (`_split_location` call removed from the ClickException branch → `test_errors_from_a_yaml_syntax_click_exception_carry_file_and_line` FAILED, `assert None == '/work/my project/project.visivo.yml'`).

Suite now **2530 passed**, `black --check --target-version py312 .` clean.

jpjsonic:

## Review fix pass — 69c93c7d

Nine findings verified and fixed, five rejected with reasoning, and one new
defect the fix work surfaced. Full suite green (2563 passed), `black --check`
clean, and all 12 mutations caught.

---

### Fixed

**1. `Include.path` deleted from the core schema (HIGH — confirmed)**

`MACHINE_SET_FIELDS` was applied by *name*, so `Include.path` went with the
bookkeeping `path` that every other object inherits from `BaseModel`. `Include`
declares its own, it is the only key an include has, and the def is
`additionalProperties: false` — so the emitted schema rejected the example
printed in `Include`'s own description.

```
$ jsonschema_rs.validator_for(core_schema()).validate(
      {"name": "demo", "includes": [{"path": "models/orders.yml"}]})
Additional properties are not allowed ('path' was unexpected)
```

51 of the repo's project files use `includes:`; every one was rejected.

Step 4 now asks **which class declared the field** — one of `BaseModel` /
`NamedModel` / `Project` (bookkeeping, strip it) or the concrete model itself
(authoring surface, keep it). Derived, not a hand-maintained exemption list.

**2. Omitted read as prohibited (MEDIUM — confirmed)**

`alerts`, `destinations` and `dbt` are hand-authored and legal, but the root is
`additionalProperties: false`, so leaving them undescribed made them illegal.
An agent asked to add a chart to a project with an `alerts:` block was told the
user's own configuration was an invalid key. Each is now emitted as a
permissive stub carrying its reason and pointing at `--full`. The machine-set
root fields stay rejected — that was always the documented intent
("validate hand-written YAML with `--core`, compiler output with `--full`").

**3. Every `password:` on every source type rejected (NEW — found by the fix)**

Not in the review; the repo-wide validation test I added for #1/#2 caught it.
Pydantic renders `SecretStrOrEnvVar` — `Union[EnvVarString, SecretStr]`, chosen
at runtime by a Python callable — as `oneOf: [{type: string}, {type: string,
format: password}]`. Both branches accept every string and `oneOf` demands
*exactly* one match, so no password validated, literal or `${env.*}`. Ten sites
across all nine source types.

New step 6 rewrites a `oneOf` whose branches are identical once annotation-only
keywords (`format`, `writeOnly`, …) are set aside — provably unsatisfiable —
to `anyOf`. Unions that really discriminate (the source-type union, distinct
`$ref`s) are untouched.

Net across #1–#3: **55 of the repo's 190 project files were rejected; now 0.**

**4. `visivo schema` stdout corrupted behind any group option (HIGH — confirmed, #2 and #8 are the same bug)**

`SCHEMA_INVOCATION` indexed `argv[1]`, so any group option before the
subcommand disabled it. Reproduced in a real process before the fix:

```
$ visivo -e .env schema > core.json      # exit 0
$ head -c 20 core.json                   # Starting Visivo...
$ tail -1 core.json                      # Visivo execution time: 0.91s
$ visivo -e .env schema --core --full > core.json
$ cat core.json                          # the error message, on stdout, exit 1
```

`_subcommand()` now scans past the group options (handling `-e .env`,
`--env-file=.env` and `-e.env`, and correctly *not* reading `visivo -e schema`
as a schema invocation). `--profile`'s own "Profiling..." / "Profiling
completed" went to stdout too — same class, also fixed. Five option shapes ×
three commands are now covered by real-subprocess tests, plus a unit test on
`_subcommand` including the negative cases.

Bonus: `_report_error_for_agent` labelled the envelope `"command": "help"` for
`visivo -e .env compile --json`, because the telemetry sanitiser maps anything
starting with a dash to "help". It now uses the resolved subcommand.

**5. `artifact` held a string that is not a path (MEDIUM — confirmed)**

`run_insight_job` appends `at <location>` and `query saved to: <path>` after
the `error:` line. `parse_job_message` treated every non-`error:` line as the
artifact (last one wins) and kept only the first line of the error, so
`errors[].file` came out as the literal `"query saved to: /…/insight_kpi.sql"`
and the lines naming the offending column were dropped. Now the prefix is
recognised and the error accumulates until a path line ends it.

**6. `compile --json` reported model-scoped metrics as absent (MEDIUM — confirmed)**

`compile_result` read only the root `Project` collections. A project whose
metrics are defined on a model — the spelling the semantic-layer docs lead with
— reported `"metrics": []`, and an agent following the documented loop
("write it, compile, confirm it appears") concludes its edit was dropped.

Generalised: model-scoped metrics and dimensions are reported qualified as
`model.name` (which is also how you reference them, `${ref(model).name}`), and
insights written inline in a chart by their own name — the same defect, which
the derived guard test surfaced. The guard walks `Project`'s own annotations, so
a model that grows another nested collection fails here rather than going quiet.

**7. `visivo/commands/options.py` (HIGH — confirmed as an ownership hit)**

`options.py` is listed in the manifest under the still-open #645, which does
modify it. The two edits would not have conflicted textually (#645 patches the
top of the file, this branch appended to the bottom), but the remediation is
free, so it is done rather than argued: the `--json` decorator now lives in
`json_output.py`, which compile/run/test already import. **`options.py` is
reverted to its base state and no longer appears in this diff.**

**8. `run --json` error/artifact only verified against hand-typed strings (MEDIUM — confirmed)**

`test_run_json_failure_reports_the_failing_job` does a real run and had a
failing job in hand but asserted only `name` and `message`. Now asserts
`job["error"]` is non-empty, differs from `summary`, is not the
`"Failed job for …"` fallback, is longer than the summary, and that each
`errors[].message` matches its job's error and each `errors[].file` is not a
log-line prefix. Mutation M9 (parse never extracts an error) fails 5 tests.

**9. No regression guard on plain `visivo run`'s exit code (MEDIUM — confirmed)**

Verified: hard-wiring `soft_failure=True` in `run_phase` passed
`tests/commands tests/jobs` — every broken pipeline would go green in CI.
Added `test_run_without_json_still_exits_non_zero_on_failure`, matching the
guards compile and test already had.

**10. Step 5 unguarded (MEDIUM — confirmed)**

Verified: `return`-ing out of `_strip_property_titles` left 36 tests green.
Added a guard that asserts no property below the `$def` level keeps a title,
that `$def` titles survive, and — so it cannot pass vacuously — that the
Pydantic dump really had titles to strip.

The related size-ceiling brittleness is addressed too: the ceiling moved from
100,000 to 125,000 bytes (current: 94,104) with a comment saying it is a prompt
budget, not a size target, since the structural guards are what actually police
the selection rule.

**11. Hand-rolled `_Project` stub (LOW — confirmed)**

`ProjectFactory()` now, per the factories-over-mocks convention, and the test
asserts `result["project_name"] == project.name` so the factory earns its place.
`_Item`/`_JobResult`/`_Runner` stay — they stand in for non-Pydantic runner
types with no factories.

---

### Rejected

**`no_jobs_run` missing from the run envelope (LOW).** Real symptom, wrong
remedy. `errors` being non-empty makes `json_command` exit 1, while plain
`visivo run` logs "No jobs run" and exits **0** — so adding the error would
break the module's own documented invariant that `--json` leaves the exit code
unchanged. Whether a zero-job run should fail is a product decision, not a
review fix. The data an agent needs is already in the envelope, so I documented
it instead: a `!!! note "total: 0 means nothing ran"` block in the contract page
and in `run_result`'s docstring, telling a consumer to read
`job_counts.total`, not `success`.

**`mkdocs.yml` ownership (HIGH as reported).** #644 is **merged**, and its merge
commit `85afaae3` is an ancestor of this branch's base. No open PR owns
`mkdocs.yml` (#658 owns `mkdocs/index.md` and `mkdocs/llms.txt` only). The one
line is the nav entry without which the new page is unreachable. Kept, and
re-derived rather than assumed — the manifest check is in the commit.

**Bare `ValidationError` branch never resolves file/line (LOW).** The symptom is
real and I reproduced it — a model missing a source yields all-null locators —
but the diagnosis names the wrong branch. That error routes through
`LineValidationError` (raised by `core_parser`), not the `ValidationError`
branch at line 314:

```
TYPE: LineValidationError
loc: ()   type: value_error   input type: dict   get_line_message: None
```

`loc` is empty and `input` is a reconstructed dict rather than a
`YamlOrderedDict`, so there is nothing structural to key on; scraping the name
out of the message text is exactly what we should not do. Meanwhile the branch
the finding names has no `files` to resolve *against* by construction, so the
proposed fix would change nothing. Fixing this properly means giving the
project-level validator structured context in `visivo/models/` — a separate
change, worth a follow-up. The existing assertion `errors[0]["file"] is None`
for a synthetic Pydantic model is correct, not an enshrined gap.

**`agent-contract.md` absent from `mkdocs/llms.txt` (LOW).** Agreed and
unactionable here: `llms.txt` is owned by the open #658. Flagging it as
requested — **#658 (or a follow-up) should add
`reference/agent-contract.md` to the Reference section of `mkdocs/llms.txt`**,
since an agent following that index is exactly the audience for this page.

---

### Falsification

Every guard proved to fail with its implementation removed:

```
CAUGHT  M1  machine-set strip goes back to name-only          -> 3 failed, 25 passed
CAUGHT  M2  omitted root fields go back to being dropped      -> 3 failed, 25 passed
CAUGHT  M3  step 5 (property-title strip) neutered            -> 1 failed, 27 passed
CAUGHT  M4  step 6 (unsatisfiable oneOf repair) neutered      -> 3 failed, 25 passed
CAUGHT  M5  subcommand detection goes back to argv[1]         -> 10 failed, 21 passed
CAUGHT  M6  profiling notices go back to stdout               -> 3 failed, 28 passed
CAUGHT  M7  'query saved to:' not recognised as a path        -> 2 failed, 18 passed
CAUGHT  M8  error keeps only its first line                   -> 1 failed, 19 passed
CAUGHT  M9  parse_job_message never extracts an error         -> 5 failed, 28 passed
CAUGHT  M10 run_phase always runs soft                        -> 1 failed, 15 passed
CAUGHT  M11 compile_result reads only the root collections    -> 1 failed, 19 passed
CAUGHT  M12 _NESTED_COLLECTIONS forgets model metrics         -> 2 failed, 18 passed
```

### Gates

- `pytest tests/ -q` — **2563 passed**, 2 skipped, 5 xfailed, 1 xpassed
- `black --check --target-version py312 .` — 550 files unchanged
- No viewer changes; no Pydantic model changes, so no schema regeneration
- Files touched vs. base, checked against the ownership manifest: only
  `mkdocs.yml` is listed, owned by the already-merged #644

jpjsonic:

Claude merged `main` and resolved the one conflict, in `run_phase.py`, where #666's artifact sweep and this branch's `defer_exit` both edit the end of the run.

**Conflict.** Both survive. The sweep runs as written; `defer_exit` still runs soft so `--json` can read `failed_job_results` and pick the exit code. One coherence fix: a run cut short by `SystemExit` no longer counts as clean, so the sweep stays gated on a run that actually finished (`run_completed`).

**--json purity.** The sweep logs through `Logger`, which writes to stdout — in `--json` mode that line would have landed inside the envelope. `json_command`'s `redirect_stdout` already catches it, verified in a real process (stdout is one JSON object; "Removed 2 artifact(s) no longer in the project" is on stderr), and the failure path skips the sweep and still emits the envelope. `test_run_json_keeps_the_artifact_sweep_off_stdout` now pins it, and fails both ways: with the redirect removed, and with the sweep disabled.

**Comment bar (#668),** scoped to this branch's diff. Ten comments deleted for restating the next line, labelling an obvious block, or narrating how a bug was found — including the `-p`-is-a-group-option block, replaced by naming the helper `log_off_stdout`, and the cross-PR note about #650. The rest tightened to the constraint they carry: Halo writing to stdout, `-h` not being a help alias (moved back beside the `QUIET_INVOCATION` it describes), the runner's log-line format, Pydantic's callable-discriminator `oneOf`, `Include.path`. Comment lines across the touched files: 191 → 140. Also corrected `test_phase`'s docstring, which claimed it returns the `TestRun` when it returns `(project, test_run)`.

Gates: `pytest tests/ -q` — 2581 passed, 2 skipped, 5 xfailed, 1 xpassed; `black --check --target-version py312 .` clean. No viewer files touched, so no viewer gates.

PR #664, opened 2026-08-29
Explorer: exploration results survive a tab switch (M27 tier 1)

jpjsonic:
> **Review round 2 (`1897fcaf`).** The first pass shipped a regression: threading
> the cache scope into `useModelTabPrefill` put it in the *same* dependency array
> as the pre-existing last-build parquet fetch, so a single keystroke mid-fetch
> cancelled that fetch permanently. Fixed by splitting the hook into two effects.
> Also fixed: editing a source in place was a silent stale serve; a real merge
> conflict with the still-open #659 in `SQLEditor.jsx`; an over-large cell budget;
> and a vacuous e2e assertion. Details in the review comment on this PR.

## The bug (M27)

Run a query in an exploration, glance at another one, come back — the grid is empty and every query has to be run again. A user who looks away loses their whole screen.

It is not a leak; it is a design gap. `ExplorationPane` parks the legacy `explorerStore` singleton on deactivate and rehydrates it from the persisted draft on activate, and that draft deliberately carries no rows — `snapshotExplorerWorkingState` strips `queryResult`/`enrichedResult` so the backend document stays a small JSON file. `restoreExplorerWorkingState` then rebuilds each model state from `createEmptyModelState`, whose `queryResult` is `null`. The rows only ever lived in memory, and nothing on the way back could get them.

## The fix — tier 1 only

Keep the rows **out** of the draft (that part is right) and park them in a new module-level, bounded LRU: **`viewer/src/stores/explorationResultCache.js`**.

- `CenterPanel` writes an entry when a run completes.
- A generalised `useModelTabPrefill` reads one back when a tab mounts with no result and hands it to `setModelQueryResult` — the same door a real run comes through, so `useExplorerDuckDB` re-registers the DuckDB table and computed columns / the column profiler keep working.
- A hit is a `Map` lookup. No network request at all (asserted in the e2e).

**Tier 2 is deliberately not here.** Surviving a hard reload needs a parquet twin under `target/`, and the dossier flags an open product decision — does the Django/cloud path get a twin? Without that call, tier 2 would ship a durability guarantee that only holds locally. The cache's docstring says out loud that its lifetime is one page session.

## The key, and why

```
[explorationId, modelName, sourceName, sql]   // JSON.stringify'd
```

| part | why it is in the key |
| --- | --- |
| `explorationId` | Isolation. Two explorations with an identically-named chip running identical SQL are two documents; neither may be shown the other's rows. |
| `modelName` | **A rename misses, deliberately.** The name does not determine the rows, but it determines what every downstream consumer (the DuckDB table, computed-column parenting, the promote checklist) thinks the rows *are*. That quiet mismatch costs more than the re-run it would save. |
| `sourceName` + `sql` | The actual determinants — taken from what was **executed**, never from the live buffer. |

`JSON.stringify` rather than a string join, so no name containing the separator can forge a different (exploration, chip) pair. `sql` is compared exactly (end-trimmed only, matching what `SQLEditor.handleRun` sends) — no normalisation, no parsing; a whitespace-only edit misses, which is the correct side to err on.

Getting "what was executed" honest required one small change to `SQLEditor`: it already snapshotted `queryContext` at run time so a mid-flight tab switch couldn't misroute the result — it now snapshots the SQL and source alongside it and returns them through `onQueryComplete`. The live buffer is the wrong key: the user may have run a **selection**, or typed on while the job was in flight.

## What invalidates

Three change the key, so the read simply misses — nothing to do:

- editing the SQL
- changing the source
- renaming the query chip

A rename's stranded entry is left to the LRU rather than invalidated: an orphan is by definition never read, so it never moves to the recently-used end and is **always evicted ahead of a live entry**. Reaching it would mean teaching `explorerStore.renameModelTab` — the legacy singleton, which knows nothing about which exploration is on screen — an exploration identity, and that coupling costs more than the slot. A test pins the LRU ordering that makes it safe.

Four the key cannot express, so they are told explicitly:

- **A re-run** drops every other entry for that (exploration, chip) before inserting. One live result per chip, always the newest — so re-running a second query and then typing the first one's text back in is a miss, not a resurrection of rows from before the re-run.
- **A failed run** retires that chip's rows. The same SQL erroring now (a dropped table, a source that went away) is exactly when its old rows must not come back.
- **Editing or deleting the SOURCE** (round 2). The key identifies a source by *name*, and a name is not a connection: repoint `local` from `a.duckdb` to `b.duckdb` — or change its schema, host, or credentials — and every key still matches while the rows came out of a database the user has stopped pointing at. Nothing reads `from_cache`, so there is no signal either; the grid would simply be wrong. `sourceStore.saveSource`/`deleteSource` now call `invalidateResultsForSource`, scoped to that source (editing one connection must not cost a user every result they have), and a save that *fails* keeps the rows because nothing changed.
- **Promoting a model** — and deleting the exploration — drops the entries. Once the model is a project object its rows belong to the project's build (`useModelTabPrefill`'s last-build path), and the promoted config is the checklist's, not necessarily the raw text the chip last ran.

## What bounds it

`8` entries and `250,000` cells (`rows × columns`, an O(1) proxy for bytes — serialising a 100k-row result just to measure it costs more than the cache saves). Least-recently-**used** evicted first, with a read counting as a use, so the tab you keep coming back to is the last one dropped. A single result too large for the whole budget is refused outright rather than emptying the cache for something that still would not fit.

The cell number is **sized, not guessed** (round 2 — it was 1,000,000, described in the docstring as "deliberately small", which it was not). An entry retains the run's own row objects — the cache stores the *same* array the store was handed, so a parked result costs one copy, not two — at roughly 100–200 bytes per cell for JS objects with string keys. 250k cells is therefore ~25–50 MB at full budget: eight chips of a few thousand rows, or one 25k × 10 result. There is no server-side row cap on `/api/model-query-jobs/`, so a `SELECT *` on a large table really can exceed it; that result is simply not cached, and the Run button remains the way back to it.

These bounds are the **only** thing that limits the cache — it is not cleared on leaving the Explorer, deliberately: the navigation it must survive (park an exploration, read another, come back) is indistinguishable at the component level from the navigation that would clear it, so a clear-on-exit hook that fired on the wrong one would delete the feature.

## Tests

- `explorationResultCache.test.js` — 33 unit tests: hit, miss-on-edit, miss-on-rename, miss-on-source-change, cross-exploration isolation, separator forging, re-run supersession, invalidation, LRU eviction at the **shipped** entry cap, cell-budget eviction, read-as-use recency, oversize refusal, budget accounting.
  Round 2 adds: source-scoped invalidation and its two negatives; the cell bound exercised **at the shipped value** rather than only at an injected `maxCells: 100` (holey `new Array(n)` makes a quarter-million cells free to allege, since `cellsIn` only reads `.length`); and the LRU ordering that makes a renamed chip's orphan safe, modelled on the real sequence (orphan inserted *after* the live entry, so insertion order alone would evict the wrong one).
- `useModelTabPrefill.test.js` — 14 new: restores with no network call, restores before DuckDB is ready, every miss falls through to the last-build path, never overwrites live rows, off entirely outside an exploration (the pre-M27 contract for existing callers).
  Round 2 adds the regression suite: a keystroke mid-fetch and a source rebind mid-fetch must not cancel the last-build prefill (both use a **deferred** `fetchModelJobs` — without one the fetch resolves before the rerender and nothing is proven); a cache hit still suppresses the build fetch entirely; the handoff ref does not latch after an invalidation; and the mid-typing restore is pinned as a deliberate affordance rather than left implicit.
- `sourceStore.resultCache.test.js` — **new**, 4 tests: `saveSource`/`deleteSource` retire that source's results, other sources are untouched, and a save that fails keeps the rows.
- `CenterPanel.test.jsx` — 8 new, driving the **full** round trip through the real store and the real cache: run → unmount → rebuild the model state from the draft (exactly what activate does) → assert the rendered grid. Includes "parks the rows under the query that RAN, not the tab that is active when they land".
- `SQLEditor.test.jsx` — 2 new: reports the selection that ran; reports the SQL/source that ran even after the user keeps typing mid-flight.
- `workspaceExplorationsStore.test.js` — 5 new: promote/delete invalidation, and the negatives (a promote that fails, an insight promote, a delete that fails all leave the rows alone).
- **New e2e story** `exploration-result-continuity.spec.mjs` (3 tests, registered in both `exploration-mutations`'s `testMatch` and `parallel`'s `testIgnore` per the double-registration rule).

### Gates

| gate | result |
| --- | --- |
| `yarn test --watchAll=false` | **368 suites / 7133 tests passed** |
| `yarn lint` | **0 errors** |
| e2e `exploration-result-continuity` (sandbox :8061/:3061) | **3 passed** |
| `black --check --target-version py312 .` | 543 files unchanged |
| Python | **no Python files touched** — tier 1 is viewer-only |
| `git merge-tree` vs #659 | **clean** (was conflicting in `SQLEditor.jsx` — see Conflict discipline) |

A wider run of the whole serial `exploration-mutations` project (well beyond this lane's gates) shows failures across `exploration-dnd-pull-in`, `exploration-preview`, `dashboard-newchart` and others. **They are not this PR's.** Setting `explorationId={null}` in `ExplorationWorkbench` switches the M27 cache off entirely — at which point `useModelTabPrefill` behaves exactly as `origin/main` — and `exploration-dnd-pull-in` fails with the identical four tests. Those specs are red in this sandbox independently of M27.

### Falsification

Every guard was removed one at a time, tests kept, and the failure recorded. All 16 mutations failed; the restored tree is green (289 tests across the 7 touched suites).

| mutation | tests that failed |
| --- | --- |
| key drops `modelName` | 10, incl. `the key › MISSES after the chip is renamed`, `session result cache › MISSES after the chip is renamed`, `CenterPanel › … after the chip is renamed` |
| key drops `sql` | 5, incl. `the key › MISSES after the query is edited`, `re-running › retires the chip's PREVIOUS query…`, `CenterPanel › … after the SQL is edited` |
| key drops `explorationId` | 4, incl. `the key › never serves one exploration the rows of another`, `CenterPanel › never shows a sibling exploration the rows` |
| key drops `sourceName` | 3, incl. `the key › MISSES after the source is changed` |
| re-run no longer drops siblings | `retires the chip's PREVIOUS query, so reverting the text is a miss` (+ budget accounting) |
| eviction removed | `evicts the least recently used entry at the shipped entry cap`, `evicts on the cell budget…`, `counts a READ as a use…` |
| read no longer bumps recency | `counts a READ as a use, so the tab you keep returning to survives` |
| oversize refusal removed | `refuses a single result too big for the whole budget…` |
| hook never reads the cache | 5, incl. `restores the parked rows on return, with no network call at all`; **and the e2e** `run a query, park the tab, come back — same rows, no re-run` |
| failed-run invalidation removed | `does not resurrect rows once the same query starts failing` |
| cache keyed on the live buffer | `parks the rows under the query that RAN, not the tab that is active when they land` |
| `SQLEditor` stops snapshotting the run | 3, incl. `reports the SELECTION that ran, not the whole buffer` |
| promote invalidation removed | `promoting a model retires that chip's parked rows` |
| delete invalidation removed | `deleting an exploration drops every result parked for it` |
| pane stops passing `explorationId` | `hands the workbench this exploration's id` |
| workbench stops passing `explorationId` | `hands CenterPanel the exploration id so parked results stay per-document` |

**Round 2** adds nine more, same method — one implementation detail stashed at a time, tests untouched, restored green (58 tests across the three suites):

| mutation | tests that failed |
| --- | --- |
| collapse the two effects back into one (**the regression**) | `survives a keystroke in the SQL editor while the parquet is loading`, `survives the source being rebound while the parquet is loading` |
| drop the `servedFromCacheRef` handoff | `a cache HIT still suppresses the build fetch entirely`, `restores the parked rows on return, with no network call at all`, `still restores on a SECOND return…` |
| let `servedFromCacheRef` latch | `falls back to the build after a hit is invalidated` |
| stop invalidating on `saveSource` | `saveSource retires the results that source produced`, `leaves results from OTHER sources alone` |
| stop invalidating on `deleteSource` | `deleteSource retires them too` |
| make source invalidation a panic-clear | `drops every result it produced, across explorations`, `leaves results from OTHER sources alone` |
| drop the oversize refusal (at the **shipped** `maxCells`) | `at the SHIPPED cell budget › refuses one cell more, rather than emptying itself for it` |
| stop evicting on the cell budget (at the **shipped** `maxCells`) | `at the SHIPPED cell budget › evicts on the shipped budget, not just the injected one` |
| stop counting a read as a use | `counts a READ as a use…`, `evicts a renamed chip's orphaned entry ahead of a live one` |

The e2e was falsified against the live sandbox rather than argued about. With `explorationId` deleted from the key: the **old** test 3 still passed (it was vacuous — the sibling's chip had an empty buffer, and empty sql switches caching off outright), the **new** test 3 failed. With the key correct, all three pass.

## Conflict discipline

None of the 16 open-PR files were touched. `workspaceExplorationsStore.js` takes two lines (one inside the existing promote loop, one in `deleteExploration`) plus an import; `ExplorationPromoteModal.jsx`, `promoteChecklist.js`, `Chart.jsx`, `ExplorerChartPreview.jsx`, `explorerStore.js` and everything under `visivo/` are untouched.

Round 2 went further and removed a *textual* conflict that was not on any owned-file list. This branch and the still-open #659 both touch `SQLEditor.jsx` and `sourceStore.js`:

```
$ git merge-tree --write-tree <round-1 tip> <#659 tip>
CONFLICT (content): Merge conflict in viewer/src/components/explorer/SQLEditor.jsx
exit=1

$ git merge-tree --write-tree <round-2 tip> <#659 tip>
exit=0                                       # clean, zero conflicts
```

Both PRs were adding a ref immediately after `runContextRef` and a statement immediately after `executeQuery`. Nothing here needed those anchors: the new refs are declared **above** the `runContextRef` comment and assigned **above** its statement, and the `sourceStore` import goes above the existing one — leaving #659's context lines byte-identical to main. #659's patch now applies to this branch with plain `git apply`, and the merged file carries both PRs' additions correctly. No behaviour changed to achieve it.

Linear: VIS-1309.

### DIFF
--- a/viewer/e2e/stories/exploration-result-continuity.spec.mjs
+++ b/viewer/e2e/stories/exploration-result-continuity.spec.mjs
@@ -23,7 +23,11 @@
  *      to remove.
  *   2. Edit the SQL first, and coming back shows the run-your-query state
  *      rather than rows that answer a question the user has stopped asking.
- *   3. A sibling exploration never sees them.
+ *   3. A sibling exploration never sees them — asserted with the SAME SQL
+ *      typed into the sibling's chip, so every part of the cache key matches
+ *      except the exploration. Left with the sibling's default empty buffer
+ *      the assertion is vacuous: an empty query switches caching off outright,
+ *      so the grid would be empty whether or not the key isolates.
  *
  * Tier 2 (surviving a hard RELOAD, via a parquet twin under `target/`) is
  * deliberately out of scope and unasserted here — see the cache module's
@@ -175,15 +179,28 @@ test.describe('Exploration results survive a tab switch (M27)', () => {
     await gotoExplorerHome(page);
     await newExploration(page);
 
-    await typeSql(page, `SELECT * FROM ${TABLE}`);
+    const sql = `SELECT * FROM ${TABLE}`;
+    await typeSql(page, sql);
     await runQuery(page);
     await expect(resultsGrid(page)).toBeVisible({ timeout: 20000 });
 
     await backToHome(page);
     await newExploration(page);
 
-    // A brand-new exploration whose chip may well carry the same generic
-    // name: its grid is empty until IT runs something.
+    // TYPE THE SAME SQL, and do not run it. This is what makes the test an
+    // isolation test rather than a coincidence: a brand-new exploration's
+    // auto-created chip starts with an EMPTY buffer, and
+    // `explorationResultCacheKey` returns null for empty sql — so caching is
+    // off for that tab no matter what the key is made of, and an empty grid
+    // would prove nothing. With the same SQL in the buffer, the sibling's
+    // chip carries the same generic name against the same source running the
+    // same text, and every component of the key MATCHES except
+    // `explorationId`. `useModelTabPrefill` reads on the keystroke, so if the
+    // exploration were ever dropped from the key this grid would fill with
+    // the other document's rows — which is the failure that must be
+    // impossible.
+    await typeSql(page, sql);
+
     await expect(emptyResults(page)).toBeVisible({ timeout: 20000 });
     await expect(resultsGrid(page)).toHaveCount(0);
   });
--- a/viewer/e2e/stories/exploration-result-continuity.spec.mjs
+++ b/viewer/e2e/stories/exploration-result-continuity.spec.mjs
@@ -1,0 +1,190 @@
+/**
+ * Story: exploration results survive a tab switch (M27, tier 1).
+ *
+ * THE BUG the field test reported: run a query, glance at another
+ * exploration, come back — the grid is empty and every query has to be run
+ * again. `ExplorationPane` parks the legacy `explorerStore` singleton on
+ * deactivate and rehydrates it from the persisted draft on activate, and that
+ * draft deliberately carries no `queryResult` (it is meant to stay a small
+ * JSON document, `explorerStore.js`'s snapshot docstring). So the rows were
+ * simply gone.
+ *
+ * THE FIX: `viewer/src/stores/explorationResultCache.js` — a page-lifetime,
+ * bounded LRU keyed by exploration + query chip + source + the SQL that
+ * actually ran. `CenterPanel` writes on run completion; `useModelTabPrefill`
+ * reads on mount and hands the rows back through `setModelQueryResult`.
+ *
+ * The three tests below are the three claims that matter, and two of them are
+ * refusals:
+ *
+ *   1. Come back and the rows are there — with NO new query job started. The
+ *      network assertion is the real one: "the grid has rows" could also be
+ *      satisfied by silently re-running, which is the very cost this exists
+ *      to remove.
+ *   2. Edit the SQL first, and coming back shows the run-your-query state
+ *      rather than rows that answer a question the user has stopped asking.
+ *   3. A sibling exploration never sees them.
+ *
+ * Tier 2 (surviving a hard RELOAD, via a parquet twin under `target/`) is
+ * deliberately out of scope and unasserted here — see the cache module's
+ * docstring for why that half needs a product decision first.
+ *
+ * Precondition: sandbox running (integration project), e.g.
+ *   VISIVO_SANDBOX_NAME=resultContinuity VISIVO_SANDBOX_BACKEND_PORT=8061 \
+ *   VISIVO_SANDBOX_FRONTEND_PORT=3061 bash scripts/sandbox.sh start
+ *   PLAYWRIGHT_BASE_URL=http://localhost:3061 npx playwright test exploration-result-continuity
+ *
+ * Mints/mutates real backend exploration records — runs in the serial
+ * `exploration-mutations` playwright project (playwright.config.mjs), never
+ * `parallel`. See the DOUBLE-REGISTRATION RULE note in playwright.config.mjs:
+ * this filename must appear in BOTH `exploration-mutations`'s `testMatch` and
+ * `parallel`'s `testIgnore`.
+ */
+
+import { test, expect } from '@playwright/test';
+import { typeSql, runQuery } from '../helpers/explorer.mjs';
+import { BASE_URL, API } from '../helpers/sandbox.mjs';
+
+const TABLE = 'test_table';
+
+async function listExplorationIds(page) {
+  const res = await page.request.get(`${API}/api/explorations/`).catch(() => null);
+  if (!res || !res.ok()) return [];
+  const data = await res.json().catch(() => []);
+  return (data || []).map(e => e.id);
+}
+
+async function gotoExplorerHome(page) {
+  await page.goto(`${BASE_URL}/workspace/exploration`);
+  await page.waitForLoadState('networkidle');
+  await expect(page.getByTestId('workspace-middle-explorer')).toBeVisible({ timeout: 30000 });
+}
+
+async function newExploration(page) {
+  await page.getByTestId('explorer-home-new-exploration').click();
+  await expect(page.getByTestId('workspace-middle-exploration')).toBeVisible({ timeout: 30000 });
+  // `useExplorerWorkbenchInit`'s "auto-create a model tab when empty" effect
+  // lands asynchronously on a cold load; typing before it does silently
+  // vanishes (`setActiveModelSql` no-ops with no active model).
+  await page.waitForFunction(() => !!window.useStore.getState().explorerActiveModelName, {
+    timeout: 10000,
+  });
+  await page.waitForURL(/\/workspace\/exploration\/exp_/, { timeout: 10000 });
+  return new URL(page.url()).pathname.split('/').pop();
+}
+
+/** Back to Explorer Home, so a second exploration can be minted. */
+async function backToHome(page) {
+  await page.getByTestId('workspace-view-switcher-explorer').click();
+  await expect(page.getByTestId('explorer-home-gallery')).toBeVisible({ timeout: 30000 });
+}
+
+async function activateTab(page, id) {
+  await page.getByTestId(`workspace-tab-select-exploration:${id}`).click();
+  await expect(page.getByTestId('workspace-middle-exploration')).toBeVisible({ timeout: 30000 });
+  await page.waitForFunction(
+    explorationId => window.location.pathname.endsWith(explorationId),
+    id,
+    { timeout: 10000 }
+  );
+}
+
+const resultsGrid = page => page.getByTestId('explorer-results-grid');
+const emptyResults = page => page.getByTestId('empty-results');
+
+// The grid virtualises its rows into absolutely-positioned divs (no <tbody>),
+// so the honest "how much data is on screen" signal is DataTable's own footer
+// count rather than a DOM row tally.
+const rowCountLabel = page => resultsGrid(page).getByText(/total rows/);
+
+test.describe('Exploration results survive a tab switch (M27)', () => {
+  // Multi-step: mint two explorations, run a real query job, switch twice.
+  // The 30s global default is tight for that once this project's
+  // combined-load contention with `parallel` is in play.
+  test.describe.configure({ timeout: 90000 });
+
+  let idsBeforeTest = [];
+
+  test.beforeEach(async ({ page }) => {
+    idsBeforeTest = await listExplorationIds(page);
+  });
+
+  test.afterEach(async ({ page }) => {
+    const idsAfterTest = await listExplorationIds(page);
+    for (const id of idsAfterTest.filter(id => !idsBeforeTest.includes(id))) {
+      await page.request.delete(`${API}/api/explorations/${id}/`).catch(() => {});
+    }
+  });
+
+  test('run a query, park the tab, come back — same rows, no re-run', async ({ page }) => {
+    await gotoExplorerHome(page);
+    const idA = await newExploration(page);
+
+    await typeSql(page, `SELECT * FROM ${TABLE}`);
+    await runQuery(page);
+    await expect(resultsGrid(page)).toBeVisible({ timeout: 20000 });
+    const rowsBefore = await rowCountLabel(page).textContent();
+    expect(rowsBefore).toMatch(/^[1-9]/); // real rows landed, not an empty result
+
+    // Park A by activating a second exploration, exactly as a user reading
+    // something else would.
+    await backToHome(page);
+    const idB = await newExploration(page);
+    expect(idB).not.toBe(idA);
+
+    // Count query jobs started from HERE on. A silent re-run would satisfy
+    // "the grid has rows" just as well as a cache hit does — and it is the
+    // cost this whole feature exists to remove, so it is what gets asserted.
+    const startedJobs = [];
+    page.on('request', req => {
+      if (req.method() === 'POST' && req.url().includes('/api/model-query-jobs/')) {
+        startedJobs.push(req.url());
+      }
+    });
+
+    await activateTab(page, idA);
+
+    await expect(resultsGrid(page)).toBeVisible({ timeout: 20000 });
+    await expect(rowCountLabel(page)).toHaveText(rowsBefore, { timeout: 10000 });
+    expect(startedJobs).toEqual([]);
+  });
+
+  test('edit the SQL first, and coming back asks you to run it — it does not show the old rows', async ({
+    page,
+  }) => {
+    await gotoExplorerHome(page);
+    const idA = await newExploration(page);
+
+    await typeSql(page, `SELECT * FROM ${TABLE}`);
+    await runQuery(page);
+    await expect(resultsGrid(page)).toBeVisible({ timeout: 20000 });
+
+    // The query changes. These rows now answer a question the user has
+    // stopped asking, and nothing on screen would say so.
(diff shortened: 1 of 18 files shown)


jpjsonic:

Merged `main` in — #656 (M28, the chart-fills-the-pane lane) landed after this branched and touched two of the same files.

Both were mechanical:

- **`viewer/playwright.config.mjs`** — a real textual conflict, but only because both lanes appended a new spec to the same two lists. Kept both registrations (`exploration-result-continuity` and `explorer-chart-fills-pane`) in `parallel`'s `testIgnore` and `exploration-mutations`'s `testMatch`, per the double-registration rule.
- **`viewer/src/components/explorer/CenterPanel.jsx`** — auto-merged, disjoint regions: #656 added the `chart-preview-pane` testid inside `renderChartSection`; this lane's changes are the `explorationId` prop, `handleQueryComplete`, and the `useModelTabPrefill` call.

Gates re-run on the merge commit: **7117 tests / 367 suites passed**, `yarn lint` clean.

jpjsonic:

## Review round 2 — `1897fcaf`

Nine findings triaged. **Three real defects fixed, one merge hazard removed, three rejected with reasons.** Every claim below was reproduced by execution before it was acted on.

---

### 1. The regression this PR introduced (high) — FIXED

**Confirmed by execution.** Threading the cache scope into `useModelTabPrefill` put `sourceName`/`sql` in the *same* dependency array as the last-build parquet fetch — an async fetch that `attemptedRef` allows exactly once per model per mount. `CenterPanel` feeds `sql` from `selectActiveModelSql`, which `SQLEditor`'s `onSave` updates on **every keystroke**; `sourceName` moves again when VIS-1082's `applyResolvedDefaultSource` rebinds an auto-created tab once `defaults` land.

Either one landing mid-fetch ran the cleanup (`cancelled = true`), discarded the resolved response, and then hit the `attemptedRef` early return on re-entry. The grid stayed empty for the rest of the mount, beside a parquet that exists. **Pre-M27 it showed the rows.** A probe with a deferred `fetchModelJobs` failed on the reviewed tip and passed with only the scope removed from the deps — the reviewer's diagnosis was exactly right.

Fixed by splitting the hook into two effects with the triggers each actually has:

- **cache read** — keyed on the live scope, because it *must* re-fire when the source is rebound or the buffer changes;
- **build fetch** — back on its pre-M27 dependency list, untouched by the scope.

`servedFromCacheRef` is the handoff. Effects run in declaration order, so on the commit where the cache hits, the build effect sees the ref rather than racing a parquet against the rows just restored (it is still holding that render's stale `hasResult === false`). A **miss clears the ref**, so an entry invalidated later still falls through to the build instead of leaving the tab with nothing.

### 2. Editing a source in place was a silent stale serve (medium) — FIXED

The key identifies a source by **name**, and a name is not a connection. Repoint `local` from `a.duckdb` to `b.duckdb` — or change its schema, host, or credentials — and every parked key still matches while the rows came out of a database the user has stopped pointing at. Nothing reads `from_cache` (grep finds only the write site), so there is no signal either: the grid would just be wrong, with no re-run.

New `invalidateResultsForSource(sourceName)`, called from `sourceStore.saveSource` and `deleteSource`. Scoped by source, not a panic-clear — editing one connection must not cost a user every result they have. A save that **fails** keeps the rows, because nothing changed.

### 3. Merge hazard with #659 — REMOVED (no revert needed)

`SQLEditor.jsx` is not on this run's owned-file list (#659's entry there is `onboarding/**`, `Project.jsx`, `telemetry/**`, `data_views.py`), so this was not a forbidden-file hit — **but the conflict was real** and worth removing rather than arguing about:

```
$ git merge-tree --write-tree c7404970 <#659 tip>     # the reviewed tip
CONFLICT (content): Merge conflict in viewer/src/components/explorer/SQLEditor.jsx
exit=1

$ git merge-tree --write-tree 1897fcaf <#659 tip>     # after this commit
exit=0                                                # clean, zero conflicts
```

Nothing here needed #659's anchors. The new refs are declared **above** the `runContextRef` comment and assigned **above** its statement, leaving #659's three context lines byte-identical to main. #659's patch now applies to this branch with plain `git apply` — no 3-way, no conflict — and the merged file carries both PRs' refs and both statements correctly. Behaviour unchanged; `SQLEditor.test.jsx` already merged clean and still does.

### 4. Bounds and honesty (low) — FIXED

- `DEFAULT_MAX_CELLS` was documented as *"deliberately small"* at 1M cells, which is 50–150 MB of row objects retained for the life of the page. It is **250k** now, with the arithmetic written down and the no-server-row-cap caveat stated. The docstring also says plainly why there is **no clear-on-exit**: the navigation this must survive (park an exploration, read another, come back) is indistinguishable at the component level from the navigation that would clear it, so the bounds have to stand on their own.
- The shipped cell bound is now exercised **at its real value**, not only at an injected `maxCells: 100`. Holey `new Array(n)` makes that free — `cellsIn` only reads `.length`, so a quarter-million cells costs nothing to allege.

### 5. The vacuous e2e (low) — FIXED, and the vacuousness proved

The reviewer was right, and it is now demonstrated rather than asserted. With `explorationId` deleted from the key:

| | old test 3 | new test 3 |
|---|---|---|
| falsified key | **passes** (vacuous) | **fails** |
| correct key | passes | passes |

The sibling's auto-created chip starts with an **empty** buffer, and `explorationResultCacheKey` returns `null` for empty sql — caching was off for that tab no matter what the key contained. It now types the **same SQL**, so chip name, source, and text all match and the exploration is the only difference. All 3 specs pass against a live sandbox on the committed code.

---

### Rejected, with reasons

- **Renamed-chip orphans consuming the budget** — real, but left to the LRU on purpose, now written into the module docstring. An orphan is by definition never read, so it never moves to the recently-used end and is **always evicted ahead of a live entry**; it serves nothing wrong and costs a slot until pressure arrives. Fixing it properly means teaching `explorerStore.renameModelTab` — the legacy singleton, which knows nothing about which exploration is on screen — an exploration identity, and that coupling is a larger correctness risk than the slot is worth. The other rename call site is `ExplorationPromoteModal`, which this run may not touch. The LRU ordering that makes this safe is now pinned by a test modelling the real rename sequence (orphan inserted *after* the live entry, so insertion order alone would evict the wrong one).
- **`playwright.config.mjs` / `CenterPanel.jsx` ownership** — spent. #656 merged as `911583a4`, which *is* `origin/main` (`gh pr view 656` → `MERGED`, mergeCommit `911583a4`). No action, correctly flagged as already discharged.
- **Rows appearing mid-typing** — kept, and now tested rather than left implicit. `put` holds at most one entry per `(exploration, chip)`, always the newest run's, so a key match means these really are that chip's rows for exactly that source and exactly that text — the same rows a tab switch would restore. It errs on the safe side of the invalidation contract; the surprise is worth a test that says so out loud, not a behaviour change.

---

### Gates

| | |
|---|---|
| Viewer suite | **368/368 suites, 7133 tests** pass |
| Viewer lint | clean, 0 errors |
| e2e `exploration-result-continuity` | **3/3** against a live sandbox (`:8061`/`:3061`, integration project) |
| Falsification | **9 cases**, each stashing one implementation detail with tests untouched — every one fails the tests that name it, all green on restore |
| Python | not touched (0 `.py` files in the diff) |
| `git merge-tree` vs #659 | clean (was conflicting) |

**On the wider e2e picture, honestly:** a run of the whole serial `exploration-mutations` project (well beyond this lane's gates, and not run in round 1 either) is red in `exploration-dnd-pull-in`, `exploration-preview`, `dashboard-newchart` and others. Those are **not this PR's**. Setting `explorationId={null}` in `ExplorationWorkbench` switches the M27 cache off entirely — at which point `useModelTabPrefill` is behaviourally identical to `origin/main` — and `exploration-dnd-pull-in` fails with the identical four tests. Worth someone's attention, but not this branch's to fix.

<details><summary>Falsification output</summary>

```
CASE 1 — collapse the two effects back into one (the M27 regression)
  ● survives a keystroke in the SQL editor while the parquet is loading
  ● survives the source being rebound while the parquet is loading
CASE 2 — drop the servedFromCacheRef handoff
  ● a cache HIT still suppresses the build fetch entirely
  ● restores the parked rows on return, with no network call at all
  ● still restores on a SECOND return, unlike the once-per-mount build fetch
CASE 3 — let servedFromCacheRef latch (never cleared on a miss)
  ● falls back to the build after a hit is invalidated
CASE 4 — stop invalidating on saveSource
  ● saveSource retires the results that source produced
  ● leaves results from OTHER sources alone
CASE 5 — stop invalidating on deleteSource
  ● deleteSource retires them too
CASE 6 — make source invalidation a panic-clear (drop the predicate)
  ● drops every result it produced, across explorations
  ● leaves results from OTHER sources alone
CASE 7 — drop the 'too big for the whole budget' refusal  [at the SHIPPED maxCells]
  ● at the SHIPPED cell budget › refuses one cell more, rather than emptying itself for it
CASE 8 — stop evicting on the cell budget  [at the SHIPPED maxCells]
  ● at the SHIPPED cell budget › evicts on the shipped budget, not just the injected one
CASE 9 — stop counting a READ as a use
  ● counts a READ as a use, so the tab you keep returning to survives
  ● evicts a renamed chip’s orphaned entry ahead of a live one
BASELINE / RESTORED: 58 passed, 58 total
```
</details>

tgsoverly (approved):
Good without the comments.

PR #668, opened 2026-08-30
Document a stricter comment bar in CLAUDE.md

tgsoverly:
Per direct feedback: comments across recent PRs were too verbose — restating logic in prose that a reader has to re-derive later, rather than letting code and tests speak for themselves.

Elevates the existing Python-only "code comments" note to a top-level, cross-language section:

- Default to no comment. Prefer a clearer name, an extracted function, or a test case.
- Add one only when the reasoning can't be inferred from code/tests, or the logic is genuinely complex.
- Keep it to a line or two even then.
- Applies when touching existing comments too, not just new ones.

No functional change. My three other open PRs (#619, #666, #667) are getting the same treatment applied to their diffs as a follow-up.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -2,6 +2,16 @@
 
 This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
 
+## Code Comments
+
+Default to no comment. Prefer a clearer name, an extracted function, or a test case that pins the behavior — those stay true as the code changes; a comment restating logic in prose does not.
+
+Add a comment only when:
+- the reasoning can't be inferred from the code or tests (a non-obvious external constraint, a workaround for a specific upstream bug), or
+- the logic is genuinely complex enough that a reader needs a map before diving in.
+
+When one is justified, keep it to a line or two — state the constraint, not the history of how it was found. This applies to Python and JS alike, and to comments already in a PR as much as new ones: when touching a file, trim existing comments down to this bar rather than leaving them.
+
 ## Development Commands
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -2,6 +2,16 @@
...
 
+## Code Comments
+
+Default to no comment. Prefer a clearer name, an extracted function, or a test case that pins the behavior — those stay true as the code changes; a comment restating logic in prose does not.
+
+Add a comment only when:
+- the reasoning can't be inferred from the code or tests (a non-obvious external constraint, a workaround for a specific upstream bug), or
+- the logic is genuinely complex enough that a reader needs a map before diving in.
+
+When one is justified, keep it to a line or two — state the constraint, not the history of how it was found. This applies to Python and JS alike, and to comments already in a PR as much as new ones: when touching a file, trim existing comments down to this bar rather than leaving them.
```
jpjsonic, line 13: I think this history bit is really important. The most useless aspect of the comments is the broken history lesson that it tried to inject. Do we want to limit it to python and JS? 

I wonder if we should include some language in here for commit messages and PR descriptions too to make them use more focused language

jpjsonic (approved):

CLAUDE.md
tgsoverly: The history should be contained in the PR, and that can be examined.  Just thinking on how a human would do it.  Code is minimal and self-explanatory; if you need more, you can dig for it.

PR #688, opened 2026-09-18
Include what a selected consumer reads in a `+name+` filter

tgsoverly:
## The bug

`+name+` selects a node, its ancestors and its descendants. The descendants are consumers that will be rebuilt — but a consumer that reads a *second* model was selected without that model, so it ran with nothing to read.

**It only shows up when the run starts from an empty directory.** Locally `target/` still holds the previous run's outputs, so the missing dependency is quietly satisfied by a file on disk. A cloud run builds in a fresh `TemporaryDirectory` (`core: runner/entrypoint.py:983`), where the same filter fails.

## How it was found

A project whose insights filter on an input:

```yaml
models: [restaurants, station-counts, cuisine-list]
inputs:  [cuisine-select]          # reads cuisine-list
insights:
  - restaurant-markers             # charts restaurants, filters on cuisine-select
  - station-bubbles                # charts station-counts, filters on cuisine-select
```

Editing the input makes core compute `dag_filter=+cuisine-select+` (`api/apps/deploys/services/changes.py:195`). That selects `cuisine-list` (its ancestor) and both insights (its descendants) — but neither `restaurants` nor `station-counts`, which are ancestors of the *insights*, not of the input.

Reproduced exactly outside the cloud:

```
$ visivo run -df "+cuisine-select+"     # in an empty directory
Wrote schema for model cuisine-list                    [SUCCESS]
Schema file not found for model 'station-counts'
Failed job for insight station-bubbles: Missing schema for model: station-counts.
Schema file not found for model 'restaurants'
Failed job for insight restaurant-markers: Missing schema for model: restaurants.
```

The models are *absent* rather than failed — no job was ever created for them — which is what makes this confusing to read. The error names the insight, and the actual problem is a scope that cannot run.

After the change, same command, same empty directory:

```
Updated data & wrote schema for model station-counts   [SUCCESS]
Updated data & wrote schema for model restaurants      [SUCCESS]
Updated data for insight restaurant-markers            [SUCCESS]
Updated data for insight station-bubbles               [SUCCESS]
Run finished in 4.43s
```

## Scope of the change

Closed over the ancestor side **only when it is unbounded** (`+name`), which already means "and everything this needs".

A bounded radius is an explicit limit and is left alone, so `-df model`, `-df +model`, `-df 1+model+1` and every other bounded form behave exactly as before. All 15 existing `test_filter_dag` assertions pass unchanged — the first version of this patch closed over every form and broke `1+model+1`, which is what pointed at the narrower rule.

## Tests

- `test_filter_dag_includes_what_a_selected_consumer_reads` — two models feeding one consumer; fails without the change, passes with it
- `test_filter_dag_respects_a_bounded_radius` — a bounded filter does **not** widen; passes either way, and is there to stop this being "fixed" into a full closure later

2598 passed, `black --check` clean.

## Not addressed

The DAG itself is correct — the input's children are identified properly, which is why the insights were selected at all. This is only about a filter that selects a node without what it reads.

Separately, that project's DuckDB source reports `0 tables, 0 columns` because introspection lists attached-catalog tables by bare name and then `DESCRIBE`s them unqualified. Unrelated to this, and rarer.


tgsoverly:
Superseded by the runner fix — see the follow-up PR. The filter is for showing an object's graph; widening it to make a run succeed changed what it means. Whether an unbuilt dependency has to run is the runner's decision, and that is where it now lives.

PR #689, opened 2026-09-18
Build a dependency the filter left out, when nothing has built it

tgsoverly:
Replaces #688, which fixed this in the dag filter. That was the wrong layer — @tgsoverly's point: the filter is there to show you an object's graph, and widening it to make a run work changes what it means. The models did need to run, but deciding *that* is the runner's job.

## The bug

`+input+` selects the insights downstream of an input but not the models those insights chart. The runner started them anyway — in the filtered dag an insight whose model was filtered out has out-degree 0, so it looks ready — and they failed with `Missing schema for model: <name>`.

**Only against an empty directory.** A local `target/` still holds the previous run's outputs, so the insight reads its model from disk and nothing looks wrong. A cloud run builds in a fresh `TemporaryDirectory` (core `runner/entrypoint.py:983`), where it fails every time.

## The change

`DagRunner` adds back the dependencies of what it was given — but only the ones nothing has built yet, which is the condition the user described: *not changed, but also has no runs, so it has to run.*

The filter is untouched and keeps meaning exactly what it says.

Cold and warm, same filter, on the project that surfaced this:

```
$ visivo run -df "+cuisine-select+"     # fresh directory
Updated data & wrote schema for model station-counts   [SUCCESS]
Updated data & wrote schema for model restaurants      [SUCCESS]
Updated data for insight restaurant-markers            [SUCCESS]
Updated data for insight station-bubbles               [SUCCESS]
Run finished in 4.36s

$ visivo run -df "+cuisine-select+"     # again, now warm
Wrote schema for model cuisine-list                    [SUCCESS]
Updated data for insight restaurant-markers            [SUCCESS]
Updated data for insight station-bubbles               [SUCCESS]
Run finished in 1.88s
```

The second run leaves the models alone: the filter still runs only what it named.

The project node is skipped when expanding — it consumes everything, so its dependencies are the whole graph and closing over it would turn every filtered run into a full one.

## Tests

- `test_runner_builds_a_dependency_the_filter_left_out` — an input, and an insight reading both that input and a model of its own. Fails without the change.
- `test_runner_leaves_a_dependency_that_is_already_built` — asserts the model's schema file is not rewritten on a second filtered run, so the optimisation is pinned rather than assumed.

2598 passed, `black --check` clean.

## Note

`_is_built` answers only for the types another job reads (models, sources, inputs) and returns true for anything else, so an unknown type is never dragged into a run.


tgsoverly:
Good call — those f-strings were a check that could drift from the writer and then quietly answer the wrong question. Pushed `visivo/output_paths.py`: one definition per artifact, with the writers going through it too.

| | |
| -- | -- |
| `run_dir(output_dir, run_id)` | the directory a run owns |
| `schema_file(run_dir, name)` | models and sources share `schemas/` |
| `input_metadata_file(run_dir, name)` | an input's computed options |
| `model_data_file(run_dir, name)` | a model's rows |

Now routed through it: both schema aggregators (`SchemaAggregator`, `ModelSchemaAggregator`), the model job's success message, the input job's write, and — the one that matters most here — `field_resolver`, which is what raises the `Missing schema for model` this PR is about. So the writer, the reader that fails, and the runner's check are all the same builder.

Two shapes because callers hold two different things: jobs build `{output_dir}/{run_id}` first, while `field_resolver` is already handed the run directory and passes it straight in. `run_dir()` covers the first, everything else takes a run directory.

What pins the agreement is `test_runner_leaves_a_dependency_that_is_already_built`: the first run writes through the jobs, the second asks the runner whether it is built. If the two sides ever stop agreeing, the model rebuilds and the mtime assertion fails. I noted that in its docstring rather than adding a unit test that would just restate the format strings.

2598 passed, `black --check` clean, and the original repro still behaves — cold run builds the models (4.78s), warm run leaves them alone (1.76s).

